PR #3, opened 2023-06-15
Patch 1

martinjm97:


README.md
```diff
@@ -61,26 +61,16 @@ pip install enzyme-jax
...
-
-# Build our extension.
 # Will create a whl in bazel-bin/enzyme_jax-VERSION-SYSTEM.whl
 bazel build :enzyme_jax
 ```
 
-After changing LLVM, it is necessary to rebuild and reinstall JAX.
+Finally, install the built library with:
+```sh
+pip install bazel-bin/enzyme_jax-VERSION-SYSTEM.whl
```
wsmoses, line 72: This is repeated three lines down, maybe remove the latter one? Also maybe give a warning thata you cannot run code using it, in the root of the git directory (and thus you need to cd into test)

martinjm97: I did it. I resolved the comment. I wrote the documentation. YAY!

wsmoses (approved):

PR #6, opened 2023-07-10
Attempt custom derivative via primitive

wsmoses:

PR #7, opened 2023-07-11
A mechanism to extract LLVM IR from JAX via XLA

ftynse:
This relies on a debug feature of XLA, but that feature has been quite stable so far. Note that the LLVM IR is not yet connected to JIT.


wsmoses:
Macos CI found a real error:
```
ERROR: /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-macmini-aarch64-3.0/build/default-macmini-aarch64-3-0/julialang/enzyme-jax/baztmp/524d6caf5a5363f7a423c1abcb043984/external/xla/xla/service/cpu/BUILD:185:11: no such target '@llvm-project//llvm:X86CodeGen': target 'X86CodeGen' not declared in package 'llvm' defined by /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-macmini-aarch64-3.0/build/default-macmini-aarch64-3-0/julialang/enzyme-jax/baztmp/524d6caf5a5363f7a423c1abcb043984/external/llvm-project/llvm/BUILD.bazel (did you mean 'CodeGen'? Tip: use `query "@llvm-project//llvm:*"` to see all the targets in that package) and referenced by '@xla//xla/service/cpu:cpu_compiler'
```

Ubuntu CI is a ficticious error (incidentally fixed by https://github.com/EnzymeAD/Enzyme-JAX/pull/6 so we can rebase that upon landing)

wsmoses:
I suppose the solution is to have this (https://github.com/openxla/xla/blob/e02756722f3c3157305ecefe9bc19aa0242462b8/xla/service/cpu/BUILD#L328) moved into the tsl select . Appears to be only an issue for arm mac [where we don't build X86]

ftynse:
> I suppose the solution is to have this (https://github.com/openxla/xla/blob/e02756722f3c3157305ecefe9bc19aa0242462b8/xla/service/cpu/BUILD#L328) moved into the tsl select . Appears to be only an issue for arm mac [where we don't build X86]

Why don't we build X86? It looks like we should according to https://github.com/EnzymeAD/Enzyme-JAX/blob/main/WORKSPACE#L17, which generates `targets.bzl`, which in turn feeds into https://github.com/llvm/llvm-project/blob/main/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel#L2271 that creates the `X86CodeGen` target. I suppose we can still enable the target even if it's not the native architecture, or does it lead to some other problem?

Also, if you can repro this with a standalone xla build, we can report this to xla and have it fixed properly.

wsmoses:
To reduce build time and binary size on arm mac, we don't build X86: https://github.com/EnzymeAD/Enzyme-JAX/blob/ab31488b69bc271d4d657d64cd6c85900150914c/.buildkite/pipeline.yml#L23

PR #10, opened 2023-11-02
Make JAX import into WORKSPACE work.

ingomueller-net:
The current import was non-functional, and that only remained unnoticed
because there is no BUILD dependency to JAX, so JAX isn't actually
built. (The CI tests install JAX through pip, so the JAX dependency
isn't needed there.)

The fix consists in moving the configuration from JAX's WORKSPACE file
(which is emptied when imported via `http_archive`) to the WORKSPACE
file of this project (plus fixing how its dependency `ducc` is loaded).
Now adding `@jax//jax` to some of the build targets actually (and
successfully) builds JAX.


ingomueller-net:
Something seems stuck. Maybe the build process got stuck, or it might need approval before it starts running?

wsmoses (approved):

ingomueller-net:
One of the build targets failed without error (that is visible to me). Can you check what is the problem and restart it?

ingomueller-net:
@wsmoses: Did you retrigger the build? It failed again -- presumable because it never actually ran? If that check isn't supposed to run, maybe you can force-merge the PR? I don't have the permissions...

PR #12, opened 2023-11-03
Update all dependencies.

ingomueller-net:
This PR depends on and therefore includes #10.

This PR updates the following dependencies in the WORKSPACE file to a
version from around November 2:

* LLVM
* XLA
* JAX (even though there is no actual BUILD dependency)
* Enzyme

The PR also adds direct dependencies to the following projects:

* rules_cc (because LLVM requires a newer version than the one that XLA
  imports)

For context: @ftynse pointed me to this repo to copy the Bazel recipe to another project that depends on JAX and I didn't make it work there yet. Since I got stuck on that project, I came back to this one to see if I can make it work here. Turns out I could. I hope that this will serve me as a good base to get the other project to work as well. I don't need this project currently, but since I grinded through making it work, you might as well benefit from it ;)


BUILD
```diff
@@ -29,19 +27,15 @@ cc_library(
...
 
-load("@rules_python//python:packaging.bzl", "py_wheel")
-
-load(":package.bzl", "py_package")
+load("@rules_python//python:packaging.bzl", "py_package", "py_wheel")
 
 py_package(
     name = "enzyme_jax_data",
     # Only include these Python packages.
-    packages = ["@//enzyme_jax:enzyme_call.so", "@llvm-project//clang:builtin_headers_gen"],
```
ingomueller-net, line 39: Warning: I don't understand what `packages` is supposed to do. If I use the line as is, the `.so` file is *not* included in the wheel...

BUILD
```diff
@@ -29,19 +27,15 @@ cc_library(
     ],
 )
 
-load("@rules_python//python:packaging.bzl", "py_wheel")
-
-load(":package.bzl", "py_package")
```
ingomueller-net, line 34: By reading the code, I had the feeling that the `package.bzl` from this repo had no effect other than calling the function of the same name in `rules_python` (plus *allowing* but not *using* the `prefix` parameter), so I removed it.

BUILD
wsmoses: So it's supposed to force the so (And those headers) to be made available in the pip package.

wsmoses: It does look like CI thinks that it can no longer find them, so we may need something else here (it worked in the past so I'm not sure what would've changed, and I'm curious if putting it back fixes anything)

ingomueller-net: When it was there, CI broke. I could observe the same behavior locally: only the py files ended up in the wheel. I don't know what changed, though...

wsmoses: Ah yeah so the problem is that the `package.bzl` forces the dependencies to be included as well, so by getting rid of our version of package.bzl you've removed the ability to import the actual dependencies.

Revert back to our package.bzl and see if it works?

ingomueller-net: I have also tried that. I started looking into the wheel rule because it broke, even before updating rules_python. Also, the old package.bzl doesn't have filtering logic for all I can tell.

Is there anything in the current behavior that is undesirable?

ingomueller-net:
I went through the packing rule again. If I revert the last commit, i.e., all changes to that rule, then the resulting `*.whl` does not contain the `.so` file:

```bash
python -c "from zipfile import ZipFile; from pprint import pprint; pprint(ZipFile('bazel-bin/enzyme_jax-0.0.5-py3-none-manylinux2014_x86_64.whl').namelist())" | grep \\.so
# no output
```

Commenting out the `prefix` or `packages` arguments does not change that behavior. Only if I use the function from the `@rules_python` package instead of the one from the reverted `package.bzl` *in addition* to commenting out `prefix` and `packages` parameters does it work. I also tried changing the `packages` parameter to `//enzyme_jax:enzyme_call` instead of `//enzyme_jax:enzyme_call.so` (i.e., without file extension) -- doing that doesn't fix the behavior. Using the rule from `@rules_python` without `packages` parameter does work for the old version of `rules_python`, though. With the new version, no form of the `packages` argument leads in the resulting wheel containing the `.so` file.

In short, after the changes in the previous commits, only the fixes in the last commit lead to a valid wheel.

(I just force-pushed an updated last commit where I remove the comment for the also-removed `packages` argument.)

ingomueller-net:
OK, I investigated more. Values in the `packages` arguments have all `.` replaced by `/` ([here](https://github.com/bazelbuild/rules_python/blob/3eebda1a/python/private/py_package.bzl#L41) -- WTF??) and are used as *prefix* for matching. So with `packages = ["enzyme_jax/enzyme_call"],`, the wheel actually contains the `*.so` file. That filter, instead, filtered out the headers. But I guess that wasn't the intention: they where part of the `packages` filter originally. As I see it the original `packages` filter said to include all files that where already selected as `deps`, so even when they were working, they had no effect (and when they were not, they broke the wheel). TL;DR: I think we should drop the argument as done in the last commit.

ingomueller-net:
OK, I fixed another bunch of problems. I am able to run `test/test.py` locally now 😄 Hopefully this will pass CI as well...

ingomueller-net:
You made it work :) Kudos!

PR #19, opened 2023-12-06
fix ir.pyt test

ftynse:
jax.arg_info was missing in test strings


wsmoses:
Somehow I had a version where that was removed (and I subsequently recall removing it). Will investigate briefly before merging, otherwise LGTM

PR #21, opened 2023-12-12
Python format

ftynse:
This is the same formatter as LLVM is using.


wsmoses (approved):

ftynse:
The failing test timed out...

PR #34, opened 2024-02-16
reuse Common.td from Enzyme

ftynse:


src/enzyme_ad/jax/Implementations/CommonExt.td
```diff
@@ -0,0 +1,8 @@
+// Symlink-forwarded from main Enzyme.
+include "src/enzyme_ad/jax/Implementations/Common.td"
+
```
wsmoses, line 3: oh sorry, we should really just push this to Enzyme-MLIR itself

PR #41, opened 2024-03-09
Handle inactive args from context

wsmoses:


src/enzyme_ad/jax/primitives.py
```diff
@@ -917,18 +924,36 @@ def cpp_call(
 
 
 def enzyme_jvp(arg_primals, arg_tangents, **kwargs):
+    print("arg_tan", arg_tangents)
+    print("kwargs", kwargs)
```
ftynse, line 928: Leftover debug spew

src/enzyme_ad/jax/primitives.py
```diff
@@ -1050,6 +1074,9 @@ def fwd_partial_eval(trace, *args, **kwargs):
 
 
 def primal_partial_eval(trace, *args, **kwargs):
+    print("trace ", trace)
+    print("args", args)
+    print("kwargs", kwargs)
```
ftynse, line 1079: Ditto. We can use `import logging` and potentially even attach to the jax logger if desired.

PR #55, opened 2024-03-14
Fix 0 dim reshape concat case

ivanradanov:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -859,6 +859,8 @@ struct ConcatAppendingReshape final
 
         auto prevshape = t.getOperand().getType().getShape();
         auto postshape = t.getType().getShape();
+        if (prevshape.size() == 0)
+          return failure();
```
wsmoses, line 863: I think we can still handle this case by adding 1 to nextshape here perhaps?

wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
ivanradanov: I think concatenate needs at least one dimension to work

wsmoses: Oh yeah it does sorry I meant to delete the comment before approval

PR #83, opened 2024-05-27
Add `stablehlo.einsum`, `stablehlo.unary_einsum` diffrules

### LINKED ISSUE #82: Missing differentiation rules for `einsum`, `unary_einsum`
I'm getting the following error in [EnzymeAD/Reactant.jl](https://github.com/EnzymeAD/Reactant.jl) when trying to differentiate `stablehlo.einsum`:

```
julia> f = Reactant.compile(grad, (a′, b′))
error: could not compute the adjoint for this operation %2 = "stablehlo.einsum"(%1, %0) <{einsum_config = "ij,jk->ik"}> : (tensor<2x2xf64>, tensor<2x2xf64>) -> tensor<2x2xf64>
Pipeline failed
```

I open the issue here because I believe here is where the EnzymeMLIR rules are declared for HLO dialects right?

mofeing:
Fixes #82 

@wsmoses I don't know much about LLVM, so I don't know how to implement them. But I wrote the prototypes and some annotations on how to continue the implementation.

### DIFF
--- a/src/enzyme_ad/jax/Implementations/HLODerivatives.td
+++ b/src/enzyme_ad/jax/Implementations/HLODerivatives.td
@@ -645,9 +645,6 @@ def : HLODerivative<"TransposeOp", (Op $x),
 def UnaryEinsum : HLOInst<"UnaryEinsumOp">;
 def Einsum : HLOInst<"EinsumOp">;
 
-// TODO can we refere to `complex.conj` without declaring it here?
-def Conj : Inst<"ConjOp", "complex">;
-
 def EinsumConfig : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, [{
   op.getEinsumConfig();
 }]>;
--- a/src/enzyme_ad/jax/Implementations/HLODerivatives.td
+++ b/src/enzyme_ad/jax/Implementations/HLODerivatives.td
@@ -662,11 +662,26 @@ def : HLODerivative<"UnaryEinsumOp", (Op $x),
                     ]
                   >;
 
+def EinsumGradConfigLhs : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, [{
+  StringRef config = op.getEinsumConfig();
+  auto [input, output] = config.split(StringRef("->"));
+  auto [lhs, rhs] = input.split(',');
+  const StringRef new_config = (output + "," + rhs + "->" + lhs).toStringRef();
+  getStrAttr(builder, new_config);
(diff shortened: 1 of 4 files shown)


wsmoses (commented):
You should copy the gradpad test which computes derivatives

wsmoses:
@mofeing look here for example of how to deal with the einsum config. https://github.com/EnzymeAD/Enzyme-JAX/blob/b357c8dbe21a1de78dfdab6891dcd20411b3b291/src/enzyme_ad/jax/Implementations/HLODerivatives.td#L362

I can help as well, but my question for you is offhand do you know the corresponding derivative rule for the config

mofeing:
> I can help as well, but my question for you is offhand do you know the corresponding derivative rule for the config

@wsmoses The reverse derivative for `einsum` is very easy. Basically it's just another `einsum` per input where you replace the computing . Imagine the following `einsum`:

$$ C = \mathtt{einsum}(''$ia,$ib->$ic'', A, B) $$

where $A,B,C$ are tensors and $ia,ib,ic$ denote the list of indices of each tensor. The string is the einsum config.

Then the reverse derivatives are the following:

$$ dA = \mathtt{einsum}(''$ib,$ic->$ia'', conj(B), dC) $$

$$ dB = \mathtt{einsum}(''$ia,$ic->$ib'', conj(A), dC) $$

I see that we need to do the following:
1. Modify the einsum config like the patterns above.
2. Apply the conjugate only if `A` or `B` respectively, are complex. The main reason is that `complex.conj` only works on `complex` types, not on `f32`/`f64`.

wsmoses:
[whoops meant to comment, not close...my b]

wsmoses:
In any case, want to try to add the config update?

mofeing:
Yeah, I see that `GlobalExpr` lets you write C++ code, right? I just don't know what I have access to and the methods of `op` inside that code.

wsmoses:
Yeah. You have op which is the current operation so it’s essentially the
MLIR c++ api for your op

mofeing:
@wsmoses just uploaded the `stablehlo.einsum` test with complex data. it might still error due to the result not being exactly the same, but it should be easily fixable.

# To do

- [ ] add activity support to `stablehlo.einsum` and `stablehlo.unary_einsum` diff rules
- [ ] fix diff rule for `stablehlo.einsum` on complex data

wsmoses:
@mofeing tests seem to fail, worth investigating?

PR #84, opened 2024-05-29
Fix Python tests

mofeing:


test/lit_tests/ir.pyt
```diff
@@ -62,9 +62,9 @@ print(f.lower(ones, twos).compiler_ir(dialect="stablehlo"))
 
 # CHECK: module @jit_f attributes {mhlo.num_partitions = 1 : i32, mhlo.num_replicas = 1 : i32} {
 # CHECK-NEXT:  func.func public @main
-# CHECK-NEXT:     %[[i0:.+]] = stablehlo.constant dense<2> : tensor<1xi64>
-# CHECK-NEXT:     %[[i1:.+]]:3 = stablehlo.custom_call @jaxzyme.aug(%0, %arg0, %arg1) : (tensor<1xi64>, tensor<2x3xf32>, tensor<5x7xf32>) -> (tensor<6x9xf32>, tensor<4x6xf32>, tensor<16xi8>)
-# CHECK-NEXT:     return %[[i1]]#0, %[[i1]]#1, %[[i1]]#2, %arg0, %arg1 : tensor<6x9xf32>, tensor<4x6xf32>, tensor<16xi8>, tensor<2x3xf32>, tensor<5x7xf32>
+# CHECK-NEXT:     %c = stablehlo.constant dense<2> : tensor<1xi64>
```
wsmoses, line 65: So the `[[i0:.+]]` syntax denotes a variable i0 which is defined by the regex .+ which should match this always. 

So I don't understand why this change is needed

mofeing: Ah okay. But this is what was crashing... Maybe `FileCheck` is not correctly matching those regex?

wsmoses:
Superceeded by https://github.com/EnzymeAD/Enzyme-JAX/pull/86

PR #87, opened 2024-06-01
Add all StableHLO ops from spec

mofeing:
This PR refactors `HLODerivatives.td` to add all the missing ops from the StableHLO spec.

One question that strikes me is whether we should add CHLO ops, since they lower to StableHLO and we could support them by running `chlo-legalize-to-stablehlo` pass before Enzyme.

It also adds the `HLOInactiveOp` class to annotate non-differentiable ops.


wsmoses (approved):

mofeing:
@wsmoses seems like `buildkite/enzyme-jax/pr` is not allowing me to merge?
Captura de pantalla 2024-06-01 a las 17 02 43

PR #90, opened 2024-06-01
Add a bunch of rules for scalar and non-differentiable functions

mofeing:
# To do

- [x] Register StableHLO dialect when using CHLO dialect
- [x] Fix tests on complex data (specially reverse diff)
  - [x] `stablehlo.complex`
  - [x] `chlo.conj`
  - [x] `stablehlo.einsum`
  - [x] `stablehlo.dot_general`


src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -541,15 +543,77 @@ def GradEinsumConfigRhs : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, [{
...
 def : HLODerivative<"AddOp", (Op $x, $y),
                     [
                       (DiffeRet),
                       (DiffeRet),
                     ]
                   >;
 
+def : HLOInactiveOp<"AfterAllOp">;
+
+def : HLOInactiveOp<"AndOp">;
```
wsmoses, line 563: Don't mark this as inactive since we want it to catch bithack errors (and activity analysis will be smart enough otherwise).

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -541,15 +543,77 @@ def GradEinsumConfigRhs : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, [{
...
+]>;
+
 def : HLODerivative<"AddOp", (Op $x, $y),
                     [
                       (DiffeRet),
                       (DiffeRet),
                     ]
                   >;
 
+def : HLOInactiveOp<"AfterAllOp">;
```
wsmoses, line 561: This should be handled later since properly handling parallelism means reversing the parallel dependencies

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -541,15 +543,77 @@ def GradEinsumConfigRhs : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, [{
...
+                      (CheckedMul (DiffeRet), (Div (Neg $y), (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y))))),
+                      (CheckedMul (DiffeRet), (Div $x, (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y)))))
+                    ]
+                  >;
+
 def : HLOReadOnlyIdentityOp<"BroadcastInDimOp">;
 
+def : HLOInactiveOp<"BroadcastAndOp">;
+
+def : HLOInactiveOp<"BroadcastOrOp">;
```
wsmoses, line 592: Same comment on not marking these per bithacks

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -603,6 +691,14 @@ def : HLODerivative<"MaxOp", (Op $x, $y),
                   (Select (Compare $x, $y, (LT)), (SelectIfActive $y, (Shadow $y), (HLOConstantFP<"0"> $y)), (SelectIfActive $x, (Shadow $x), (HLOConstantFP<"0"> $x)))
                   >;
 
+def : HLODerivative<"MinOp", (Op $x, $y),
+                  [
+                    (Select (Compare $x, $y, (GT)), (HLOConstantFP<"0"> $x), (DiffeRet)),
+                    (Select (Compare $x, $y, (GT)), (DiffeRet), (HLOConstantFP<"0"> $x))
```
wsmoses, line 697: For improving optimization can we use LT for this too but swap lhs and rhs (that way min and max derivatives can be CSEd)

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -612,11 +708,19 @@ def : HLODerivative<"MulOp", (Op $x, $y),
 
 def : HLODerivative<"NegOp", (Op $x), [(Neg (DiffeRet))]>;
 
+def : HLOInactiveOp<"NotOp">;
+
```
wsmoses, line 712: Same bithack comment

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -612,11 +708,19 @@ def : HLODerivative<"MulOp", (Op $x, $y),
...
+def : HLOInactiveOp<"OrOp">;
+
 def : HLOMemoryIdentityOp<"PadOp", [], [-1], (Op $op, $padval), [
   (Slice (TypeOf $op), (DiffeRet), (PadToSliceStart), (PadToSliceLimit), (PadToSliceStride)),
   (AssertingInactiveArg)
 ]>;
 
+def : HLOInactiveOp<"PartitionIdOp">;
+
+def : HLOInactiveOp<"PopcntOp">;
```
wsmoses, line 722: And here (tho partition ID is okay)

src/enzyme_ad/jax/Implementations/HLODerivatives.td
mofeing: mmm then `ShiftLeft` and `ShiftRight` should also **not** be marked as inactive? since they are usually used in bithacks

wsmoses: Yeah I wouldn't do any of the integer ones unless there's a reason they should be zero like round

src/enzyme_ad/jax/Implementations/HLODerivatives.td
mofeing: so `PartitionIdOp` is inactive...?

wsmoses: Yeah exactly that cannot contain any differential data (like threadid, etc)

wsmoses (commented):
Looks good to me, but add some tests?

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -705,7 +705,7 @@ def : HLODerivative<"SqrtOp", (Op $x),
 
 def : HLODerivative<"TanhOp", (Op $x),
                     [
```
wsmoses, line 707: Is there a reason for this. Mul seems like it would be more efficient?

mofeing: Nah, I just tried to be more consistent mathematically. I can turn it back to `Mul` but shouldn't some pass change the square to a `Mul` already?

wsmoses: Not presently so it’s worth keeping it like this for now

wsmoses: Also it may hit an issue in complex since the constant needs to be the correct type

wsmoses (approved):

PR #93, opened 2024-06-18
Fix grad sum

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Reactant.jl/issues/14

PR #100, opened 2024-06-27
Add compile to llvm backend

wsmoses:


src/enzyme_ad/jax/enzyme_call.cc
```diff
@@ -868,7 +868,7 @@ class CpuKernel {
     // force deletion of mod first explicitly
     mod = nullptr;
     return std::make_pair(res, tmpBuf);
-  }
+  }
```
mofeing, line 871: Just a small formatting request
```suggestion
  }
```

PR #104, opened 2024-07-18
update llama

ftynse:
- fix rmsnorm to compute what it should
- use complex numbers instead of rotation matrix
- don't append


ftynse:
Adapting this from another version that is known to run on proper data. Note that the weights for dense layers coming from pytorch are transposed and my version is using `x @ weights.mT` whereas this version uses `weights @ x`.

wsmoses (approved):

PR #142, opened 2024-10-16
Jmp/scatter gather

wsmoses:
@pengmai sorry for prematurely making a PR, but @mofeing if helpful and needed


wsmoses:
btw @Pangoraw and/or @avik-pal I don't think this PR is being actively worked on so feel free to take it over if the dynamic_gather and related stuff is blocking you

PR #149, opened 2024-10-29
IfOp

Pangoraw:
- region inlining if predicate is known
 - transform to select for values which are not defined in inner regions
 - forward
 - reverse

I don't think the reverse could be used as is in a XLA pipeline since Enzyme is not yet able to remove all enzyme ops (which I think it technically could if there is no loop?).

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -6271,6 +6271,46 @@ struct TransposeIsReshape final
   }
 };
 
+struct IfInline final : OpRewritePattern<mlir::stablehlo::IfOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(mlir::stablehlo::IfOp op,
+                                PatternRewriter &rewriter) const override {
+
+    auto iszero = matchPattern(op.getPred(), m_Zero());
+    auto isone = matchPattern(op.getPred(), m_One());
+
+    if (!iszero && !isone)
+      return failure();
+
+    auto current = op->getBlock();
+
+    auto &reg = isone ? op.getTrueBranch() : op.getFalseBranch();
+
+    if (reg.empty()) {
+      rewriter.eraseOp(op);
+      return success();
+    }
+    assert(reg.hasOneBlock());  // stablehlo.if only allows 1 or 0 block in the
+    auto *block = &reg.front(); // regions
+
+    auto term = block->getTerminator();
+    rewriter.replaceAllOpUsesWith(op, term->getOperands());
+    rewriter.eraseOp(term);
+
+    auto newBlock = rewriter.splitBlock(current, Block::iterator(op));
+
+    rewriter.inlineRegionBefore(reg, newBlock);
+
+    rewriter.mergeBlocks(block, current);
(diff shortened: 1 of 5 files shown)


test/lit_tests/batchtests/autodiff.mlir
```diff
@@ -0,0 +1,11 @@
...
+  func.func @f(%arg0: tensor<f32>) -> tensor<f32> {
+    %0 = stablehlo.multiply %arg0, %arg0 : tensor<f32>
+    return %0 : tensor<f32>
+  }
+
+  func.func @main(%arg0: tensor<10xf32>) -> tensor<10xf32> {
+    %0 = enzyme.batch @f(%arg0) { batch_shape = array<i64: 10> } : (tensor<10xf32>) -> tensor<10xf32>
+    return %0 : tensor<10xf32>
+  }
+}
```
wsmoses, line 11: can you add the test checks?

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6311,6 +6311,76 @@ struct IfInline final : OpRewritePattern<mlir::stablehlo::IfOp> {
...
+
+    rewriter.setInsertionPointToEnd(&replacement.getFalseBranch().front());
+    rewriter.replaceOpWithNewOp<mlir::stablehlo::ReturnOp>(
+        replacement.getFalseBranch().front().getTerminator(), falseReturns);
+
+    rewriter.replaceOp(op, results);
+    return success();
+  }
+};
+
```
wsmoses, line 6383: Can you also add the pattern if the condition is a constant?

test/lit_tests/batchtests/autodiff.mlir
Pangoraw: ah sorry I did not mean to include this in this pr. I will open up another PR that bumps Enzyme + JAX with this test.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
Pangoraw: If the condition is constant then the if is removed, leaving only the content of the corresponding region due to:

https://github.com/EnzymeAD/Enzyme-JAX/blob/d91410b4a0bd41ed19c70b8b0486af190408a6e3/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L6274-L6281

wsmoses (approved):
lgtm [though also we'll need to figure out a good interface for the push/pop removal]

Pangoraw:
> though also we'll need to figure out a good interface for the push/pop removal

Yes. I also took a look at BatchOpInterface for IfOp and in its current state the interface only allows to return a single operation so we could not really unroll the batching like we discussed. So maybe `BatchOpInterface` has to be modified to provide a `IRBuilder` to `createBatch` and implementers do the result mapping themselves.

wsmoses:
> > though also we'll need to figure out a good interface for the push/pop removal
> 
> Yes. I also took a look at BatchOpInterface for IfOp and in its current state the interface only allows to return a single operation so we could not really unroll the batching like we discussed. So maybe `BatchOpInterface` has to be modified to provide a `IRBuilder` to `createBatch` and implementers do the result mapping themselves.

Yeah I think so, as is it assumed a single op would suffice for batching

PR #151, opened 2024-11-04
Generic batch op interface

Pangoraw:
Depends on https://github.com/EnzymeAD/Enzyme/pull/2142.


mofeing:
I'm dubious if we should go this way? AFAIK Jax doesn't do this unrolling. Like batching over a lot of operations should still be done without unrolling.

Do you know for which ops should the unrolling be done?

Pangoraw:
Another possible generic strategy (maybe as a fallback) is to wrap the op in a `stablehlo.map` but there are no derivative rules for map yet.

test/lit_tests/batchtests/if.mlir
```diff
@@ -0,0 +1,21 @@
...
+    }) :  (tensor<i1>) -> tensor<f64>
+    return %result : tensor<f64>
+  }
+  func.func @main() {
+    %arg0 = stablehlo.constant dense<[[42.0, -42.0], [0.0, 1.0]]> : tensor<2x2xf64>
+    %0 = enzyme.batch @relu_broadcast_scalar(%arg0) {batch_shape = array<i64: 2, 2>} : (tensor<2x2xf64>) -> tensor<2x2xf64>
+    check.expect_eq_const %0, dense<[[42.0, 0.0], [0.0, 1.0]]> : tensor<2x2xf64>
+    return
+  }
+}
```
wsmoses, line 21: add the test lines?

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -1086,36 +1203,47 @@ class AutoDiffSort
...
+    ReduceOp::attachInterface<AutoDiffReduceCF<ReduceOp>>(*context);
+    WhileOp::attachInterface<AutoDiffReduceCF<WhileOp>>(*context);
+    BroadcastInDimOp::attachInterface<AutoDiffBroadcastInDimRev>(*context);
+    SliceOp::attachInterface<AutoDiffSliceRev>(*context);
+    ReduceOp::attachInterface<AutoDiffReduceRev>(*context);
+    ConcatenateOp::attachInterface<AutoDiffConcatenateRev>(*context);
+
+    ConstantOp::attachInterface<SHLOConstantOpBatchInterface>(*context);
+    TransposeOp::attachInterface<SHLOTransposeOpBatchInterface>(*context);
+    IfOp::attachInterface<SHLOUnrollOpBatchInterface<IfOp>>(*context);
```
wsmoses, line 1239: hm the control flow ops should definitely not use the interface here at a significant perf loss. Since effectively it puts the batching outside the if/while by unrolling, whereas putting it on the inside retains the ability to fuse/vectorize/etc

Pangoraw: yeah, it is not ideal as it generates a lot of code (proportional to the number of samples in the batch) but for IfOp and WhileOp it cannot know whether or not the same branch/number of iterations will be taken for all samples. I will look at how xla handles `stablehlo.map` under the hood. Or maybe we can emit the op inside a `stablehlo.map` instead.

wsmoses: you can tho! check if it's a constant which is a splat [see examples in enzymehlo-opt], which means its the same across all iters, or a broadcast from a single value

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -616,14 +616,282 @@ class AutoDiffConcatenateRev
...
+// For some ops with nested regions, identify if we can batch the inner regions
+// instead
+static LogicalResult tryToBatchInner(Operation *src, OpBuilder &builder,
+                                     IRMapping &mapper,
+                                     ArrayRef<int64_t> batchSizes) {
+  if (auto ifOp = dyn_cast<IfOp>(src)) {
+    auto iszero = matchPattern(ifOp.getPred(), m_Zero());
+    auto isone = matchPattern(ifOp.getPred(), m_One());
+
+    if (!iszero && !isone)
```
wsmoses, line 722: can we also test if the input operand would be a broadcastop of a single value?

Pangoraw: added it but I am having a hard time coming up with an input to test this where the batched version of the predicate is a broadcast_in_dim.

wsmoses: yeah I think long term this will require an extension to the batching interface to mark some operands as non-batched, in any case is fine for now

PR #153, opened 2024-11-06
remove transform.with_named_sequence attribute

Pangoraw:
To clean up `Reactant.@code_hlo` output

### DIFF
--- a/test/lit_tests/generate_td.mlir
+++ b/test/lit_tests/generate_td.mlir
@@ -22,6 +22,7 @@
 
 // TD: module attributes {transform.with_named_sequence} {
 // TD:  transform.named_sequence @__transform_main(%[[ROOT:.+]]: !transform.any_op) {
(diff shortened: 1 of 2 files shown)


test/lit_tests/generate_td.mlir
```diff
@@ -22,6 +22,7 @@
 
 // TD: module attributes {transform.with_named_sequence} {
 // TD:  transform.named_sequence @__transform_main(%[[ROOT:.+]]: !transform.any_op) {
+// CLEAN-NOT: attributes {transform.with_named_sequence}
```
wsmoses, line 25: I thought there was a different pass which removed things cc @ftynse

ftynse: There is a `-test-transform-erase-schedule` (or something like that), but it won't remove the attribute.

ftynse (approved):

PR #159, opened 2024-11-08
DynamicUpdateSliceOp reverse derivative

Pangoraw:


mofeing:
Great! But I think it was doable in TableGen?

I almost have it like this:

```tablegen
def : HLOMemoryIdentityOp<"DynamicUpdateSliceOp", [], [0, 1], (Op $operand, $update, $start_indices), [
  (DynamicUpdateSlice (DiffeRet), (BroadcastInDim (HLOConstantFP<"0"> /* eltype($operand) */), /* size($update) */), $start_indices),  //  { DiffeRet[start_indices...] = fill(0); return DiffeRet; }
  (DynamicSlice (DiffeRet), $start_indices, /* size($update) */), // DiffeRet[start_indices...]
  (AssertingInactiveArg),
]>;
```

Pangoraw:
I am not sure how to handle the variadic indices in tablegen.

mofeing:
I've checked out and apparently `Variadic` is an object so you don't have a variadic number of arguments but 1 argument with variadic number of elements (like `Vararg` in Julia).

So just putting that `AssertingInactiveArg` should be enough (if I understand correctly that command)

Pangoraw:
Note that if we keep the cpp version, I should add conjugates

mofeing:
do you mean complex conjugates? if so, why?

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -417,6 +417,84 @@ class AutoDiffBroadcastInDimRev
...
+                                 MGradientUtilsReverse *gutils) const {
+    auto op = cast<DynamicUpdateSliceOp>(orig);
+    Operation *newOp = gutils->getNewFromOriginal(orig);
+    OpBuilder cacheBuilder(newOp);
+
+    auto startIndices = op.getStartIndices();
+
+    SmallVector<Value> caches;
+    caches.reserve(startIndices.size());
+
```
wsmoses, line 484: this should have a check where we only push the caches if either operand is active

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -417,6 +417,84 @@ class AutoDiffBroadcastInDimRev
...
+      diffe = gutils->diffe(orig->getResult(0), builder);
+      auto operandDiffe = builder.create<DynamicUpdateSliceOp>(
+          orig->getLoc(), diffe,
+          cast<AutoDiffTypeInterface>(update.getType())
+              .createNullValue(builder, orig->getLoc()),
+          startIndices);
+
+      gutils->addToDiffe(operand, operandDiffe, builder);
+    }
```
wsmoses, line 448: we should check that there's an optimization which changes the pattern of

x = dyn_update_slice %v, 0's
y = old + x

into
q = dynamic_slice old
t = q + v
y = dynamic_update_slice old, t
^ aka not adding 0's when unnecesary

wsmoses:
> Note that if we keep the cpp version, I should add conjugates

actually here it should be fine.

Semantically reverse mode basically does

conj(x) =  f'(...) * conj(x)

However, here the work is linear so I think the conj'd we'd add will always cancel out

mofeing:
exactly, conjugation shouldn't be added explicitly here

mofeing (approved):

mofeing:
@Pangoraw from my side, we can merge it

maybe we can add a test with complex numbers just in case, but should be ok like it is now

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
mofeing: shouldn't some form of constant propagation + DCE do this?

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
Pangoraw: done. but hopefully the interface is not called in this case due to the op being constant:

https://github.com/EnzymeAD/Enzyme/blob/f8a48926b72284c8d8c6e892fcd2c44191ef06a9/enzyme/Enzyme/MLIR/Interfaces/EnzymeLogicReverse.cpp#L43-L47

PR #160, opened 2024-11-08
WhileOp reverse derivative

Pangoraw:
- [ ] conjugates

### DIFF
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -439,6 +439,15 @@ public:
       Block *oBB = &orig->getRegion(1).front();
       auto term = oBB->getTerminator();
 
+      for (auto operand : oBB->getArguments()) {
+        // All arguments should have no use outside this block therefore we can
+        // set their diffe to zero upon entering the reverse block to simplify
+        // the work of the remove-unnecessary-enzyme-ops pass.
+        if (!gutils->isConstantValue(operand)) {
+          gutils->zeroDiffe(operand, bodyBuilder);
+        }
+      }
+
       int revIdx = 1;
       for (auto &&[active, operand] :
            llvm::zip(operandsActive, term->getOperands())) {
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -372,11 +372,7 @@ public:
 
     // The reverse of the while loop is a for loop where the number
     // of iterations is either known or cached from the augmented primal.
-    WhileLoopInfo info(cast<WhileOp>(orig));
-    Value numIters =
-        info.computeInfo().succeeded() && info.isConstant()
-            ? makeI64Constant(orig->getLoc(), builder, info.getNumIters())
-            : gutils->popCache(caches[0], builder);
+    Value numIters = gutils->popCache(caches[0], builder);
 
     auto unrankedTensorType = RankedTensorType::get({}, builder.getI64Type());
     auto iterVar =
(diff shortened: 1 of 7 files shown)


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -340,6 +340,207 @@ class AutoDiffWhileFwd
...
+
+    return success(valid);
+  }
+
+  SmallVector<Value> cacheValues(Operation *orig,
+                                 MGradientUtilsReverse *gutils) const {
+    // The primal is augmented to store the number of iterations
+
+    auto newWhile = cast<WhileOp>(gutils->getNewFromOriginal(orig));
+    auto cond = &newWhile.getCond().front();
```
wsmoses, line 486: Usually one of the variables is a loop induction variable -- which means we don't need to store this. Not sure if this optimization would go here or in a canonicalization though.

Way more powerful versions of this would be here: https://github.com/llvm/Polygeist/blob/77c04bb2a7a2406ca9480bcc9e729b07d2c8d077/lib/polygeist/Passes/CanonicalizeFor.cpp#L662

The pro of doing this early though, is that we can potentially determine a static loop count and then make all of the augmented tensors <Loopsizexf32> instead of <?x32>

Pangoraw: I was able to reuse the analysis from enzyme-hlo-unroll which we can make more robust in the future to support more complex patterns.

wsmoses: in this case I don't think we even care if the value is a constant, we just cache that value regardless of constant or not

wsmoses: [and if it's constant cache fixups ought clean it up automatically for us]

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -340,6 +352,215 @@ class AutoDiffWhileFwd
...
+      for (auto operand : operands) {
+        body->addArgument(operand.getType(), orig->getLoc());
+      }
+      OpBuilder bodyBuilder(body, body->end());
+      auto one = bodyBuilder.create<ConstantOp>(
+          orig->getLoc(), unrankedTensorType,
+          SplatElementsAttr::get(unrankedTensorType,
+                                 ArrayRef<Attribute>(IntegerAttr::get(
+                                     bodyBuilder.getI64Type(), 1))));
+      Value bodyIterVar =
```
wsmoses, line 440: as weird, annoying as it is, we should probably use the createAdd from the autodiff type interface. Even though here we have to raise it back into a stablehlo.add, it does mean that new types will use the right add [if for example not correct in stablehlo.add]. I'm okay with this though if you feel strongly.

wsmoses: ah apologies, this add is for a new induction variable.

In this case same comment applies of we should re-use an existing induction variable [almost all cases presently], when possible.

wsmoses: Alternatively [or perhaps in addition to], we really should import most of that while op optimization stuff from polygeist [which also does redundant induction variable elimination somewhere iirc]

src/enzyme_ad/jax/Implementations/WhileLoopInfo.h
```diff
@@ -0,0 +1,47 @@
...
+    return getConstantStep().has_value() &&
+           getConstantStart().has_value() &&
+           getConstantLimit().has_value();
+  }
+
+  std::optional<int64_t> getConstantStep();
+  std::optional<int64_t> getConstantStart();
+  std::optional<int64_t> getConstantLimit();
+
+  /// Needs to be constant
```
wsmoses, line 41: Yeah here for the get number of induction variables, we can take a builder and compute the number of iterations

test/lit_tests/diffrules/stablehlo/while3.mlir
```diff
@@ -0,0 +1,98 @@
...
+    }, {
+    ^bb0(%arg0: tensor<i64>, %arg1: tensor<f64>):
+      %new_sum = stablehlo.multiply %arg1, %arg1 : tensor<f64>
+      %new_i = stablehlo.add %arg0, %one : tensor<i64>
+      stablehlo.return %new_i, %new_sum : tensor<i64>, tensor<f64>
+    }) : (tensor<i64>, tensor<f64>) -> (tensor<i64>, tensor<f64>)
+    return %results1 : tensor<f64>
+  }
+}
+
```
wsmoses, line 38: can you add forward and reverse checks here?

test/lit_tests/diffrules/stablehlo/while3.mlir
```diff
@@ -0,0 +1,98 @@
...
+// CHECK-NEXT:    %9 = "enzyme.get"(%5) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:    "enzyme.set"(%5, %cst) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
+// CHECK-NEXT:    %10:2 = stablehlo.while(%iterArg = %c_1, %iterArg_2 = %9) : tensor<i64>, tensor<f64>
+// CHECK-NEXT:     cond {
+// CHECK-NEXT:      %14 = stablehlo.compare  LT, %iterArg, %c : (tensor<i64>, tensor<i64>) -> tensor<i1>
+// CHECK-NEXT:      stablehlo.return %14 : tensor<i1>
+// CHECK-NEXT:    } do {
+// CHECK-NEXT:      %14 = stablehlo.add %iterArg, %c_0 : tensor<i64>
+// CHECK-NEXT:      "enzyme.set"(%4, %iterArg_2) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
+// CHECK-NEXT:      %15 = "enzyme.get"(%4) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
```
wsmoses, line 77: we should make sure that remove-enzyme-ops can deal with this %4 case without extra work.

%4 is defined right before this in a dominating way allowing %15 to be replaced with %iterArg_2.

Then now that only set's are called on it, we should be able to remove the whole thing.

No while special case handling required here

test/lit_tests/diffrules/stablehlo/while3.mlir
```diff
@@ -0,0 +1,98 @@
...
+// CHECK-NEXT:      %22 = "enzyme.get"(%1) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:      %23 = arith.addf %22, %21 : tensor<f64>
+// CHECK-NEXT:      "enzyme.set"(%1, %23) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
+// CHECK-NEXT:      %24 = "enzyme.get"(%1) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:      "enzyme.set"(%1, %cst) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
+// CHECK-NEXT:      stablehlo.return %14, %24 : tensor<i64>, tensor<f64>
+// CHECK-NEXT:    }
+// CHECK-NEXT:    %11 = "enzyme.get"(%0) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:    %12 = arith.addf %11, %10#1 : tensor<f64>
+// CHECK-NEXT:    "enzyme.set"(%0, %12) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
```
wsmoses, line 95: and also here with %0

test/lit_tests/diffrules/stablehlo/while3.mlir
```diff
@@ -0,0 +1,98 @@
...
+// CHECK-NEXT:      stablehlo.return %14 : tensor<i1>
+// CHECK-NEXT:    } do {
+// CHECK-NEXT:      %14 = stablehlo.add %iterArg, %c_0 : tensor<i64>
+// CHECK-NEXT:      "enzyme.set"(%4, %iterArg_2) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
+// CHECK-NEXT:      %15 = "enzyme.get"(%4) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:      "enzyme.set"(%4, %cst) : (!enzyme.Gradient<tensor<f64>>, tensor<f64>) -> ()
+// CHECK-NEXT:      %16 = "enzyme.pop"(%3) : (!enzyme.Cache<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:      %17 = "enzyme.pop"(%2) : (!enzyme.Cache<tensor<f64>>) -> tensor<f64>
+// CHECK-NEXT:      %18 = stablehlo.multiply %15, %17 : tensor<f64>
+// CHECK-NEXT:      %19 = "enzyme.get"(%1) : (!enzyme.Gradient<tensor<f64>>) -> tensor<f64>
```
wsmoses, line 82: The %1 cache here confuses me. In principle your good design setup such that we don't have any gradient +='s that span the outside of the while regions. Yet here %4 is not able to be mem2reg'd within scope. So something is clearly going awry...

wsmoses:
cc @mofeing I think the same argument of linearity from earlier applies here where we don't need to care about conjugates [which is nice]

PR #167, opened 2024-11-10
Improve enzyme gradient ops removal in while op

Pangoraw:
- inclusive ranges.
 - floating point ranges.
 - complete enzyme.set/get removal by zeroing gradients.


src/enzyme_ad/jax/Implementations/WhileLoopInfo.cpp
```diff
@@ -47,43 +49,89 @@ LogicalResult WhileLoopInfo::computeInfo() {
...
+    auto numIters_f = (limit_f - start_f) / step_f;
+
+    APSInt numIters_i(64, true);
+    bool isExact = false;
+    if (numIters_f.convertToInteger(numIters_i, llvm::APFloat::rmTowardZero,
+                                    &isExact) != llvm::APFloat::opOK ||
+        !isExact)
+      return std::nullopt;
+
+    return numIters_i.getSExtValue() + inclusive;
```
wsmoses, line 97: I don't think this is mathematically correct if step != 1

e.g. 2 * i < 3    can be 0 or 1,  as well as 2 * i <= 3

wsmoses: I think the better way to do the change is to add 1 to the limit, then use the < code

src/enzyme_ad/jax/Implementations/WhileLoopInfo.cpp
```diff
@@ -47,43 +49,89 @@ LogicalResult WhileLoopInfo::computeInfo() {
...
+      op->getLoc(), unrankedTensorType,
+      builder.create<stablehlo::DivOp>(
+          op->getLoc(),
+          builder.create<stablehlo::SubtractOp>(op->getLoc(), limit, start),
+          step));
+
+  if (inclusive) {
+    numIters = builder.create<stablehlo::AddOp>(
+        op->getLoc(), numIters,
+        builder.create<ConstantOp>(
```
wsmoses, line 129: same comment as above

src/enzyme_ad/jax/Implementations/WhileLoopInfo.h
```diff
@@ -19,25 +19,26 @@ namespace enzyme {
 struct WhileLoopInfo {
   WhileOp op;
 
-  mlir::Value start;   // garanteed to dominate the while op
-  mlir::Value limit;   // not garanteed to dominate the while op
-  mlir::Value step; // not garanteed to dominate the while op
+  mlir::Value start; // garanteed to dominate the while op
```
wsmoses, line 22: guaranteed [misspelled]

test/lit_tests/unroll2.mlir
```diff
@@ -0,0 +1,42 @@
+// RUN: enzymexlamlir-opt %s --enzyme-hlo-unroll | FileCheck %s
+
+module {
+  func.func @main(%arg0: tensor<i64>) -> tensor<i64> {
+    %c = stablehlo.constant dense<10> : tensor<i64>
+    %c_0 = stablehlo.constant dense<1> : tensor<i64>
+    %0:2 = stablehlo.while(%iterArg = %c_0, %iterArg_1 = %arg0) : tensor<i64>, tensor<i64>
```
wsmoses, line 7: so floats are weird and not necessarily commutative, so I'm not confident this will always work.

Instead could we have an optimization that changes while comparisons on floats into int's, and then we just have the int AD special case?

Pangoraw: I agree, I think most of the changes in this PR should be reverted so that Enzyme-JAX's canonical representation of for loops is only in the [0,N[ range and we can canonicalize inclusive or floating point range on top of those. In the meantime, I will make the Reactant PR emit [0,N[ for loops as well. Always having [0,N[ is handy because a cache push will be a dynamic_update_slice at loop index and cache pop is dynamic_slice at N - i.

PR #175, opened 2024-11-29
Fix complex rules for imag, real and abs

Pangoraw:
Depends on https://github.com/EnzymeAD/Enzyme/pull/2183.


wsmoses:
okay now that the Enzyme PR is merged, you need to update the hash, otherwise lgtm

Pangoraw:
it seems that bumping the Enzyme commit broke some CI builds, we need an xla bump in this case to update the llvm version as well ?

wsmoses:
Our main might be slightly out of date tho so we may need to do some minor fix ups to enzyme every so often

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6172,6 +6173,26 @@ struct ImagOpCanon final : OpRewritePattern<mlir::stablehlo::ImagOp> {
   }
 };
 
+// (conj (complex a, (neg b))) -> (complex a b)
+struct ConjComplexNegate final : OpRewritePattern<mlir::chlo::ConjOp> {
```
wsmoses, line 6177: can this be added to the same default python hlo_opts string

wsmoses:
bump was fixed here: https://github.com/EnzymeAD/Enzyme-JAX/pull/187 so you should rebase then be good ot go

wsmoses (approved):

PR #178, opened 2024-12-03
Convolution backward

Pangoraw:
~~Implemented only for stablehlo (not mhlo) since they have slight differences in attribute types that made it cumbersome. We can port to mhlo if needed.~~


Pangoraw:
> out of curiosity what was the error in mhlo land

At least:
 - windowReversal is MHLO_BoolElementsAttr in mhlo and DenseBoolArrayAttr in stablehlo, so one cannot index in windowReversal in mhlo (it seems it need to unwrap one more attribute).
 - window strides is I64ElementsAttr in mhlo and DenseI64ArrayAttr in stablehlo and I cannot manage to use SmallVector<int64_t> to satisfy both

is there a specific variable where I can `#ifdef MHLO` ?

wsmoses:
See https://github.com/EnzymeAD/Enzyme-JAX/blob/402b2ce27f3b15ad5eb12fd223ebb170a8b65f9a/src/enzyme_ad/jax/Implementations/MHLOAutoDiffOpInterfaceImpl.cpp#L35 and https://github.com/EnzymeAD/Enzyme-JAX/blob/402b2ce27f3b15ad5eb12fd223ebb170a8b65f9a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp#L36

Pangoraw:
thanks, I'll try to make it work for both.

wsmoses (approved):
TIL about the literal syntax!

PR #181, opened 2024-12-09
 support dilation and feature/batch group count in convolution reverse 

Pangoraw:


test/lit_tests/diffrules/stablehlo/convolution.mlir
```diff
@@ -19,6 +18,9 @@ module {
...
-// REVERSE-NEXT{LITERAL}:    %1 = stablehlo.convolution(%arg0, %arg2) dim_numbers = [f, 0, 1, b]x[i, 0, 1, o]->[0, 1, b, f], window = {stride = [1, 1], pad = [[0, 0], [0, 0]], lhs_dilate = [1, 1], rhs_dilate = [1, 1], reverse = [false, false]} {batch_group_count = 1 : i64, feature_group_count = 1 : i64, precision_config = [#stablehlo<precision DEFAULT>, #stablehlo<precision DEFAULT>]} : (tensor<8x66x66x512xf32>, tensor<8x64x64x512xf32>) -> tensor<3x3x512x512xf32>
-// REVERSE-NEXT{LITERAL}:    return %0, %1 : tensor<8x66x66x512xf32>, tensor<3x3x512x512xf32>
+// REVERSE-NEXT{LITERAL}:    %1 = stablehlo.reshape %0 : (tensor<8x66x66x512xf32>) -> tensor<8x66x66x1x512xf32>
+// REVERSE-NEXT{LITERAL}:    %2 = stablehlo.transpose %1, dims = [1, 2, 0, 3, 4] : (tensor<8x66x66x1x512xf32>) -> tensor<66x66x8x1x512xf32>
+// REVERSE-NEXT{LITERAL}:    %3 = stablehlo.reshape %2 : (tensor<66x66x8x1x512xf32>) -> tensor<8x66x66x512xf32>
```
Pangoraw, line 23: We should add an optimization to get rid of this pattern in a follow-up:

```
(squeeze (transpose (unsqueeze x))) -> x
```

wsmoses (approved):

test/lit_tests/diffrules/stablehlo/convolution.mlir
Pangoraw: yes that's right it has no effect only when permuting 1 sized dims.

This is to support batch group count which is moved in the feature dimension during the reverse, having a static select tablegen operator would be nice for these things:

```
(StaticSelect (Cond), $a, $b)
```

would create either `$a` or `$b` based on the condition (would also be useful for #176 IIUC).

mofeing: there's this https://llvm.org/docs/TableGen/ProgRef.html#if-select-statements-based-on-a-test but i don't how to use it 😅

Pangoraw: yeah, actually I think this reshape is wrong here because it's not actually a noop. We actually have the optimization I was referring to already, it's called TransposeIsReshape

Pangoraw: I fixed the problem in the reshape/transpose, the operations are completely removed in the case of batchGroupCount == 1 (which will always be the case from NNlib.conv).

PR #189, opened 2024-12-14
Adds support for LSP

vimarsh6739:
@mofeing you can use this too


mofeing:
what for? i think it was @glou-nes who tried to get the compile commands for the LSP

vimarsh6739:
You're right, I just took a look at the slack thread again - I still think this is good to have.

PR #193, opened 2024-12-16
feat: `dynamic_gather` to `gather` opt pass

avik-pal:


Pangoraw:
There need to be a corresponding pattern for the new opt in [TransformOps.td](https://github.com/EnzymeAD/Enzyme-JAX/blob/main/src/enzyme_ad/jax/TransformOps/TransformOps.td).

And then add the pattern in the python opt pipeline here: https://github.com/EnzymeAD/Enzyme-JAX/blob/293a73dcd1801d109924b128184d828c8decb991/src/enzyme_ad/jax/primitives.py#L91

wsmoses (approved):

PR #194, opened 2024-12-16
Reduce window reverse when op is maximum

Pangoraw:
For the maxpooling operator.


wsmoses (commented):
add to transform ops and hlo_opts?

wsmoses (approved):
lgtm, but fix the format so CI is happy?

PR #197, opened 2024-12-17
`enzyme.broadcast` conversion

jumerckx:
Corresponding Enzyme PR for batched autodiff: https://github.com/EnzymeAD/Enzyme/pull/2181
Is `arith-raise` the correct place for this or does this need a new pass?


wsmoses (approved):
nah imo this is totally a fine place to put it

PR #204, opened 2024-12-27
mwe for stablehlo

joelberkeley:


example.cpp
```diff
@@ -0,0 +1,140 @@
...
+    xla::XlaBuilder builder("root");
+    auto xlaScalarf64 = xla::ShapeUtil::MakeScalarShape((xla::PrimitiveType) 12);
+    auto arg = xla::Parameter(&builder, 0, xlaScalarf64, "arg");
+    auto proto = builder.Build(xla::Square(arg))->proto();
+
+    mlir::MLIRContext ctx;
+    mlir::DialectRegistry registry_;
+    ctx.appendDialectRegistry(registry_);
+    mlir::mhlo::registerAllMhloDialects(registry_);
+    mlir::stablehlo::registerAllDialects(registry_);
```
wsmoses, line 79: @joelberkeley if you're going to directly emit enzyme dialect here, I believe also need to do ctx.insert<enzyme::EnzymeDialect>();

joelberkeley: do you mean `registry_.insert<enzyme::EnzymeDialect>()`?

joelberkeley: I tried `registry_.insert<enzyme::EnzymeDialect>()` (in my framework, not in the mwe), and I'm still seeing that error. Note that line is already called in `prepareRegistry` https://github.com/EnzymeAD/Enzyme-JAX/blob/fd186b7f8ba82c41bf89a2e623d009746f13426f/src/enzyme_ad/jax/RegistryUtils.cpp#L79

joelberkeley: @wsmoses can I ask if the approach looks correct broadly?

Issue #184, opened 2024-12-12
Fuse multiple `dynamic_gather` into a single `dynamic_gather` + multiple `slice`?


avik-pal:
```mlir
module {
  func.func @main(%arg0: tensor<6x6xf64>) -> tensor<6x6xf64> {
    %cst = stablehlo.constant dense<1.000000e+00> : tensor<6x6xf64>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<6x6xf64>
    %c = stablehlo.constant dense<[[1, 0], [2, 1], [3, 2], [4, 3], [5, 4], [0, 0], [1, 1], [2, 2], [3, 3], [4, 4], [5, 5], [0, 1], [1, 2], [2, 3], [3, 4], [4, 5]]> : tensor<16x2xi64>
    %c_1 = stablehlo.constant dense<[[0, 1], [1, 2], [2, 3], [3, 4], [4, 5]]> : tensor<5x2xi64>
    %c_2 = stablehlo.constant dense<[[0, 0], [1, 1], [2, 2], [3, 3], [4, 4], [5, 5]]> : tensor<6x2xi64>
    %c_3 = stablehlo.constant dense<1> : tensor<2xi64>
    %c_4 = stablehlo.constant dense<[[1, 0], [2, 1], [3, 2], [4, 3], [5, 4]]> : tensor<5x2xi64>
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<6x6xf64>) -> tensor<6x6xf64>
    %1 = "stablehlo.dynamic_gather"(%0, %c_4, %c_3) <{dimension_numbers = #stablehlo.gather<collapsed_slice_dims = [0, 1], start_index_map = [0, 1], index_vector_dim = 1>}> : (tensor<6x6xf64>, tensor<5x2xi64>, tensor<2xi64>) -> tensor<5xf64>
    %2 = "stablehlo.dynamic_gather"(%0, %c_2, %c_3) <{dimension_numbers = #stablehlo.gather<collapsed_slice_dims = [0, 1], start_index_map = [0, 1], index_vector_dim = 1>}> : (tensor<6x6xf64>, tensor<6x2xi64>, tensor<2xi64>) -> tensor<6xf64>
[...]

PR #207, opened 2024-12-28
Add "concat of consecutive gathers is replaced with a larger gather"

glou-nes:
Add one of the pattern in #184.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -27,6 +27,9 @@
 #include "xla/mlir_hlo/mhlo/IR/hlo_ops.h"
 
 #include "stablehlo/dialect/TypeInference.h"
+#include <cstdint>
+#include <llvm/ADT/STLExtras.h>
+#include <llvm/ADT/SmallVector.h>
 
 #define DEBUG_TYPE "enzyme"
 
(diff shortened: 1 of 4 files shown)


test/lit_tests/concanetegather.mlir
```diff
@@ -0,0 +1,41 @@
...
+           %2 = "stablehlo.dynamic_gather"(%0, %c_2, %c_3) <{dimension_numbers = #stablehlo.gather<collapsed_slice_dims = [0, 1], start_index_map = [0, 1], index_vector_dim = 1>}> : (tensor<6x6xf64>, tensor<6x2xi64>, tensor<2xi64>) -> tensor<6xf64>
+           %3 = "stablehlo.dynamic_gather"(%0, %c_1, %c_3) <{dimension_numbers = #stablehlo.gather<collapsed_slice_dims = [0, 1], start_index_map = [0, 1], index_vector_dim = 1>}> : (tensor<6x6xf64>, tensor<5x2xi64>, tensor<2xi64>) -> tensor<5xf64>
+           %4 = stablehlo.concatenate %1, %2, %3, dim = 0 : (tensor<5xf64>, tensor<6xf64>, tensor<5xf64>) -> tensor<16xf64>
+           %5 = "stablehlo.scatter"(%cst_0, %c, %4) <{scatter_dimension_numbers = #stablehlo.scatter<inserted_window_dims = [0, 1], scatter_dims_to_operand_dims = [0, 1], index_vector_dim = 1>}> ({
+           ^bb0(%arg1: tensor<f64>, %arg2: tensor<f64>):
+             stablehlo.return %arg2 : tensor<f64>
+           }) : (tensor<6x6xf64>, tensor<16x2xi64>, tensor<16xf64>) -> tensor<6x6xf64>
```
wsmoses, line 20: Can you simplify the test case to just be gathers of some args and a concatenation thereof

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6650,6 +6650,61 @@ struct DynamicGatherOpIsNotDynamic
   }
 };
 
+struct ConcatenateGather : public OpRewritePattern<stablehlo::ConcatenateOp> {
+  using OpRewritePattern<stablehlo::ConcatenateOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(stablehlo::ConcatenateOp op,
+                                PatternRewriter &rewriter) const override {
+    if (op.getDimension() != 0)
+      return failure();
```
wsmoses, line 6659: Can this be generalized past dim 0

glou-nes: thank! I now use concat const folding and it can be relaxed.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6650,6 +6650,49 @@ struct DynamicGatherOpIsNotDynamic
...
+struct ConcatGather : public OpRewritePattern<stablehlo::ConcatenateOp> {
+  using OpRewritePattern<stablehlo::ConcatenateOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(stablehlo::ConcatenateOp op,
+                                PatternRewriter &rewriter) const override {
+    SmallVector<std::tuple<Attribute, Value, ArrayRef<int64_t>, bool>>
+        gatherParameters;
+    SmallVector<DenseIntElementsAttr> sliceSizesAttrs;
+    for (auto ope : op->getOperands())
+      if (auto operation = llvm::dyn_cast_if_present<stablehlo::GatherOp>(
```
wsmoses, line 6662: this can just be operation = ope.getDefiningOp<stablehlo::GatherOp>()

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6650,6 +6650,49 @@ struct DynamicGatherOpIsNotDynamic
...
+              ope.getDefiningOp())) {
+
+        DenseIntElementsAttr sliceSizesAttr;
+        if (!matchPattern(operation.getStartIndices(),
+                          m_Constant(&sliceSizesAttr)))
+          return failure();
+        sliceSizesAttrs.push_back(sliceSizesAttr);
+        gatherParameters.push_back(
+            {operation.getDimensionNumbers(), operation.getOperand(),
+             operation.getSliceSizes(), operation.getIndicesAreSorted()});
```
wsmoses, line 6672: ideally here we should be able to do this if any consecutive ops in the concat are compatible gathers [e.g. it is okay if some ops arent gathers, we'll just concat those segments ourselves into a big gather, then make an outergather where that sequence is replaced with the new gather op

glou-nes: I think I don't get it. If I understand correctly, for instance: 
```
concat [gather0, gather1, non_gather, gather2]
=>
concat[gather0+1, gather3, gather2]
=>
gather4
```
Are you suggesting to insert a gather operation to replace the non_gather operation, with the condition that this transformation is valid only if a partition of the `non_gather` tensor can be constructed using the constrained parameters and input of other gathers?

wsmoses: Not quite I’m suggesting something like

Concat(op1, op2, gather1, gather2, op3, gather4., op4, gather5, gather6, gather7, op8, op9)

becomes

concat(op1, op2, gather 1+2, op3, gather4, op4, gather 5+6+7, op8, op9)

glou-nes: Alright! Thank for the example,  I pushed a version with this behavior.

wsmoses (approved):

glou-nes:
I want to add two additional tests here: one with a non gather operation and one with two different gather parameters.

wsmoses:
@glou-nes anything blocking here?

glou-nes:
> @glou-nes anything blocking here?

I need to recheck the logic properly, I will add the missing examples today.

PR #208, opened 2024-12-31
feat: common simplifications

avik-pal:
- [x] `a + (-b)` => `a - b` if `-b` isn't used anywhere
- [x] `<op>(transpose(a), transpose(b))` => `transpose(<op>(a, b))` if only use
- [x] `<op>(transpose(a), cst)` => `transpose(<op>(a, transposed_cst))`
- [x] no_nan simplifications
  - [x] `a - a` => `0`
  - [x] `c = a + b`; `d = c - b` => `d = a`
  - [x] `c = a + b`; `d = b - c` => `d = -a`

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -2679,6 +2679,23 @@ struct SubSimplify : public OpRewritePattern<mlir::stablehlo::SubtractOp> {
   }
 };
 
+struct NoNanSelfSubSimplify
+    : public OpRewritePattern<mlir::stablehlo::SubtractOp> {
+  using OpRewritePattern<mlir::stablehlo::SubtractOp>::OpRewritePattern;
+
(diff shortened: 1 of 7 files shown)


avik-pal:
@wsmoses is `a - a` to `0` for floats not a valid transform? There is a is integer check in SubSimplify

```cpp
    if (isa<IntegerType>(op.getType().getElementType()) &&
        op.getLhs() == op.getRhs()) {
```

wsmoses:
To be compliant it might also require checking it isn't nan (or enabling the equivalent of fast math)

avik-pal:
does stablehlo provide fast math flags? arith seems to provide but can't find anything in stablehlo specs

wsmoses:
No we've usually been just parameterizing the canonicalizations here where relevant (eg search for like using or something)

wsmoses:
https://github.com/EnzymeAD/Enzyme-JAX/blob/52e14543628df5e3f806dce096243c4603d3e103/src/enzyme_ad/jax/Passes/Passes.td#L49

wsmoses (approved):

PR #215, opened 2025-01-03
feat: convert to rsqrt

avik-pal:


wsmoses (approved):

wsmoses:
we need to upgrade the version of gcc used by CI [so it's not running rn], merging

PR #218, opened 2025-01-04
const prop for log and logp1

### LINKED ISSUE #212: Missing constprop through `stablehlo.log`
```mlir
module {
  func.func @main(%arg0: tensor<f64>) -> tensor<f64> {
    %cst = stablehlo.constant dense<5.000000e-01> : tensor<f64>
    %0 = stablehlo.log %cst : tensor<f64>
    return %0 : tensor<f64>
  }
}
```

vimarsh6739:
resolves #212

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -2209,46 +2209,46 @@ struct DynamicUpdateSliceConstProp final
   }
 };
 
+template <auto f>
+LogicalResult unaryConstProp(Operation *op, PatternRewriter &rewriter) {
+  // return if not constant
+  DenseElementsAttr inputAttr;
+  if (!matchPattern(op->getOperand(0), m_Constant(&inputAttr)))
+    return failure();
+
+  stablehlo::Tensor inputTen;
+  RankedTensorType ty = cast<RankedTensorType>(op->getResultTypes()[0]);
+
+  if (inputAttr.isSplat()) {
+
+    ty = RankedTensorType::get(
+        {}, cast<ShapedType>(op->getResultTypes()[0]).getElementType());
+    inputTen = stablehlo::makeTensor(inputAttr.resizeSplat(ty));
+  } else {
+    inputTen = mlir::stablehlo::constantOp(inputAttr);
+  }
+  // get the resultType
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -2209,6 +2209,57 @@ struct DynamicUpdateSliceConstProp final
...
+    // return if not constant
+    DenseElementsAttr inputAttr;
+    if (!matchPattern(op.getOperand(), m_Constant(&inputAttr)))
+      return failure();
+
+    // get the resultType
+    auto resultType = op.getType().cast<ShapedType>();
+
+    // Convert constant to tensor, compute log, then convert back to attribute
+    stablehlo::Tensor inputTen = mlir::stablehlo::constantOp(inputAttr);
```
wsmoses, line 2226: Look at some of the other examples, can you add a fast path if the attribute is a splat (so we don’t need to compute the same value for a huge constant of the same inputs)

vimarsh6739: Kind of like here? https://github.com/EnzymeAD/Enzyme-JAX/blob/fd186b7f8ba82c41bf89a2e623d009746f13426f/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L2186

wsmoses: more like this: https://github.com/EnzymeAD/Enzyme-JAX/blob/fd186b7f8ba82c41bf89a2e623d009746f13426f/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L3237

wsmoses: otherwise we end up iterating through all inputs: https://github.com/openxla/stablehlo/blob/c06b1a10d66069505ccbab72b21f6fa0547c617d/stablehlo/reference/Ops.cpp#L1911

vimarsh6739: Ah I see so replacement will still be the same, we are just playing around with the tensor type

vimarsh6739:
Can we merge this in?

wsmoses (approved):

PR #220, opened 2025-01-05
feat: DynamicSliceOp adjoint

avik-pal:


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -801,6 +801,69 @@ class AutoDiffSliceRev
                           MGradientUtilsReverse *gutils) const {}
 };
 
+class AutoDiffDynamicSliceRev
+    : public ReverseAutoDiffOpInterface::ExternalModel<AutoDiffDynamicSliceRev,
+                                                       DynamicSliceOp> {
```
wsmoses, line 806: can we do this inside the tablegen file https://github.com/EnzymeAD/Enzyme-JAX/blob/main/src/enzyme_ad/jax/Implementations/HLODerivatives.td with corresponding globalops as needed?

That will enable automatically generating the batched version/etc as needed

avik-pal: I am currently getting

```
envs/dyn_slice_deriv.mlir:14:10: error: expects number of start_indices to match operand rank: 1 vs 2.
    %4 = stablehlo.dynamic_slice %0, %c_4, %3, sizes = [32, 1] : (tensor<32x1xf32>, tensor<i64>, tensor<i64>) -> tensor<32x1xf32>
         ^
envs/dyn_slice_deriv.mlir:14:10: error: 'stablehlo.dynamic_update_slice' op failed to infer returned types
    %4 = stablehlo.dynamic_slice %0, %c_4, %3, sizes = [32, 1] : (tensor<32x1xf32>, tensor<i64>, tensor<i64>) -> tensor<32x1xf32>
         ^
envs/dyn_slice_deriv.mlir:14:10: note: see current operation: %386 = "stablehlo.dynamic_update_slice"(%385, %382, %384) : (tensor<32x1xf32>, tensor<32x1xf32>, tensor<i64>) -> tensor<32x1xf32>
```

does this have something to do with the dynamic_slice having variadic inputs and if so how do I do that in tablegen

wsmoses: ah yeah... we need to teach tablegen about variadic inputs (it presently thinks it only applies to the first arg)

wsmoses: you might be able to work around it in the time being with a GlobalExpr that just does op.getStartIndices() or something

PR #233, opened 2025-01-11
fix: missing neg in rsqrt derivative

avik-pal:


Pangoraw:
The corresponding test needs to be updated:

https://github.com/EnzymeAD/Enzyme-JAX/blob/4e28cbc8523af07a4f0f47972edd022bc9c91192/test/lit_tests/diffrules/stablehlo/rsqrt.mlir#L19-L29

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -1107,7 +1107,7 @@ def : HLOInactiveOp<"RngBitGeneratorOp">;
 def : HLODerivative<"RsqrtOp", (Op $x),
                     [
                       // (Select (FCmpUEQ $x, (ConstantFP<"0"> $x)), (ConstantFP<"0"> $x), (FDiv (DiffeRet), (FMul (ConstantFP<"2"> $x), (Call<(SameFunc), [ReadNone,NoUnwind]> $x))))
-                      (Div (DiffeRet), (Mul (HLOConstantFP<"2"> $x), (Mul $x, (Sqrt $x))))
+                      (Neg (Div (DiffeRet), (Mul (HLOConstantFP<"2"> $x), (Mul $x, (Sqrt $x)))))
```
wsmoses, line 1110: Minor but can you put the neg on the mul and not on the div? This will enable better constant propagation (since often times differrt will be 1 or 0 or a phi thereof so then it simplifies nicely)

wsmoses: Also actually why not just make the constant 2 into -2?

wsmoses: This also only changes one AD mode, presumably you’ll need to change the 2 to be a -2 in the line above as well

wsmoses (approved):

PR #237, opened 2025-01-14
Possible CI fix for linux?

vimarsh6739:
Based on https://github.com/tensorflow/tensorflow/issues/70199#issuecomment-2282276126

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,16 +40,14 @@ jobs:
         with:
           submodules: recursive
       
-      - run: sudo rm -rf ~/.cache/bazel ~/.cache/bazelisk
+      - run: sudo rm -rf ~/.cache/bazel ~/.cache/bazelisk ~/.bazel/*
 
-      - name: Check compiler env
+      - name: Check compiler env and setup stuff
         run: |
           g++ --version 
           gcc --version
           export CC=/usr/bin/gcc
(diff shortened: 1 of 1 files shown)


vimarsh6739:
well the error is different so there's some progress

.github/workflows/build.yml
```diff
@@ -21,8 +21,18 @@ jobs:
       - name: add llvm
         run: |
             if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+
               sudo apt-get update
-              sudo apt-get install -y git gcc g++ #python3 python3-dev python3-pip
+              sudo apt-get install -y git gcc g++ gcc-12 g++-12 #python3 python3-dev python3-pip
```
wsmoses, line 26: what if you remove the old gcc/g++

vimarsh6739: let me try - I think it was an issue with it not being able to find headers

wsmoses:
just for fun can you do `rm -rf .bazel/*`

vimarsh6739:
are CI instances cached?

vimarsh6739:
also it seems like gcc-12 is already the default, I just deleted a bunch of the update-alternatives

wsmoses:
The machine is persistent (or rather its a cluster of machines which have persistent state), so the fact you ran something earlier changing it, is what happened

PR #259, opened 2025-01-20
Raise powi and absi to the math dialect

### LINKED ISSUE #246: Raise __nv_powi to math dialect
powi takes (f64, i32) -> f64

ftynse:
Closes #246.


wsmoses (approved):
LGTM but we have to wait until LLVM integrate of upstream (otherwise we'll be unable to lower/codegen)

PR #261, opened 2025-01-22
Apply `NoNanAddSubSimplify` pattern to integer unconditionally

### LINKED ISSUE #244: `NoNanAddSubSimplify` can be applied unconditionally to integer inputs

chelini:
fixes #244


chelini:
Are we also interested in quantized types?

wsmoses:
Potentially, but I haven't used them yet -- so I'm also not sure the ebst way to test quantized types

PR #265, opened 2025-01-23
WIP Add various llvm raising passes

ivanradanov:
TODO need to fix some issues with LLVMToMemref, also upstream patch for affinescope handling


ivanradanov:
I use this pipeline
```
enzymexlamlir-opt  --pass-pipeline='builtin.module(convert-llvm-to-cf,convert-llvm-to-arith,canonicalize,enzyme-lift-cf-to-scf,canonicalize,promote-scf-while,canonicalize,convert-llvm-to-arith)'
```

src/enzyme_ad/jax/Passes/LLVMToArith.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    // TODO overflow flags
+    return oneToOneRewrite(op, TargetOp::getOperationName(), operands,
+                           attrConvert.getAttrs(), rewriter);
+  }
+};
+} // namespace
+
+#define AttrConvertOverflowToArith AttrConvertPassThrough
+
```
wsmoses, line 97: minor, but for ease can we merge this with https://github.com/EnzymeAD/Enzyme-JAX/blob/b0663d35bf308b73b9593d9f0f5c3d732eb58ec3/src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp#L426 (that doesn't have cmp, but does have addf)

Issue #270, opened 2025-01-25
Optimization pass segfaults on differentiating matmul of complex arrays


mofeing:
Following EnzymeAD/Reactant.jl#604, it seems like there is some bug on some optimization bass when differentiating matmul of complex arrays.

I'm not sure which optimization pass is the one that crashes but when calling `@code_hlo`:
- `optimize=false` works
- `optimize=:before_enzyme` works
- `optimize=:no_enzyme` works
- `optimize=:after_enzyme` segfaults

```julia
julia> using Reactant, Enzyme
AssertionError("Could not find registered platform with name: \"cuda\". Available platform names are: ")

julia> a = ones(ComplexF64, 2, 2)
2×2 Matrix{ComplexF64}:
 1.0+0.0im  1.0+0.0im
 1.0+0.0im  1.0+0.0im

julia> a_re = Reactant.to_rarray(a)
2×2 ConcreteRArray{ComplexF64, 2}:
 1.0+0.0im  1.0+0.0im
 1.0+0.0im  1.0+0.0im

julia> b = 2.0 * ones(ComplexF64, 2, 2)
2×2 Matrix{ComplexF64}:
 2.0+0.0im  2.0+0.0im
 2.0+0.0im  2.0+0.0im

julia> b_re = Reactant.to_rarray(b)
2×2 ConcreteRArray{ComplexF64, 2}:
 2.0+0.0im  2.0+0.0im
 2.0+0.0im  2.0+0.0im

julia> function f(x,y)
           return Enzyme.gradient(Reverse, *, x, y)
       end
f (generic function with 1 method)

julia> @code_hlo optimize=false f(a_re,b_re)
module {
  func.func private @"Const{typeof(*)}(*)_autodiff"(%arg0: tensor<2x2xcomplex<f64>>, %arg1: tensor<2x2xcomplex<f64>>) -> (tensor<2x2xcomplex<f64>>, tensor<2x2xcomplex<f64>>, tensor<2x2xcomplex<f64>>) {
    %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<2x2xcomplex<f64>>) -> tensor<2x2xcomplex<f64>>
    %1 = stablehlo.transpose %arg1, dims = [1, 0] : (tensor<2x2xcomplex<f64>>) -> tensor<2x2xcomplex<f64>>
[...]

PR #272, opened 2025-01-25
fix cmplx stop

wsmoses:
@mofeing re https://github.com/EnzymeAD/Enzyme-JAX/issues/270 got it in gdb

So basically we were simplifying an add of tensor complex, but that only supported float and otherwise assumed int (which is wrong here since it was complex).

```
(lldb) p op.dump()
%6 = stablehlo.add %cst, %cst_2 : tensor<2x2xcomplex<f64>>
(lldb) p op.getOperand(0)
(mlir::Value) {
  impl = 0x0000600000a23de0
}
(lldb) p op.getOperand(0).dump()
%cst = stablehlo.constant dense<(0.000000e+00,0.000000e+00)> : tensor<2x2xcomplex<f64>>
(lldb) p op.getOperand(1).dump()
%cst_2 = stablehlo.constant dense<(1.000000e+00,0.000000e+00)> : tensor<2x2xcomplex<f64>>
```

For now this PR should stop it from failing, but clearly this should be updated to also constprop complex values

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -2757,7 +2757,7 @@ struct AddSimplify : public OpRewritePattern<mlir::stablehlo::AddOp> {
             op, op.getType(), res.cast<ElementsAttr>());
         return success();
       }
-    } else {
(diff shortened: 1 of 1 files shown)


mofeing:
Great! Thanks!
We can open an issue not to forget that optimization

PR #277, opened 2025-01-26
Use the new pass registration

chelini:
The old is deprecated.


chelini:
@wsmoses can we land this?

PR #286, opened 2025-01-28
Fix bazel build with CF to SCF raising

tyb0807:


wsmoses (commented):
I think I already added those deps

tyb0807:
Alright, reformatted and remove unused deps

PR #288, opened 2025-01-28
fix compilation

chelini:
bazel-out/darwin_arm64-dbg/bin/src/enzyme_ad/jax/Passes/Passes.h.inc:589:21: error: use of undeclared identifier 'ub'


wsmoses:
@tyb0807 seems like clear revert of https://github.com/EnzymeAD/Enzyme-JAX/pull/287

PR #289, opened 2025-01-28
Annotate kernel params

chelini:


src/enzyme_ad/jax/Passes/Passes.td
```diff
@@ -20,6 +20,16 @@ def RemoveDuplicateFuncDefPass : Pass<"remove-duplicate-func-def", "mlir::Module
 
 def PropagateConstantBoundsPass : Pass<"propagate-constant-bounds", "ModuleOp"> {
   let summary = "Propagate constant bounds";
+  let description = [{
+    Propagate constant bounds information:
+    1. thread index 
+    2. Block dimension 
+    3. Block index.
+    Additionally, set all the following attributes for all the kernel pointers:
+    1. align 16
```
wsmoses, line 29: technically I looked it up and we can assume 128 byte aligned: https://github.com/openxla/xla/blob/109a8ff382f2d174ee13504fb4f140a745cb25a2/xla/service/gpu/gpu_constants.h#L41

src/enzyme_ad/jax/Passes/PropagateConstantBound.cpp
```diff
@@ -129,6 +151,28 @@ struct PropagateConstantBoundsPass
                                       gridIdzOp.getOperation());
       });
     });
+
+    auto result = moduleOp->walk([&](enzymexla::KernelCallOp callOp) {
+      auto symbolName = callOp.getFn();
+      auto callee = symTable.lookup<FunctionOpInterface>(symbolName);
+      if (!callee)
+        return WalkResult::advance();
+      MLIRContext *ctx = callee->getContext();
```
wsmoses, line 160: Something I realize here. this (and the above) is wrong. A callee could have multiple callers. You need to conservatively min the dereferencable size over all callers (and same for the range above, though that's a max)

chelini: Mind taking another look? thanks!

src/enzyme_ad/jax/Passes/PropagateConstantBound.cpp
```diff
@@ -34,15 +34,32 @@ struct PropagateConstantBoundsPass
...
+  // to the target. If the target already has a range, we update it by taking
+  // the maximum between the current value and the old value to be conservative.
   static void attachConstantRangeIfConstant(MLIRContext *ctx,
                                             Operation *maybeCst,
                                             Operation *target) {
     APInt intValue;
-    if (matchPattern(maybeCst, m_ConstantInt(&intValue)))
-      target->setAttr("range", LLVM::ConstantRangeAttr::get(
-                                   ctx, 32, 0, intValue.getSExtValue()));
+    if (matchPattern(maybeCst, m_ConstantInt(&intValue))) {
```
wsmoses, line 44: I don't think this is the best way to do things.

If, for example, there is already range metadata (that is weaker), we won't do the update.

I think the thing to do here is to
1) get all kernelcall ops
2) then get all called functions
3) for each called function
    a) loop over all callers, getting the max over all ranges/etc
    b) set the derived values/aliasing/etc

chelini: I did not understand your previous comment then. What do you mean by "weaker range"? Can you provide an example?

wsmoses: Yeah supposed that the existing kernel code before this pass has a range of 1-100000, but all the callers have a range 1-100. We would be taking the max over the previous setting and all the actual callers

src/enzyme_ad/jax/Passes/PropagateConstantBound.cpp
```diff
@@ -34,28 +34,80 @@ struct PropagateConstantBoundsPass
...
-                                   ctx, 32, 0, intValue.getSExtValue()));
+  static int32_t getSizeInBytes(Type ty) {
+    int32_t bitWidth = 0;
+    if (auto inType = dyn_cast<IntegerType>(ty)) {
+      bitWidth = inType.getWidth();
+      if (bitWidth == 1)
+        return 1;
+    }
+    if (auto floatType = dyn_cast<FloatType>(ty))
+      bitWidth = floatType.getWidth();
```
wsmoses, line 45: people also use complex of int/float too

wsmoses: oh shoot, yeah please just add a second PR then xD

PR #291, opened 2025-01-28
feat: simplify select(compare(...)...) to max or min

avik-pal:


Pangoraw:
Won't it change the derivative in some cases? because of the rules:

https://github.com/EnzymeAD/Enzyme-JAX/blob/c38ca3f187ef11de6b2292f3cc55c5eb60530d15/src/enzyme_ad/jax/Implementations/HLODerivatives.td#L1038-L1052

avik-pal:
> Won't it change the derivative in some cases? because of the rules:

For the equality case?

Pangoraw:
> > Won't it change the derivative in some cases? because of the rules:
> 
> For the equality case?

Yes I am thinking of something like this:

```mlir

// -----// IR Dump Before DifferentiateWrapperPass (enzyme-wrap) //----- //
module {
  func.func @main(%arg0: tensor<f32>, %arg1: tensor<f32>) -> (tensor<f32>, tensor<f32>) {
    %0 = stablehlo.compare  GE, %arg0, %arg1 : (tensor<f32>, tensor<f32>) -> tensor<i1>
    %1 = stablehlo.select %0, %arg0, %arg1 : tensor<i1>, tensor<f32>
    %2 = stablehlo.maximum %arg0, %arg1 : tensor<f32>
    return %1, %2 : tensor<f32>, tensor<f32>
  }
}


module {
  func.func @main(%arg0: tensor<f32>, %arg1: tensor<f32>, %arg2: tensor<f32>, %arg3: tensor<f32>) -> (tensor<f32>, tensor<f32>, tensor<f32>, tensor<f32>) {
    %0 = stablehlo.compare  GE, %arg0, %arg2 : (tensor<f32>, tensor<f32>) -> tensor<i1>
    %1 = stablehlo.select %0, %arg1, %arg3 : tensor<i1>, tensor<f32>
    %2 = stablehlo.select %0, %arg0, %arg2 : tensor<i1>, tensor<f32>
    %3 = stablehlo.compare  LT, %arg0, %arg2 : (tensor<f32>, tensor<f32>) -> tensor<i1>
    %4 = stablehlo.select %3, %arg3, %arg1 : tensor<i1>, tensor<f32>
    %5 = stablehlo.maximum %arg0, %arg2 : tensor<f32>
    return %2, %1, %5, %4 : tensor<f32>, tensor<f32>, tensor<f32>, tensor<f32>
  }
}

**Comment by avik-pal** [comment]
I agree. I will revert the equality case

**Comment by wsmoses** [comment]
Honestly I'm fine with this doing the thing on the equality case too.

The derivatives are still equivalent, though computed differently.

Ideally we have some code to fixup and find redundant selections (like above) where if we have
```
x = m > n ; has other use
y = m <= n 
select (y, b, c)
```
with
```
x = m > n ; has other use
select (x, c, b)
```

Aka picking to use the negated condition and swap operands if it gets rid of a conditional

avik-pal:
1 thing to note in the select compare case for equality is that the gradient is uniquely defined. But for Max/MinOp we can return any of the subgradients. In our case, we attribute the gradient to the 2nd argument.

wsmoses:
This is incidentally something already considered by the Enzyme differentiation model (where we guarantee a subgradient but not necessarily a specific subgradient. In particular, due to potential optimizations like this)

wsmoses:
x/ref https://github.com/EnzymeAD/Enzyme.jl/issues/114 i think?

PR #295, opened 2025-01-28
try fixing GPU CI

Pangoraw:


wsmoses (approved):
Lol

Pangoraw:
So even with the flags there is the following error:

```
gcc: error: unrecognized command line option '-mavxvnni'; did you mean '-mavx512vnni'?

**Comment by wsmoses** [comment]
Try adding (--define=xnn_enable_avxvnni=false) too

**Comment by wsmoses** [comment]
@giordano did you see this error before?

**Comment by giordano** [comment]
May have to do with the compiler version. What are you using?

**Comment by giordano** [comment]
It was introduced in [gcc 11](https://gcc.gnu.org/gcc-11/changes.html), so if you're using an older version that'd explain it.

File: .buildkite/gpu_pipeline.yml [hunk 1]
```diff
@@ -39,18 +39,21 @@ steps:
...
-      if [[ "${GCC_VERSION}" -le 12 ]]; then
+      GCC_VERSION=\$(gcc --version | head -1 | awk '{ print \$3 }' | cut -d. -f1)
+      if [[ "\${GCC_VERSION}" -le 12 ]]; then
         # Someone wants to compile some code which requires flags not understood by GCC 12.
         BAZEL_BUILD_FLAGS+=(--define=xnn_enable_avxvnniint8=false)
       fi
-      if [[ "${GCC_VERSION}" -le 11 ]]; then
+      if [[ "\${GCC_VERSION}" -le 11 ]]; then
         # Someone wants to compile some code which requires flags not understood by GCC 11.
+        BAZEL_BUILD_FLAGS+=(--define=xnn_enable_avxvnni=false)
```
giordano, line 49: As mentioned at https://github.com/EnzymeAD/Enzyme-JAX/pull/295#issuecomment-2620321642, this is only for GCC < 10, for 11 you don't need it.  I presume you're using GCC 10 or something.

Pangoraw: ah right, it seems the worker has gcc v9

Pangoraw:
Is it possible to update the build environment on the worker?

wsmoses:
I'm going to merge this as a strict improvement, but nevertheless we should follow up fixing this

PR #296, opened 2025-01-29
Add RemoveIV pattern

tyb0807:
Extracted from #265. Fixed a bug and added test

### DIFF
--- a/src/enzyme_ad/jax/BUILD
+++ b/src/enzyme_ad/jax/BUILD
@@ -121,6 +121,63 @@ gentbl_cc_library(
     tblgen = "//:enzymexlamlir-tblgen",
 )
 
+td_library(
+    name = "RaisingTransformOpsTdFiles",
+    srcs = [
+        "TransformOps/RaisingTransformOps.td",
+    ],
+    deps = [
+        "@llvm-project//mlir:TransformDialectTdFiles",
+    ]
+)
+
+gentbl_cc_library(
+    name = "RaisingTransformOpsIncGen",
+    tbl_outs = [(
+            ["-gen-op-decls"],
+            "TransformOps/RaisingTransformOps.h.inc",
+        ), (
+            ["-gen-op-defs"],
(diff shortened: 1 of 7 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -7557,6 +7559,125 @@ void mlir::transform::addConcatenateOpCanon(RewritePatternSet &patterns,
   patterns.insert<ConcatenateOpCanon>(maxConstantExpansion, &context, benefit);
 }
 
+///////////////  End stablehlo patterns
+
+struct RemoveIVs : public OpRewritePattern<scf::ForOp> {
```
wsmoses, line 7564: we should put this in a separate pass and leave this one for just HLO ops

tyb0807: well the reason why I put it here it's because we have all the infra for writing unit tests for pattern that can be reused.
So the suggestion is to write a pass instead of a pattern here?

tyb0807: Alright I've put out a test transform framework for testing these kinds of patterns

src/enzyme_ad/jax/TransformOps/TestTransformOps.cpp
```diff
@@ -0,0 +1,165 @@
...
+    rewriter.eraseOp(forOp);
+
+    return success();
+  }
+};
+
+} // namespace transform
+} // namespace mlir
+
+#include "src/enzyme_ad/jax/TransformOps/TestTransformPatterns.cpp.inc"
```
wsmoses, line 145: Test is a weird name here, maybe RaiseTransformOps?

tyb0807: Done

wsmoses (approved):

PR #303, opened 2025-02-01
Add LLVM to Affine access pass

tyb0807:
Extracted from #265. Added tests


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -98,4 +98,40 @@ def Pointer2MemrefOp : EnzymeXLA_Op<"pointer2memref", [
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// AtAddrOp
+//===----------------------------------------------------------------------===//
+
```
wsmoses, line 104: We should instead use pointer2memref

test/lit_tests/raising/llvm_to_affine_access.mlir
```diff
@@ -0,0 +1,66 @@
...
+
+  llvm.store %2, %0 {alignment = 1 : i64} : i64, !llvm.ptr<1>
+
+  return
+}
+
+// CHECK-LABEL: func @test_load_store_conversion
+// CHECK-SAME: %[[ARG0:.*]]: !llvm.ptr<1>
+// CHECK-SAME: %[[ARG1:.*]]: i64
+// CHECK: %[[MEMREF:.*]] = "enzymexla.ataddr"(%[[ARG0]]) {{.*}} memref<?xi8, 1>
```
wsmoses, line 16: The original load is typed as i64, can we preserve that here instead of doing an vector of i8?

test/lit_tests/raising/llvm_to_affine_access.mlir
```diff
@@ -0,0 +1,66 @@
...
+// CHECK-DAG: %[[IDX1:.*]] = arith.index_cast %[[ARG1]]
+// CHECK-DAG: %[[IDX2:.*]] = arith.index_cast %[[ARG2]]
+// CHECK: affine.vector_load %[[MEMREF]][symbol(%[[IDX1]]) * 64 + symbol(%[[IDX2]]) * 8] {{.*}} vector<8xi8>
+// CHECK: affine.vector_store
+
+// -----
+
+func.func @test_struct_access(%arg0: !llvm.ptr) {
+  %ptr = llvm.getelementptr %arg0[0, 0] : (!llvm.ptr) -> !llvm.ptr, !llvm.struct<(i64)>
```
wsmoses, line 50: I feel like it is better to apply this pass after converting all of the arguments to memrefs, that way we know the multidimensional sizes and underlying types

test/lit_tests/raising/llvm_to_affine_access.mlir
ivanradanov: My idea to solve this issue was to have custom load op which does 
affine.vector_load_ty %memref[%idx] vector<4xi8> : memref<i8x?>, i64 -> f64

which basically combines the affine.vector_load and the bitcast afterwards in one operation.

Because the current vector load + bitcast approach may lose the information about the original type alignment etc and break things.

test/lit_tests/raising/llvm_to_affine_access.mlir
ivanradanov: Do we ever get multidimensional arrays in our compilation pipeline?

I can see a problem with correctly handling that since all of the loads inside will be with a linear index so e.g. 

```
f(%ptr : !llvm.ptr) {
  ...
  load %j * 4 + %i
  
```
should become this
``` 
f(%ptr : !memref<i8x10x4>) {
  ...
  affine.load[%j][%i]
```
But that is legal only if we can prove that 0 <= %i < 4

Otherwise it needs to be 
```
  affine.load[%j + %i / 4][%i mod 4]
```

I guess we should generate the second one and try to use available information to simplify it to the first one

ivanradanov: We can also run it before converting the args to memref but we should have a simplification/pattern which goes load(pointer2memref(memref2pointer(memref))) to load(memref) and adjusts the indices appropriately.

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -98,4 +98,40 @@ def Pointer2MemrefOp : EnzymeXLA_Op<"pointer2memref", [
...
+        ValueRange(source));
+    }]>,
+    OpBuilder<(ins "Value":$source, "Type":$dstType), [{
+      return build($_builder, $_state,
+        TypeRange(dstType),
+        ValueRange(source));
+    }]>
+  ];
+}
+
```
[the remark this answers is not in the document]
ivanradanov: It says 

> This operation is intended solely as step during lowering, it has no side effects. A reverse operation that creates a memref from an index interpreted as a pointer is explicitly discouraged.

 [here](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefextract_aligned_pointer_as_index-memrefextractalignedpointerasindexop)

test/lit_tests/raising/llvm_to_affine_access.mlir
tyb0807: For TPU I think we'll need to recover the shapes and element types anyway, how hard is it to put that logic into this pass vs. create a separate pass that converts arguments to memref with the correct shapes + raises load/store of arguments to memref.load/store?

ivanradanov:
we merge this for now?
 Only this 
 ```
 //test:bench_vs_xla
 ```
 seems to be failing

wsmoses:
we can ignore the python tests for now, just check //test/lit_tests/...

src/enzyme_ad/jax/Passes/PropagateConstantBound.cpp
```diff
@@ -263,7 +265,105 @@ struct PropagateConstantBoundsPass
           }
         }
       }
+
+      if (!tensor_types)
+        continue;
+
```
wsmoses, line 271: can we separate the memref arg portion out of this PR (e.g. into https://github.com/EnzymeAD/Enzyme-JAX/pull/314), otherwise LGTM and go ahead and merge

ivanradanov:
@tyb0807 I have reverted the memref arg stuff from this pr, you can still find it at b76a15a44923419747bcf2a265fcbd48d0d73f21 if you need it. Sorry for the confusion...

PR #315, opened 2025-02-06
Fix complex import from stablehlo

wsmoses:
@gleasonk temporarily vendoring your code from https://github.com/openxla/stablehlo/pull/2709 per two people ironically having hit an issue on complex solved by it in the past day.

Will replace things to call stablehlo proper once merged and integrated


GleasonK:
> Will replace things to call stablehlo proper once merged and integrated

Sounds good. Sandeep pointed out in code review that it isn't clear that sub-byte types will work great with the `bitWidth / 8` code, so I've updated my patch as well. Should be in shortly.

PR #325, opened 2025-02-08
feat: more passes for scatter

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -7646,6 +7649,77 @@ struct ScatterUpdateComputationConstProp
   };
 };
 
+struct ScatterIndicesAreUnique : public OpRewritePattern<stablehlo::ScatterOp> {
```
wsmoses, line 7652: we also want this for gather too presumably

wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: gather only has indices_are_sorted, not the unique indices attr https://github.com/openxla/stablehlo/blob/main/docs/spec.md#inputs-44

PR #334, opened 2025-02-10
If predicate value propagation

Pangoraw:
All uses of a if predicate which are either in the true or false branch can be replaced with the true or false value accordingly.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -6939,6 +6939,43 @@ struct IfRemoveUnused final : OpRewritePattern<mlir::stablehlo::IfOp> {
   }
 };
 
+struct IfPredPropagation final : OpRewritePattern<mlir::stablehlo::IfOp> {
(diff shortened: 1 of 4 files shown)


wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6939,6 +6939,43 @@ struct IfRemoveUnused final : OpRewritePattern<mlir::stablehlo::IfOp> {
...
+
+    for (OpOperand &use : pred.getUses()) {
+      Operation *owner = use.getOwner();
+      Region *useRegion = owner->getParentRegion();
+
+      if (trueBranch->isAncestor(useRegion)) {
+        anyModified = true;
+        rewriter.modifyOpInPlace(owner, [&]() { use.assign(makeBool(true)); });
+      } else if (falseBranch->isAncestor(useRegion)) {
+        anyModified = true;
```
wsmoses, line 6970: this can also be generalized to check if the negated predicate is in the branch as well (particularly useful for icmp/fcmp where things may not be the same)

PR #335, opened 2025-02-11
feat: batchnorm adjoint

avik-pal:


src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -1234,3 +1234,5 @@ def : HLODerivative<"TransposeOp", (Op $x),
                     ],
                     (SelectIfActive $x, (Transpose (ResultTypes), (Shadow $x), (Permutation)), (HLOConstantFP<"0">))
                   >;
+
+def : HLOInactiveOp<"BatchNormInferenceOp">;
```
wsmoses, line 1238: is this really inactive?

avik-pal: yeah probably not a good idea. more like this being used inside AD is almost always a mistake

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -1840,6 +1840,74 @@ class AutoDiffSort
...
+                                         MGradientUtilsReverse *gutils,
+                                         SmallVector<Value> caches) const {
+    auto op = cast<BatchNormTrainingOp>(orig);
+
+    if (!gutils->isConstantValue(op->getResult(0)) ||
+        !gutils->isConstantValue(op->getResult(1)) ||
+        !gutils->isConstantValue(op->getResult(2))) {
+      gutils->zeroDiffe(op->getResult(0), builder);
+
+      auto opOperand0 = gutils->popCache(caches[0], builder);
```
wsmoses, line 1857: this looks like it ought be doable in tablegen, was there a reason why not?

avik-pal: the 3 gradients are from the same grad op. I couldn't figure out how to do it without inserting 3 separate ops

wsmoses: ah I see, yeah that makes sense. 

we can extend this in the future but this is good now

wsmoses (approved):

wsmoses:
presumably for your follow up use cases we'll also need fwd mode of as well (perhaps with batching)

PR #338, opened 2025-02-13
Also support func.func in memref arg propagation

ivanradanov:


src/enzyme_ad/jax/Passes/LLVMToMemrefAccess.cpp
```diff
@@ -138,45 +138,46 @@ struct LLVMToMemrefAccessPass
 
       // Rewrite function signatures and body operations for all affected
       // kernels
-      SmallVector<Type> newFtyArgs;
+      FunctionType newFuncTy;
```
wsmoses, line 141: explicitly null initialize this?

src/enzyme_ad/jax/Passes/LLVMToMemrefAccess.cpp
```diff
@@ -138,45 +138,46 @@ struct LLVMToMemrefAccessPass
...
 
-          callee->erase();
+        // Copy attributes from the old function to the new one
+        for (auto attr : callee->getAttrs()) {
+          newFunc->setAttr(attr.getName(), attr.getValue());
         }
+        // Update function_type attr
+        newFunc->setAttr("function_type", TypeAttr::get(newFuncTy));
+
+        callee->erase();
```
wsmoses, line 180: so we don't forget why this is here, can you add a quick test?

PR #345, opened 2025-02-15
Do while raising

arpitj1:
Pass to raise doWhile struct in scf.while format to for loop.


src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -908,6 +908,120 @@ struct WhileToForHelper {
...
+        else 
+          return failure();
+      }
+      Value conditionValue = conditionOp.getCondition();
+      Value lowerBound = whileOp.getOperand(0);
+      
+
+      Value upperBound;
+      if (auto cmpOp = conditionValue.getDefiningOp<arith::CmpIOp>()) {
+        arith::CmpIPredicate predicate = cmpOp.getPredicate();
```
ivanradanov, line 954: Do we need to check the predicate to see if it the loop has a positive step or negative step?

test/lit_tests/doWhile.mlir
```diff
@@ -0,0 +1,30 @@
+// RUN: enzymexlamlir-opt --canonicalize-scf-for %s | FileCheck %s
+module {
+  func.func @do_while_example() -> index {
+    %c0 = arith.constant 0 : index
+    %c1 = arith.constant 1 : index
+    %c5 = arith.constant 5 : index
+
+    %result = scf.while (%i = %c0) : (index) -> index {
+      // "Before" region (executes first)
+      %updated = arith.addi %i, %c1 : index
```
ivanradanov, line 10: Can you add a "test.body"(%i) : (index) -> () which will act as the "body" so that we can see the transformation is correct

you will need to add the mlir-opt option for accepting unregistered dialects.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -908,6 +908,120 @@ struct WhileToForHelper {
...
 
+//This works fine for doWhile with no iter_args case only as of now
+//A more proper approach: copy before region to outside for loop, and copy both before and after region to inside for loop ?
+struct MoveDoWhileToFor : public OpRewritePattern<WhileOp> {
+  using OpRewritePattern<WhileOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(WhileOp whileOp,
+                                PatternRewriter &rewriter) const override {
+      auto module = whileOp->getParentOfType<ModuleOp>();
+      // 1. Trace scf.condition to get condition -> condition is arith.cmpi -> From condition get UB.
```
wsmoses, line 919: can you check explicitly here tht there are no iter_args and return failure if so (rather than risk incorrectness/crash)

ivanradanov (changes requested):

test/lit_tests/doWhile.mlir
wsmoses, line 10: can we do a check here with some generic call(%i, %updated) as well to ensure its set up properly?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -908,6 +908,120 @@ struct WhileToForHelper {
...
 };
 
+//This works fine for doWhile with no iter_args case only as of now
+//A more proper approach: copy before region to outside for loop, and copy both before and after region to inside for loop ?
+struct MoveDoWhileToFor : public OpRewritePattern<WhileOp> {
+  using OpRewritePattern<WhileOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(WhileOp whileOp,
+                                PatternRewriter &rewriter) const override {
+      auto module = whileOp->getParentOfType<ModuleOp>();
```
wsmoses, line 918: also do we need this module var?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -689,14 +689,84 @@ struct WhileToForHelper {
   AddIOp addIOp;
   BlockArgument indVar;
   size_t afterArgIdx;
-  bool computeLegality(bool sizeCheck, Value lookThrough = nullptr) {
+
+  bool checkPredicate() {
+    auto cmpRhs = cmpIOp.getRhs();
```
wsmoses, line 694: also cc @Pangoraw for context (this is supporting scf.while -> for for loops whose computation is in the before region), cc'd since you were looking at hlo related earlier

test/lit_tests/canonicalizefor/doWhile.mlir
```diff
@@ -0,0 +1,168 @@
...
+    
+    return %result : index
+  }
+}
+// CHECK-LABEL:   module @simple {
+// CHECK:           func.func @do_while() -> index {
+// CHECK:             %[[VAL_0:.*]] = arith.constant 5 : index
+// CHECK:             %[[VAL_1:.*]] = arith.constant 0 : index
+// CHECK:             %[[VAL_2:.*]] = arith.constant 1 : index
+// CHECK:             scf.for %[[VAL_3:.*]] = %[[VAL_1]] to %[[VAL_0]] step %[[VAL_2]] {
```
wsmoses, line 26: for my sake I wrote this out, and it looks reasonable, but can someone (@ivanradanov or @Pangoraw double check the internal logic)?

the above code runs
```
i = 0
keepalive(i = 0)
updated = i + 1 = 1
if !(updated < 5) break

i = updated = 1
keepalive(i = 1)
updated = i + 1 = 2
if !(updated < 5) break


i = updated = 2
keepalive(i = 2)
updated = i + 1 = 3
if !(updated < 5) break

i = updated = 3
keepalive(i = 3)
updated = i + 1 = 4
if !(updated < 5) break

i = updated = 4
keepalive(i = 4)
updated = i + 1 = 5
if !(updated < 5) break;
  -> breaks
```

ivanradanov:
@arpitj1 The current tests look good
But sadly it doesn't seem to be working for loops with cmpi.ne conditions of which we have a lot in the code.

Can you take a look if you can make it work?
https://github.com/EnzymeAD/Enzyme-JAX/blob/e6985765d9d7be1b016a5b13682641f3ea8152f5/test/lit_tests/canonicalizefor/doWhile.mlir#L187
(I have rebased on main and pushed some changes here, you can take it from here)

wsmoses:
going to merge as is, we can extend in a second PR

arpitj1:
> @arpitj1 The current tests look good But sadly it doesn't seem to be working for loops with cmpi.ne conditions of which we have a lot in the code.
> 
> Can you take a look if you can make it work? I have added a test case here:
> 
> https://github.com/EnzymeAD/Enzyme-JAX/blob/e6985765d9d7be1b016a5b13682641f3ea8152f5/test/lit_tests/canonicalizefor/doWhile.mlir#L187
> 
> 
> (I have rebased on main and pushed some changes here, maybe you can take it from here)

@ivanradanov Created a PR with cmpi.ne support :
https://github.com/EnzymeAD/Enzyme-JAX/pull/372/files#diff-f543ae472c15af82ffba436aa45204e6bd31dc63e16ec4956f0fc875b2173664

PR #348, opened 2025-02-15
Use integer data flow analysis to simplify IR

chelini:


src/enzyme_ad/jax/Passes/Passes.td
```diff
@@ -11,7 +11,7 @@
 
 include "mlir/Pass/PassBase.td"
 
-def CanonicalizeLoopsPass : Pass<"canonicalize-loops"> {
+def CanonicalizeLoopsPass : Pass<"canonicalize-loops", "::mlir::func::FuncOp"> {
```
wsmoses, line 14: Do we need to enforce funcop or can we run more generally (perhaps inside the pass walk all function interfaces)

chelini: yes we can use FuncOpInterface.

src/enzyme_ad/jax/Passes/CanonicalizeLoops.cpp
```diff
@@ -139,6 +148,195 @@ struct RemoveAffineParallelSingleIter
...
+    for (size_t j = 0; j < idx; j++)
+      uoff += uboundGroup[j];
+
+    // Get the constant bounds if available
+    auto lb = lbounds[loff].dyn_cast<AffineConstantExpr>();
+    auto ub = ubounds[uoff].dyn_cast<AffineConstantExpr>();
+
+    if (lb && ub) {
+      // Create APInt values with 32 bits (can adjust width if needed)
+      return {APInt(/*numBits=*/32, lb.getValue(), /*isSigned=*/true),
```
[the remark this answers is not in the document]
chelini: Ack, let me change

wsmoses (approved):

Issue #350, opened 2025-02-15
Crash during convert-polygeist-to-llvm and lower-jit passes


wsmoses:
https://pastebin.com/RnHcKmHr

wsmoses:
even more simpliy
```
module {
  func.func @foo(%673 : f64) -> f64 {
    %674 = math.cbrt %673 : f64
    return %674 : f64
  }
}
```

PR #351, opened 2025-02-16
Use the right builder for CallOp

chelini:
The CallOp lowering pattern was incorrectly passing all operands as a single array to LLVM::CallOp. This caused issues with operand segment sizes since LLVM::CallOp expects the callee and arguments to be passed separately.

This change eliminates the need for manual operand segment size handling since the operation definition now correctly structures the operands.

Fix: #350

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -905,8 +905,8 @@ public:
     }
 
     auto newCallOp = rewriter.create<LLVM::CallOp>(
-        callOp->getLoc(), callResultTypes, adaptor.getOperands(),
-        callOp->getAttrs());
+        callOp->getLoc(), callResultTypes, callOp.getCallee(),
+        adaptor.getOperands());
 
     if (numResults <= 1) {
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
```diff
@@ -905,8 +905,8 @@ struct CallOpLowering : public ConvertOpToLLVMPattern<func::CallOp> {
     }
 
     auto newCallOp = rewriter.create<LLVM::CallOp>(
-        callOp->getLoc(), callResultTypes, adaptor.getOperands(),
-        callOp->getAttrs());
+        callOp->getLoc(), callResultTypes, callOp.getCallee(),
+        adaptor.getOperands());
```
ftynse, line 909: Can we preserve attributes?

ftynse (approved):

ftynse (approved):

src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
chelini: thanks, done.

wsmoses (approved):

PR #352, opened 2025-02-16
fix func->llvm.call lowering

ftynse:
For some reason, one has to manually set operand segment sizes.


chelini:
I did the same :) https://github.com/EnzymeAD/Enzyme-JAX/pull/351

PR #356, opened 2025-02-17
Preserve memory accesses attributes when raising llvm -> affine

tyb0807:


wsmoses (approved):
definitely required (and will merge), but did you manage to reproduce the arg attribute part?

PR #357, opened 2025-02-17
Attempt fix workspace updates

wsmoses:


hhkit:
Hmm, still doesn't work, but at least the error has changed!

```
ERROR: Failed to load Starlark extension '@tsl//third_party/py:python_wheel.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @tsl
This could either mean you have to add the '@tsl' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: Error computing the main repository mapping: cycles detected during computation of main repo mapping
```

hhkit:
I am glad to say that the dependencies are no longer broken! I'm still crawling through the full build if you want a proper sanity check, but for now, at least the immediate failures have been resolved.

PR #358, opened 2025-02-17
[CI] Resurrect GitHub Actions build job

giordano:


giordano:
@wsmoses is this what you had in mind? I deleted the buildkite stuff but it's always recoverable from the history, if necessary, or I can leave it there for the time being.

.github/workflows/build.yml
```diff
@@ -7,54 +8,47 @@ on:
...
-          sudo rm bazel-bin/*.whl || echo
-          HERMETIC_PYTHON_VERSION=${{ matrix.python }} bazel build --define=xnn_enable_avxvnniint8=false --define=xnn_enable_avx512fp16=false :wheel
-          cp bazel-bin/*.whl .
-
-      - name: Upload Build
-        uses: actions/upload-artifact@v4
-        with:
-          name: Wheel-${{ matrix.os }}
-          path: "*.whl"
+          bazel build --color=yes //:enzymexlamlir-opt --test_output=errors
```
wsmoses, line 52: we need to retain the wheel building capabilities as well, if that can be added (in particular the actual python pip package, which while julia isn't using, shouldn't be dropped)

giordano: `bazel build :wheel`? Something else? Can that be a separate step after the first build?

giordano:
https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/13379603495/job/37365706693?pr=358#step:6:143
```
ERROR: /home/runner/.bazel/external/hwloc/BUILD.bazel:230:11: Compiling hwloc/topology.c failed: (Exit 1): gcc failed: error executing command (from target @hwloc//:hwloc) /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections ... (remaining 34 arguments skipped)

external/hwloc/hwloc/topology.c:49:10: fatal error: sys/sysctl.h: No such file or directory
   49 | #include <sys/sysctl.h>
      |          ^~~~~~~~~~~~~~
compilation terminated.
```
I've got this error also locally, there's something wrong with hwloc configuration because this is guarded by `HAVE_SYS_SYSCTL_H`: https://github.com/open-mpi/hwloc/blob/b56eebaa847cc4bdb783a4f6a5198703e64d1d6f/hwloc/topology.c#L48-L50.  The macro seems to be defined incorrectly.

giordano:
Uhm, I suspect this was fixed by https://github.com/openxla/xla/pull/22475.

wsmoses:
we should bump the jax commit in this case

giordano:
:smiley: 

image

wsmoses (approved):
can we leave the crytpic key and also perhaps comment out the buildkite pipeline instead of remove?

otherwise lgtm

giordano:
Done.  So much green now :slightly_smiling_face:

giordano:
I can't merge this, so I'm waiting for you 🙂

PR #367, opened 2025-02-19
Add pass to raise affine loop nests to stablehlo

Pangoraw:


wsmoses (approved):
12/10 amazing.

We're going to need to generalize this with extensions for sin, internal if, internal for, etc

src/enzyme_ad/jax/Passes/AffineRaising.cpp
```diff
@@ -0,0 +1,462 @@
+//===- AffineRaising.cpp - Raise affine loops to stablehlo ----------------===//
```
wsmoses, line 1: can we call this stablehlo-raising instead of affine raising (the latter implies raise to affine)

src/enzyme_ad/jax/Passes/AffineRaising.cpp
```diff
@@ -0,0 +1,462 @@
...
+                                      SmallVectorImpl<int64_t> &strides) {
+  auto rank = accessValueMap.getNumResults();
+
+  startIndices.reserve(rank);
+  limitIndices.reserve(rank);
+  strides.reserve(rank);
+
+  for (unsigned i = 0; i < rank; i++) {
+    auto expr = accessValueMap.getResult(i);
+    if (auto constExpr = dyn_cast<AffineConstantExpr>(expr)) {
```
wsmoses, line 52: this will fail for some more complex things like i+1. A couple of suggestions for how to get around:
For the parallel for loop of interest, you can generate a new expression replacing all dims with something else via https://mlir.llvm.org/doxygen/classmlir_1_1AffineExpr.html#abd8b0817ead1cbe3e6bdb194a292272e

For example, you can set startindex = expr.replaceAllDims(dims -> affineParallel.start indices())

and then if that is a constant, you're good! (since presumably expression replace will do the constant folding within the affine expr for you, which is good).

the other ones may be a bit more tricky though

src/enzyme_ad/jax/Passes/AffineRaising.cpp
Pangoraw: I based the naming on the existing arith-raise pass which raises from arith to stablehlo. I can rename to something like raise-affine-to-stablehlo ?

wsmoses: yeah that would be great (and indeed arith raise is different from the other raising ones we're building since it has multiple different outputs, mhlo vs stablehlo)

src/enzyme_ad/jax/Passes/ArithRaising.cpp
```diff
@@ -43,34 +44,76 @@ struct ArithRaisingPass
...
+    origOp.replaceAllUsesWith(newOp);                                          \
+    origOp.erase();                                                            \
+  });
+
+    RAISE_BINARY(arith::AddFOp, stablehlo::AddOp, mhlo::AddOp);
+    RAISE_BINARY(arith::AddIOp, stablehlo::AddOp, mhlo::AddOp);
+    RAISE_BINARY(arith::SubFOp, stablehlo::SubtractOp, mhlo::SubtractOp);
+    RAISE_BINARY(arith::SubIOp, stablehlo::SubtractOp, mhlo::SubtractOp);
+    RAISE_BINARY(arith::MulFOp, stablehlo::MulOp, mhlo::MulOp);
+    RAISE_BINARY(arith::MulIOp, stablehlo::MulOp, mhlo::MulOp);
```
wsmoses, line 72: this is fine for now but I don't think we need to consider mhlo here long term, just stablehlo is fine

wsmoses (approved):

PR #371, opened 2025-02-20
Use jax.ffi.register_ffi_target instead of deprecated version in jax.lib.xla_client

dfm:
The jax.lib.xla_client submodule has been deprecated, and I'm working on cleaning up users of these APIs. In this case, the `register_custom_call_target` can be directly replaced by [`jax.ffi.register_ffi_target`](https://docs.jax.dev/en/latest/_autosummary/jax.ffi.register_ffi_target.html) with `api_version=0`.

In the long run, AFAIK the XLA team are planning on removing support for the legacy custom call API version, so it would be good to eventually migrate the kernels themselves, but I'll leave that for future work.


dfm:
Fixed formatting and force pushed.

PR #372, opened 2025-02-20
DoWhile to For extension to support cmpi.ne in scf.condition

arpitj1:


src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -667,6 +667,19 @@ bool dominateWhile(Value value, WhileOp loop) {
   }
 }
 
+std::optional<int64_t> getConstantInt(Value operand) {
+  if (operand) {
+    if (auto constOp = operand.getDefiningOp<arith::ConstantOp>()) {
```
wsmoses, line 672: can we use m_ConstantInt instead here?

arpitj1: Added m_ConstantInt

wsmoses (approved):

PR #373, opened 2025-02-22
Simplify If

chelini:
This PR simplifies IF conditions in two cases: 1) The then branch is empty, and the else is not. In this case, we negate the condition and delete the other branch. 2) The if is flattened to select if both regions contain only pure operations. This allows us to remove problematic if mentioned in this conversation: https://github.com/PRONTOLab/GB-25/issues/11#issuecomment-2670386656


ivanradanov (approved):
lgtm

PR #374, opened 2025-02-22
Add missing dependency to `src/enzyme_ad/jax:XLADerivatives`

giordano:
This seems to solve for me locally the error
```
[05:01:02] external/enzyme_ad/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp:19:10: error: module @enzyme_ad//src/enzyme_ad/jax:XLADerivatives does not depend on a module exporting 'mlir/Dialect/Affine/LoopUtils.h'
[05:01:02]    19 | #include "mlir/Dialect/Affine/LoopUtils.h"
[05:01:02]       |          ^
[05:01:02] 1 error generated.
```
As far as I understand, this is defined in https://github.com/llvm/llvm-project/blob/4e44bd027bf4e490380be770172994561616bd2d/utils/bazel/llvm-project-overlay/mlir/BUILD.bazel#L4210-L4239.

Ref: https://github.com/JuliaPackaging/Yggdrasil/pull/10587#issuecomment-2676019451

### DIFF
--- a/src/enzyme_ad/jax/BUILD
+++ b/src/enzyme_ad/jax/BUILD
@@ -419,6 +419,7 @@ cc_library(
         "@llvm-project//mlir:AffineDialect",
         "@llvm-project//mlir:AffineAnalysis",
         "@llvm-project//mlir:AffineTransforms",
+        "@llvm-project//mlir:AffineUtils",
         "@llvm-project//mlir:ArithDialect",
         "@llvm-project//mlir:ArithUtils",
         "@llvm-project//mlir:CallOpInterfaces",


giordano:
I have no idea why

1. CI here (Linux and macOS) was happy
2. CI in Reactant.jl (Linux and macOS) was happy
3. doing a build in BinaryBuilder targeting Linux (but not macOS) platforms was happy
4. I could reproduce on hydra (on Linux) the error reported at https://github.com/JuliaPackaging/Yggdrasil/pull/10587#issuecomment-2676019451 with [`build_local.jl`](https://github.com/EnzymeAD/Reactant.jl/blob/230d77e16a789a95f6c31cab785199a4a05f2227/deps/build_local.jl)

How literally anything works in Bazel will forever remain a mystery to me.

giordano:
Another example of missing dependency not caught by CI: https://github.com/EnzymeAD/Enzyme-JAX/commit/d1ea9972e1f19ca5cc769c1b9841eead62e64e14 😭

PR #375, opened 2025-02-22
Trunci propagation to if-else body

arpitj1:


src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -676,6 +678,68 @@ bool canMoveOpOutsideWhile(Operation *op, WhileOp loop) {
   return true;
 }
 
+class truncProp final : public OpRewritePattern<TruncIOp> {
```
chelini, line 681: out of curiosity why do we need to do this kind of propagation?

wsmoses: the emitted code has this =/

PR #376, opened 2025-02-22
Fix range analysis

wsmoses:


src/enzyme_ad/jax/Passes/CanonicalizeLoops.cpp
```diff
@@ -614,26 +615,184 @@ struct CanonicalizeLoopsPass
...
+                cmpiOp.getLoc(), b.getI1Type(),
+                IntegerAttr::get(b.getI1Type(), false));
+            cmpiOp.getResult().replaceAllUsesWith(cst);
+          }
+        }
+        if (pred == arith::CmpIPredicate::ugt) {
           const APInt umax = cstRange.umax();
           const APInt umin = cstRange.umin();
-          if (umax.ult(cstRhs) && umin.ult(cstRhs)) {
+          if (umax.ugt(cstRhs)) {
```
chelini, line 680: @wsmoses  I am not convinced. I think we should check the min not the max.

wsmoses: Yeah I see what you mean

we should check against https://llvm.org/doxygen/ConstantRange_8cpp_source.html#l00243 for each of these

can you give it a go?

chelini: Ok.

PR #347, opened 2025-02-15
WIP experiment with using isl to simplify affine exprs

ivanradanov:
Work in progress, currently only dimps the isl expressions and their simplified versions. the changes need to be brought back to mlir

An example of a good simplification we achieved:

```
 Got aff
{ [i0, i1, i2, i3, i4, i5] -> [(7 + floor((i0)/256) + floor((96 + i0 + 240*floor((i0)/23) + 112*floor((i0)/184) - 256*floor((i0)/256) + 16*floor((7 + 17i0 - 16*floor((i0)/16) + 6*floor((i0)/23))/374))/2160))] }
Gisted aff
{ [i0, i1, i2, i3, i4, i5] -> [(7 + floor((i0)/184) + floor((i0)/256))] }
```

But we do not manage to simplify everything, for example:

```
Got aff
{ [i0, i1, i2, i3, i4, i5] -> [(7 + floor((i0)/16) + 15*floor((i0)/23) + 7*floor((i0)/184) - 16*floor((i0)/256) + floor((7 + 17i0 - 16*floor((i0)/16) + 6*floor((i0)/23))/374) - 135*floor((112 + i0 + 240*floor((i0)/23) + 112*floor((i0)/184) - 256*floor((i0)/256) + 16*floor((7 + 17i0 - 16*floor((i0)/16) + 6*floor((i0)/23))/374))/2160))] }
Gisted aff
{ [i0, i1, i2, i3, i4, i5] -> [(7 + floor((i0)/16) + 15*floor((i0)/23) + 7*floor((i0)/184) - 16*floor((i0)/256) + floor((7 + 17i0 - 16*floor((i0)/16) + 6*floor((i0)/23))/374) - 135*floor((112 + i0 + 240*floor((i0)/23) + 112*floor((i0)/184) - 256*floor((i0)/256) + 16*floor((7 + 17i0 - 16*floor((i0)/16) + 6*floor((i0)/23))/374))/2160))] }
```

PR #378, opened 2025-02-23
[CI] Run `format` job only on C/C++ files and exclude external packages

giordano:
Ref https://github.com/EnzymeAD/Enzyme-JAX/pull/347#pullrequestreview-2619506969.


giordano:
Ah, this PR made the formatter CI job not run when only externals are touched, but of course when the formatter does run it still tries to format externals because here I didn't modify the settings of the formatter itself.

PR #386, opened 2025-02-24
raising loop iv usage to stablehlo

Pangoraw:


test/lit_tests/raising/affine_to_stablehlo5.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: enzymexlamlir-opt %s --raise-affine-to-stablehlo --arith-raise --enzyme-hlo-opt | FileCheck %s
```
wsmoses, line 1: can we make the parallel bound sufficiently large that this doesn't constprop (and remains an iota)

wsmoses (approved):

PR #398, opened 2025-02-26
Canonicalize pattern for llvm.alloca

arpitj1:
Cannonicalize llvm.alloca and it's users, if the users are limited to - lifetime start, end and memcpy (where alloca is source). This signifies that alloca was never stored too and carries garbage data.


src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
```diff
@@ -403,6 +403,49 @@ struct RemoveFreeze : public OpRewritePattern<LLVM::FreezeOp> {
   }
 };
 
+struct DeadAllocaPattern : public OpRewritePattern<LLVM::AllocaOp> {
```
wsmoses, line 406: can you rename this as ReadOnlyAllocaElim ?

src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
```diff
@@ -403,6 +403,49 @@ struct RemoveFreeze : public OpRewritePattern<LLVM::FreezeOp> {
...
+        deadUsers.push_back(user);
+        continue;
+      }
+
+      if (auto memcpy = dyn_cast<LLVM::MemcpyOp>(user)) {
+        // If copies from allocation, keep it
+        if (memcpy.getDst() == ptr)
+          return failure();
+        // If doesn't copy into allocation keep it
+        if (memcpy.getSrc() != ptr)
```
wsmoses, line 428: we don't need this second check

wsmoses (approved):
looks good with some minor comments

src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
arpitj1: Ah yes

PR #421, opened 2025-03-02
Replace scf.if that yields a boolean with predicated execution

chelini:


test/lit_tests/if-with-empyty-then.mlir
```diff
@@ -35,3 +35,25 @@ func.func @if_with_non_empty_then(%arg0: i1) {
...
+
+// -----
+
+// CHECK-LABEL: func.func @unary_if(
+// CHECK-SAME: %[[ARG0:.*]]: i64, %[[ARG1:.*]]: memref<34x99x194xf64, 1>, %[[ARG2:.*]]: i1, %[[ARG3:.*]]: f64, %[[ARG4:.*]]: f64)
+func.func @unary_if(%arg0: i64, %arg1: memref<34x99x194xf64, 1>, %arg2: i1, %arg3: f64, %arg4: f64) -> i1 {
+  %false = arith.constant false
+  %0 = "enzymexla.memref2pointer"(%arg1) : (memref<34x99x194xf64, 1>) -> !llvm.ptr<1>
+  // CHECK-NOT: scf.if
+  %1 = scf.if %arg2 -> (i1) {
```
wsmoses, line 47: @chelini I don't think this is legal, since llvm.load may not be speculatable without the condition to guard

chelini: Ouch, true.

wsmoses: @arpitj1 was working on a pass to move stuff out of an if, if readnone that might be helpful here without hitting non-speculatable instructions

PR #422, opened 2025-03-02
Auto generate StableHLO transform patterns

GleasonK:
Still has a few issues:

- Since patterns also depends on StableHLO Dialect, need to find a way to filter by only transform ops, I believe op filters exist, just ran out of time :)
- Should add a way to make pattern groups, similar to `opDocGroup` so that each patterns isn't its own transform op
- <other>


wsmoses:
just to make sure the names are what are expected, can we add them to https://github.com/EnzymeAD/Enzyme-JAX/blob/26356cc5ffaf111541b9fadb2da8ac23c3741182/src/enzyme_ad/jax/primitives.py#L127 ?

src/enzyme_ad/jax/Passes/StablehloOptPatterns.td
```diff
@@ -18,44 +39,48 @@ include "stablehlo/dialect/StablehloOps.td"
...
+          (OpTypeA:$opA
+            (OpTypeB:$opB $lhsB, (StableHLO_ConstantOp:$rhsB $valueB)),
+            (StableHLO_ConstantOp:$rhsA $valueA)),
+          (ResOpTypeA $lhsB, (ResOpTypeB $rhsA, $rhsB))>;
 
 // Naming convention: `<RootOp><ChildOp>Const`:
 
 // Pattern: (x / cst_1) * cst_2 -> x * (cst_2 / cst_1)
-def MulDivConst : BinOpLiftConstantComputationFlip<StableHLO_MulOp, StableHLO_DivOp>;
+def MulDivConst : BinOpLiftConstantComputationFlip<"mul_div_const", StableHLO_MulOp, StableHLO_DivOp>;
```
GleasonK, line 65: How useful is it having a ton of individual patterns as transform ops?

I'm thinking a few possible outcomes here -- we offer individual patterns _and_ grouped pattern ops for transform (i.e. {`mul_div_const`, `mul_mul_const`,...} and `binop_const`) -- or maybe this is something we resolve in the python layer with something like aliases?

I guess the real question is -- is there any issue with exploding the number of transform extension ops we're providing here?

I'd like to expand this to all StableHLO optim patterns so we can think about using some of the upstream ones (and upstreaming new ones) without impacting usability:
https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/optimization/StablehloAggressiveSimplificationPatterns.td

cc @ftynse @wsmoses

wsmoses: So part of the original motivation for having the names was debugging both performance and issues.

For example, one can excise specific patterns to find bugs like https://github.com/EnzymeAD/Enzyme-JAX/pull/424.

For performance it gives a search space since some patterns aren't always beneficial in practice (though from the hlo side it may seem strictly simplifying).

I don't think this means we need to have all things as individual patterns, but we can group things where convenient.

ftynse: The alternative would be to have one `stablehlo_patterns` op that is further parameterized by a list of patterns to include/exclude. But that means having (yet another) registration mechanism for that list whereas now we can use the op registration mechanism. Having logical groupings to reduce clutter makes sense to me as a short-term trade-off. Longer-term, we need to find a good way to manage this with e.g., adding ops on the fly or somehow referring to pattern name/location without creating a whole new op every time.

GleasonK: Sounds good, agree. I think a few groupings are probably best in the short term, might remove the individual granularity and just use the `StablehloPatternGroup` from this PR.

I like the file locality of this approach in the meantime to keep things trivial for users contributing patterns / also lets you refer to identifiers in TD files for slightly-better-than-stringly-typed safety. Also will make it really easy to give existing StableHLO patterns bindings in transform dialect

PR #425, opened 2025-03-03
feat: add side_effect and backend_config to jit_call

avik-pal:


wsmoses (approved):

test/lit_tests/lowering/cpujit.mlir
```diff
@@ -36,6 +36,6 @@ module {
 // CHECK-LABEL: @main
 // CHECK-SAME: (%[[ARG0:.+]]: tensor<64xi64>) -> tensor<64xi64> {
 // CHECK-NEXT:    %[[CALL:.+]] = stablehlo.custom_call @enzymexla_compile_cpu(%arg0) 
-// CHECK-SAME: {api_version = 3 : i32, backend_config = "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00", 
+// CHECK-SAME: {api_version = 4 : i32, backend_config = {attr = "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00"},
```
avik-pal, line 39: @wsmoses this PR does make this change and I am not sure it affects the kernel stuff in any way

src/enzyme_ad/jax/Passes/LowerJIT.cpp
```diff
@@ -991,13 +998,12 @@ struct LowerJITPass
...
             op.getLoc(), op.getResultTypes(), op.getInputs(),
             rewriter.getStringAttr("enzymexla_compile_cpu"),
-            /* has_side_effect*/ rewriter.getBoolAttr(false),
-            /*backend_config*/ backendstr,
+            /* has_side_effect*/ op.getHasSideEffectAttr(),
+            /*backend_config*/ dattr,
             /* api_version*/
             CustomCallApiVersionAttr::get(
                 rewriter.getContext(),
-                mlir::stablehlo::CustomCallApiVersion::
```
wsmoses, line 999: oh no this is actually an issue, this needs to keep its API as it was before (it is the calling convention for the setup)

test/lit_tests/lowering/cpujit.mlir
wsmoses: no, this actually does need to stay the same (see above)

src/enzyme_ad/jax/Passes/LowerJIT.cpp
avik-pal: `API_VERSION_STATUS_RETURNING_UNIFIED` doesn't allow for a DictAttr

wsmoses: yeah, but we shouldn't change the attributes passed in

avik-pal: if we pass backend_config to jit_call those need to be forwarded to the custom_call as well right?

wsmoses: so currently the backend config out of jitcall is the function pointer of the jitt'd function, which then is directly executed

wsmoses:
> Just to make it clear, tests here are failing, independently of the failure introduced by #426, so please don't merge this PR before _both_ issues are fixed.

yeah that later one should be fixed by https://github.com/EnzymeAD/Enzyme-JAX/pull/427

@giordano do we have a branch protection rule for the things expected to pass

giordano:
The 3 "Build Enzyme-JAX" jobs are all expected to pass, like the code formatting ones. I see there are 3 expected jobs which don't exist anymore, those can be removed.

test/lit_tests/jit_call.mlir
```diff
@@ -0,0 +1,17 @@
+// RUN: enzymexlamlir-opt %s --pass-pipeline="builtin.module(lower-jit{jit=false backend=cpu},enzyme-hlo-opt)" | FileCheck %s
+
+module {
+  llvm.func internal unnamed_addr fastcc @throw() attributes {dso_local, no_inline, sym_visibility = "private"} {
+    llvm.unreachable
+  }
+
+  // CHECK-LABEL: func.func @main(%arg0: tensor<4xf32>) -> tensor<4xf32> {
+  func.func @main(%arg0: tensor<4xf32>) -> tensor<4xf32> {
+    // CHECK: stablehlo.custom_call @enzymexla_compile_cpu() {api_version = 3 : i32, backend_config = "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00", has_side_effect = true} : () -> ()
```
mofeing, line 10: shouldn't the `backend_config` remain the same and get passed to the underlying func?

avik-pal: this needs fixing, the backend config needs to be compiled as part of CompileCall and this backend config should just be the attribute (at least that is what I figured out from the discussion above)

PR #369, opened 2025-02-19
[GHA] Also run tests, again

giordano:
Add a step to always refresh the requirements file.

PR #426, opened 2025-03-03
split induction variable if some uses are not floordiv or mod

Pangoraw:


src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -2852,8 +2944,8 @@ struct SplitParallelInductions
               if (i < AI.getIntegerSet().getNumDims()) {
                 functionOf |= E.isFunctionOfDim(i);
               } else {
-                functionOf |= E.isFunctionOfSymbol(
-                    i - AI.getIntegerSet().getNumSymbols());
+                functionOf |=
+                    E.isFunctionOfSymbol(i - AI.getIntegerSet().getNumDims());
```
wsmoses, line 2948: if we goofed this here, can we also fix it for mergeparallelinductions too?

wsmoses (approved):

Pangoraw:
I think my math is wrong for the reverse part

giordano:
Tests are failing!

wsmoses:
investigating, sorry we really need to get to the state where tests were expected to pass [I'll do that now]

wsmoses:
@Pangoraw we need to confirm there's at least one mod, to do this (thus it purely simplifies). on it

giordano:
> sorry we really need to get to the state where tests were expected to pass [I'll do that now]

They were supposed to pass before this PR 🙂 unfortunately we run tests only on macos because they fail anyway on Linux: https://github.com/EnzymeAD/Enzyme-JAX/pull/369#issuecomment-2671686651. Also, on hydra neuralgcm tests segfaults for me.

PR #429, opened 2025-03-04
vendor affine loop parallelization from upstream in affine-cfg

ftynse:
This is a partially reduced copy of the affine-parallelize pass from upstream that doesn't limit the number of nested parallel loops and always handles reductions.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -1,5 +1,6 @@
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/Passes.h"
+#include "mlir/Dialect/Affine/Utils.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -3666,6 +3678,28 @@ struct AddAddCstEnd : public OpRewritePattern<arith::AddIOp> {
...
 };
 
+static void greedilyParallelize(Operation *op) {
+  // Whether to parallelize reductions.
+  constexpr bool parallelReductions = true;
+
+  // The walker proceeds in pre-order to process the outer loops first
+  // and control the number of outer parallel loops.
+  std::vector<ParallelizationCandidate> parallelizableLoops;
+  op->walk<WalkOrder::PreOrder>([&](AffineForOp loop) {
```
wsmoses, line 3688: we may want to run this before the affinecfg stuff [or ideally as a pattern], since the parallel fusion, and index rewrite is a pattern [otherwise run before parallelize]

ftynse: it's not initially written is a pattern, so I suspect it may mess up the internal state of the rewriter; let me see what I can do

wsmoses (approved):

PR #430, opened 2025-03-04
If Yield propagate to outside if-else region

arpitj1:


wsmoses (approved):
lgtm but @chelini can you also give this a review?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+      SmallVector<Operation *> uniqueOpsToMoveAfterIfVector(
+          uniqueOpsToMoveAfterIfMap.begin(), uniqueOpsToMoveAfterIfMap.end());
+      for (auto op : llvm::reverse(uniqueOpsToMoveAfterIfVector)) {
+        rewriter.clone(*op, mappingAfterIf);
+      }
+    }
+
+    // Replace uses of the original if operation with the new one
+    int pos = 0;
+    for (auto value : originalYieldsNewValues) {
```
chelini, line 2966: you can use llvm::enumerate, something like:
```
auto [idx, value] : llvm::enumerate(originalYieldsNewValues) {

}
```
And you can drop `int pos = 0`

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+                  opsToMoveAfterIf, ifYieldOperands, elseYieldOperands,
+                  rewriter);
+    index++;
+  }
+}
+
+// Algorithm:
+// 1. Extract yield operations from both regions
+// 2. Check if yield operations match in if and else region
+// 3. If match, move them outside the loop and recursively check their operands
```
chelini, line 2789: Are they actually moved below the if correct?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+  }
+  if (isa<BlockArgument>(operandIf) || isa<BlockArgument>(operandElse) ||
+      !ifOp.getThenRegion().isAncestor(
+          operandIf.getDefiningOp()->getParentRegion()) ||
+      !ifOp.getElseRegion().isAncestor(
+          operandElse.getDefiningOp()->getParentRegion()) ||
+      !(uniqueThenUsers.size() == 1 && uniqueElseUsers.size() == 1) ||
+      (operandIf.getDefiningOp()->getName() !=
+       operandElse.getDefiningOp()->getName()) ||
+      !isReadNone(operandIf.getDefiningOp()) ||
```
chelini, line 2764: maybe I don't understand the impl for isReadNone, but I am quite surprised we check for Read memory effect. Why are we not checking for write and alloca memory effects?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+    yieldUseMap.push_back(std::make_tuple(
+        ifYieldOperands.size() - 1, opsToMoveAfterIf.back(), opOperandIndex));
+    return;
+  }
+
+  Operation *opToMove = operandIf.getDefiningOp();
+  opsToMoveAfterIf.push_back(opToMove);
+  int index = 0;
+  for (auto [thenOperand, elseOperand] :
+       llvm::zip(operandIf.getDefiningOp()->getOperands(),
```
chelini, line 2777: llvm::zip_equal.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+    ifYieldOperands.push_back(operandIf);
+    elseYieldOperands.push_back(operandElse);
+    yieldUseMap.push_back(std::make_tuple(
+        ifYieldOperands.size() - 1, opsToMoveAfterIf.back(), opOperandIndex));
+    return;
+  }
+
+  Operation *opToMove = operandIf.getDefiningOp();
+  opsToMoveAfterIf.push_back(opToMove);
+  int index = 0;
```
chelini, line 2775: you can also use llvm::enumerate to remove this variable. I left an example below.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+// Algorithm:
+// 1. Extract yield operations from both regions
+// 2. Check if yield operations match in if and else region
+// 3. If match, move them outside the loop and recursively check their operands
+// 4. If don't match then continue with the next operand
+// 5. Create a new if operation with updated yields
+// 6. Updated source operands of operations moved outside to the new yields
+// 7. Replace uses of the original if operation with the new one
+struct IfYieldMovementPattern : public OpRewritePattern<scf::IfOp> {
+  IfYieldMovementPattern(mlir::MLIRContext *context)
```
chelini, line 2795: Can we simply use: `using OpRewritePattern<scf::IfOp>::OpRewritePattern;` ?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+    auto elseYield =
+        cast<scf::YieldOp>(ifOp.getElseRegion().front().getTerminator());
+
+    // List of yield positions not modified
+    std::vector<Value> originalYieldsNewValues;
+
+    // Use SetVector to ensure uniqueness while preserving order
+    SmallVector<Value> ifYieldOperands, elseYieldOperands;
+    SmallVector<Operation *> opsToMoveAfterIf;
+    SmallVector<std::tuple<unsigned long, Operation *, unsigned long>>
```
chelini, line 2816: I would use size_t here instead of unsigned long

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+         llvm::zip(thenYield.getOperands(), elseYield.getOperands())) {
+
+      llvm::SmallPtrSet<Operation *, 4> uniqueThenUsers, uniqueElseUsers;
+      for (Operation *user : thenOperand.getUsers()) {
+        uniqueThenUsers.insert(user);
+      }
+      for (Operation *user : elseOperand.getUsers()) {
+        uniqueElseUsers.insert(user);
+      }
+      if (isa<BlockArgument>(thenOperand) || isa<BlockArgument>(elseOperand) ||
```
chelini, line 2830: mind adding a comment here. The if looks rather complex :)

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -2738,27 +2738,276 @@ struct SelectI1Simplify : public OpRewritePattern<arith::SelectOp> {
...
+          !ifOp.getThenRegion().isAncestor(
+              thenOperand.getDefiningOp()->getParentRegion()) ||
+          !ifOp.getElseRegion().isAncestor(
+              elseOperand.getDefiningOp()->getParentRegion()) ||
+          !(uniqueThenUsers.size() == 1 && uniqueElseUsers.size() == 1) ||
+          (thenOperand.getDefiningOp()->getName() !=
```
chelini, line 2836: Do we take into consideration that the operands of the op should be the same?

arpitj1: I think we can still move the op even if the operands are not same. We just have to put the operands into yield from both if and else.

PR #434, opened 2025-03-05
Raise inner parallel for loops and simple affine.if

Pangoraw:


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -634,6 +672,28 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
...
+
+    auto iv = forOp.getInductionVar();
+    auto range = getIVRange(iv);
+    if (!range.has_value())
+      return failure();
+
+    emitIVToStableHLO(builder, iv, *range, mapping, maps);
+
+    for (auto &innerOp : forOp.getBody()->without_terminator()) {
+      if (tryRaisingOpToStableHLO(&innerOp, mapping, builder, maps).failed())
```
wsmoses, line 688: I'm not sure this is legal, since this assumes that the affine.for iterations are independent/can be run in parallel?

Shouldn't an affinefor here [unless parallel] be raised to a stablehlo.while loop?

wsmoses: I mean just more simply consider the following

pfor arg1
  for arg2
     x[arg1] *= 2

wsmoses: we're working on trying to autoparallelize inner affine.for's into affine.parallel's. once this is in we can add support for inner parallel's too, which I think would permit this : https://github.com/EnzymeAD/Enzyme-JAX/pull/429

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -634,6 +792,180 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
...
+        map = alignMemoryAccess(cond, map, newCond, outputMap, builder);
+        cond = builder.create<stablehlo::AndOp>(ifOp.getLoc(), cond, newCond);
+      } else {
+        cond = newCond;
+        map = outputMap;
+      }
+    }
+
+    for (auto &innerOp : ifOp.getThenBlock()->without_terminator()) {
+      if (tryRaisingOpToStableHLO(&innerOp, mapping, builder, maps).failed())
```
wsmoses, line 839: we need an extra check here to confirm that the inner op is speculatable [i.e. won't go out of bounds/etc]

wsmoses (approved):

wsmoses:
missed the ispure check in the code, merging

PR #435, opened 2025-03-05
CI: Run lit tests on linux

Pangoraw:
testing this one

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,6 +88,10 @@ jobs:
         if: ${{ !startsWith(matrix.os, 'ubuntu-') }}
         run: |
           bazel test --color=yes --test_output=errors --test_verbose_timeout_warnings //test/...
+      - name: Run lit tests
+        if: ${{ startsWith(matrix.os, 'ubuntu-') }}
+        run: |
+          bazel test --color=yes --test_output=errors --test_verbose_timeout_warnings $(find test/ -name '*.mlir' | awk '{print $1 ".test"}')
       - name: Upload test results
         uses: actions/upload-artifact@v4
         if: ${{ !cancelled() }} 


giordano:
They're bound to fail: https://github.com/EnzymeAD/Enzyme-JAX/pull/369#issuecomment-2671686651

giordano:
Tangentially related, [I'm testing using Python 3.11](https://github.com/giordano/Enzyme-JAX/actions/runs/13674288508/job/38231232719) to solve some issues (installing tensorflow with python 3.12 fails because of distutils being removed).  That should solve a bunch of issues, but on hydra I still have segfaults in some tests (as I had mentioned in https://github.com/EnzymeAD/Enzyme-JAX/pull/426#issuecomment-2695684674 already)

giordano:
> I am filtering to run only the lit tests //lit_tests/*

Ah, testing only those may work, yes.

giordano:
> Tangentially related, [I'm testing using Python 3.11](https://github.com/giordano/Enzyme-JAX/actions/runs/13674288508/job/38231232719) to solve some issues

For the record, can confirm that with Python 3.11 packaging issues are solved, but two tests still fail:

* [`//test:keras_test` times out](https://github.com/giordano/Enzyme-JAX/actions/runs/13674288508/job/38231232273#step:9:313)
* [`//test:neuralgcm_test` segfaults](https://github.com/giordano/Enzyme-JAX/actions/runs/13674288508/job/38231232273#step:9:460), as I also experienced locally

If we can exclude these two tests we can run all the rest (no idea how it's easy to do filter out few tests with Bazel)

Pangoraw:
That's nice we can replace the filter with something like

```
bazel test ... $(bazel query //test/... | grep -v "keras_test" | grep -v "neuralgcm_test" | grep -v "llama")
```

PR #441, opened 2025-03-08
Add code for unrolling affine.for under threshold

snarang181:
Add corresponding test

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -896,6 +896,35 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
     return success();
   }
 
+  if (auto forOp = dyn_cast<affine::AffineForOp>(op)) {
+    if (!forOp.hasConstantBounds())
+      return failure(); // We want to unroll the loop, so we need constant
+                        // bounds.
+
+    Value iv = forOp.getInductionVar();
+    InductionVariableRange range = *getIVRange(iv);
+    if (range.getNumIters() <= 50) { // Tune this threshold for unrolling,
+                                     // otherwise make it stablehlo.while
+      OpBuilder::InsertionGuard guard(builder);
+      builder.setInsertionPoint(forOp);
+
+      for (int64_t ivValue = range.lb; ivValue < range.ub;
+           ivValue += range.step) {
+        Value ivConst = builder.create<stablehlo::ConstantOp>(
+            forOp.getLoc(), builder.getI64TensorAttr(ivValue));
+
+        mapping.map(iv, ivConst);
+
+        for (auto &innerOp : forOp.getBody()->without_terminator()) {
+          if (tryRaisingOpToStableHLO(&innerOp, mapping, builder, maps)
+                  .failed())
+            return failure();
+        }
+      }
+      return success();
+    }
+  }
+
   // // Inner for op
   // if (auto forOp = dyn_cast<affine::AffineForOp>(op)) {
   //   if (!forOp.hasConstantBounds())
--- a/test/lit_tests/raising/raiseaffinefor_unroll.mlir
+++ b/test/lit_tests/raising/raiseaffinefor_unroll.mlir
@@ -1,0 +1,23 @@
+// RUN: enzymexlamlir-opt %s --raise-affine-to-stablehlo
+#set = affine_set<(d0) : (-d0 + 89 >= 0)>
+module {
+    func.func private @"##call__Z31gpu__fill_south_and_north_halo_16CompilerMetadataI16OffsetStaticSizeI14_1_180__21_21_E12DynamicCheckvv7NDRangeILi2E10StaticSizeI6_1__1_ES4_I8_180__1_E5TupleI5Int64S8_E13KernelOffsetsIS9_EEE11OffsetArrayI7Float64Li3E13CuTracedArrayISF_Li3ELi1E13_194__187__1_EE17BoundaryConditionI4FluxvESJ_I6ZipperS8_ES7_I6CenterSO_4FaceE20ImmersedBoundaryGridISF_8Periodic14RightConnected7Bounded28OrthogonalSphericalShellGridISF_SS_ST_SU_28StaticVerticalDiscretizationISE_ISF_Li1ESG_ISF_Li1ELi1E5_35__EESE_ISF_Li1ESG_ISF_Li1ELi1E5_34__EESY_S10_ESE_ISF_Li2ESG_ISF_Li2ELi1E10_194__187_EE8TripolarIS8_S8_S8_EvE16GridFittedBottomI5FieldISO_SO_vvvvSI_SF_vvvE23CenterImmersedConditionEvvvES7_I10NamedTupleI53__time___last__t___last_stage__t___iteration___stage_S7_ISF_SF_SF_S8_S8_EES1D_I36__u___v___w_______U___V___T___S___e_S7_ISE_ISF_Li3ESG_ISF_Li3ELi1E13_194__99__34_EES1H_SE_ISF_Li3ESG_ISF_Li3ELi1E13_194__99__35_EESI_S18_ISP_SO_vvvvSI_SF_vvvES18_ISO_SP_vvvvSI_SF_vvvES1H_S1H_S1H_EEE#814$par16"(%arg0: memref<1x187x194xf64, 1>) {
+        affine.parallel (%arg1) = (0) to (180) {
+        %0 = affine.load %arg0[0, 51, %arg1 + 7] : memref<1x187x194xf64, 1>
+        affine.store %0, %arg0[0, 50, %arg1 + 7] : memref<1x187x194xf64, 1>
+        affine.for %arg2 = 0 to 50 {
+            %4 = affine.load %arg0[0, -%arg2 + 134, -%arg1 + 186] : memref<1x187x194xf64, 1>
+            affine.store %4, %arg0[0, %arg2 + 136, %arg1 + 7] : memref<1x187x194xf64, 1>
+        }
+        %1 = affine.load %arg0[0, 135, -%arg1 + 186] : memref<1x187x194xf64, 1>
+        %2 = affine.load %arg0[0, 135, %arg1 + 7] : memref<1x187x194xf64, 1>
+        %3 = affine.if #set(%arg1) -> f64 {
+            affine.yield %2 : f64
+        } else {
+            affine.yield %1 : f64
+        }
+        affine.store %3, %arg0[0, 135, %arg1 + 7] : memref<1x187x194xf64, 1>
+        }
+        return
+    }
+}


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -896,6 +896,35 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
...
   }
 
+  if (auto forOp = dyn_cast<affine::AffineForOp>(op)) {
+    if (!forOp.hasConstantBounds())
+      return failure(); // We want to unroll the loop, so we need constant
+                        // bounds.
+
+    Value iv = forOp.getInductionVar();
+    InductionVariableRange range = *getIVRange(iv);
+    if (range.getNumIters() <= 50) { // Tune this threshold for unrolling,
```
wsmoses, line 906: we shouldn't have a threshold here, otherwise we fail to raise at all.

We should always succeed, and potentially can run an unrolling pass elsewhere.

Ideally this is raised to a stablehlo.while of corresponding induction variables.

That said unrolling it fully is fine as a first step (but we should really prefer the former here)

snarang181:
When I run `gdb --args bazel-bin/enzymexlamlir-opt --raise-affine-to-stablehlo Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir`, I see
```
/home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:10:13: error: 'stablehlo.dynamic_update_slice' op operation destroyed but still has uses
            affine.store %4, %arg0[0, %arg2 + 136, %arg1 + 7] : memref<1x187x194xf64, 1>
            ^
/home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:10:13: note: see current operation: %0 = "stablehlo.dynamic_update_slice"(<<NULL VALUE>>, <<NULL VALUE>>, <<NULL VALUE>>, <<NULL VALUE>>, <<NULL VALUE>>) : (<<NULL TYPE>>, <<NULL TYPE>>, <<NULL TYPE>>, <<NULL TYPE>>, <<NULL TYPE>>) -> tensor<1x187x194xf64>
/home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:19:9: note: - use: %39 = "stablehlo.dynamic_update_slice"(<<UNKNOWN SSA VALUE>>, %38, %34, %35, %36) : (tensor<1x187x194xf64>, tensor<1x1x180xf64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<1x187x194xf64>

        affine.store %3, %arg0[0, 135, %arg1 + 7] : memref<1x187x194xf64, 1>
        ^
/home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:13:14: note: - use: %17 = "stablehlo.slice"(<<UNKNOWN SSA VALUE>>) <{limit_indices = array<i64: 1, 136, 187>, start_indices = array<i64: 0, 135, 7>, strides = array<i64: 1, 1, 1>}> : (tensor<1x187x194xf64>) -> tensor<1x1x180xf64>

        %2 = affine.load %arg0[0, 135, %arg1 + 7] : memref<1x187x194xf64, 1>
             ^
/home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:12:14: note: - use: %14 = "stablehlo.slice"(<<UNKNOWN SSA VALUE>>) <{limit_indices = array<i64: 1, 136, 187>, start_indices = array<i64: 0, 135, 7>, strides = array<i64: 1, 1, 1>}> : (tensor<1x187x194xf64>) -> tensor<1x1x180xf64>

        %1 = affine.load %arg0[0, 135, -%arg1 + 186] : memref<1x187x194xf64, 1>
             ^
LLVM ERROR: operation destroyed but still has uses
```
This is presumably something to do with the use of innerOp in the for-loop. Do I need to use something like `replaceAllUsesWith`?

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
snarang181: That sounds good, I will remove the threshold.

wsmoses: sg, open an issue for raising this to a proper stablehlo.while so we don't forget

wsmoses:
> When I run `gdb --args bazel-bin/enzymexlamlir-opt --raise-affine-to-stablehlo Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir`, I see
> 
> ```
> /home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:10:13: error: 'stablehlo.dynamic_update_slice' op operation destroyed but still has uses
>             affine.store %4, %arg0[0, %arg2 + 136, %arg1 + 7] : memref<1x187x194xf64, 1>
>             ^
> /home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:10:13: note: see current operation: %0 = "stablehlo.dynamic_update_slice"(<<NULL VALUE>>, <<NULL VALUE>>, <<NULL VALUE>>, <<NULL VALUE>>, <<NULL VALUE>>) : (<<NULL TYPE>>, <<NULL TYPE>>, <<NULL TYPE>>, <<NULL TYPE>>, <<NULL TYPE>>) -> tensor<1x187x194xf64>
> /home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:19:9: note: - use: %39 = "stablehlo.dynamic_update_slice"(<<UNKNOWN SSA VALUE>>, %38, %34, %35, %36) : (tensor<1x187x194xf64>, tensor<1x1x180xf64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<1x187x194xf64>
> 
>         affine.store %3, %arg0[0, 135, %arg1 + 7] : memref<1x187x194xf64, 1>
>         ^
> /home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:13:14: note: - use: %17 = "stablehlo.slice"(<<UNKNOWN SSA VALUE>>) <{limit_indices = array<i64: 1, 136, 187>, start_indices = array<i64: 0, 135, 7>, strides = array<i64: 1, 1, 1>}> : (tensor<1x187x194xf64>) -> tensor<1x1x180xf64>
> 
>         %2 = affine.load %arg0[0, 135, %arg1 + 7] : memref<1x187x194xf64, 1>
>              ^
> /home/snarang181/Enzyme-JAX/test/lit_tests/raising/raiseaffinefor_unroll.mlir:12:14: note: - use: %14 = "stablehlo.slice"(<<UNKNOWN SSA VALUE>>) <{limit_indices = array<i64: 1, 136, 187>, start_indices = array<i64: 0, 135, 7>, strides = array<i64: 1, 1, 1>}> : (tensor<1x187x194xf64>) -> tensor<1x1x180xf64>
> 
>         %1 = affine.load %arg0[0, 135, -%arg1 + 186] : memref<1x187x194xf64, 1>
>              ^
> LLVM ERROR: operation destroyed but still has uses
> ```
> 
> This is presumably something to do with the use of innerOp in the for-loop. Do I need to use something like `replaceAllUsesWith`?

@Pangoraw since you said you had something like this earlier, can you take a quick look?

also cc @chelini who might have cycles to help co debug

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
snarang181: Done -- https://github.com/EnzymeAD/Enzyme-JAX/issues/442

Pangoraw:
I ran the affine LICM and affine unrolling passes before invoking raise-affine-to-stablehlo

ivanradanov:
@snarang181 
I think the problem was that you were generating the stablehlo ops at the for op (which is in the parallel which we delete later), see the last commit. Does that look right?

snarang181:
> @snarang181 I think the problem was that you were generating the stablehlo ops at the for op (which is in the parallel which we delete later), see the last commit. Does that look right?

Thanks for looking into this @ivanradanov; and yes, that makes sense.

snarang181:
This seems to work fine on the test case but is giving a core dumped error on the full unoptimized mlir (running fullpipe.sh).  Here is the error snippet 

```
enzymexlamlir-opt: src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp:282: mlir::affine::AffineValueMap alignMemoryAccess(mlir::Value&, mlir::affine::AffineValueMap, mlir::Value*, llvm::ArrayRef<mlir::affine::AffineValueMap>, mlir::OpBuilder&): Assertion `shapeA.size() == cast<RankedTensorType>(a.getType()).getShape().size()' failed
```
I am happy to debug further, some co-debugging help to get me on the right track would be appreciated though.

wsmoses:
Probably I’d start by adding a print before each function being raised to make a minimal test case

snarang181:
> Probably I’d start by adding a print before each function being raised to make a minimal test case

Do I need to build with some args to get the LLVM_DEBUG to show on stdout?

wsmoses:
You would add -debug for that.

however here I’m suggesting adding llvm::errs() << fn << “\n”; at an appropriate point inside the pass, recompiling, and running (which requires no extra flag)

snarang181:
> I get this: 
> 
> ```
> 
> enzymexlamlir-opt: src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp:295: mlir::affine::AffineValueMap alignMemoryAccess(mlir::Value&, mlir::affine::AffineValueMap, mlir::Value*, llvm::ArrayRef<mlir::affine::AffineValueMap>, mlir::OpBuilder&): Assertion `shapeBs[i].size() == cast<RankedTensorType>(bs[i].getType()).getShape().size()' failed.
> 
> ```
> 
> 
> 
> (not 100% sure) but it seems to be because [here](https://github.com/EnzymeAD/Enzyme-JAX/blob/599f8459487434f67d1940f0fd208ccd1573649e/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp#L723) we emit a null AffineMap for a constant. I am not entirely sure what the correct one would be though
> 
> 
> 
> ```
> 
> frame #7: 0x00000000073b4e46 enzymexlamlir-opt`tryRaisingOpToStableHLO(op=0x00000000123d31b0, mapping=0x00007fffffffb210, builder=0x00007fffffffb1d0, maps=0x00007fffffffb1b0) at AffineToStableHLORaising.cpp:784:65                                                             
> 
>    781            b = mapping.lookup(op->getOperand(1));                                                                                                                                                                                                                          
> 
>    782                                                                                                                                                                                                                                                                            
> 
>    783      auto mapA = maps[a], mapB = maps[b];                                                                                                                                                                                                                                  
> 
> -> 784      auto outputMap = alignMemoryAccess(a, mapA, b, mapB, builder);                                                                                                                                                                                                        
> 
>    785      assert(a.getType() == b.getType());                                                                                                                                                                                                                                   
> 
>    786                                                                                                                                                                                                                                                                            
> 
>    787      auto IT = a.getType().cast<RankedTensorType>();                                                                                                                                                                                                                       
> 
> (lldb) p a.dump()                                                                                                                                                                                                                                                                 
> 
> %770 = "stablehlo.constant"() <{value = dense<0> : tensor<1xi64>}> : () -> tensor<1xi64>                                                                                                                                                                                          
> 
> ```
> 
> 
> 
> I am off for today but will continue investigating further tomorrow.

@ivanradanov, that's right. Feel free to ping me on slack when you get online your tomorrow and I'd like to co-debug if possible to get a feel for things. Thanks again for looking into it.

wsmoses:
cc @Pangoraw

wsmoses:
can you add the failing test here as a test file for ease?

Pangoraw:
The current approach is missing handling the converted IV as a constant when raising the memref load/store. In the current situation I guess the load/store would result in slices over all possible values. 

Solving this problem would probably be a first step into making the loops stablehlo.while since the iv would no longer be a constant but a block argument. 

In the meantime, we can run the upstream affine unrolling pass?

Pangoraw:
> (not 100% sure) but it seems to be because [here](https://github.com/EnzymeAD/Enzyme-JAX/blob/599f8459487434f67d1940f0fd208ccd1573649e/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp#L723) we emit a null AffineMap for a constant. I am not entirely sure what the correct one would be though


So these maps are used to identify for each dim in the raised tensors by which induction variable it is dependent. A scalar constant is raised to a 0-dim tensor and therefore as a map with zero dimensions.

For example the following value:

```mlir
%0 = affine.load %a[%i, %j]
```

would have a map like `(%i, %j) -> (%i, %j)`. The maps are then used to align the different axes via broadcast/transpose when two values need to interact together to produce a resulting value.

wsmoses:
resolved by https://github.com/EnzymeAD/Enzyme-JAX/pull/457

PR #443, opened 2025-03-08
Allow raising ifs with speculatively executable loads

ivanradanov:


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -505,6 +513,46 @@ expandAffineExpr(OpBuilder &builder, Location loc, AffineExpr expr,
...
+  isl_set *array = ia.getMemrefShape(ty);
+  llvm::errs() << "ARRAY\n";
+  isl_set_dump(array);
+
+  auto accessMap = ia.getAccessMap(op);
+  if (!accessMap)
+    return false;
+
+  llvm::errs() << "ACCESSMAP\n";
+  isl_map_dump(accessMap);
```
wsmoses, line 538: @ivanradanov if there is an inner affine.for would this still work or potential crash?

we definitely want to exclude any if conditions, but we do need to ensure we have all the index vars

ivanradanov: Currently the use of this function is limited to only this case:

```
`scope` -> affine.if {
             ...
`op` ->      affine.load
```
so no indexes will be gone. I guess I will add a dominance check to make sure we don't misuse it in the future

wsmoses:
```
MEMREFSHAPE
{ [i0] }
MEMREFSHAPE
{ [i0] : i0 >= 0 and i0 <= 193 }
ARRAY
{ [i0] : i0 >= 0 and i0 <= 193 }
{ [dim] -> [o0] }
({ [dim] -> [(3 + dim)] })
ACCESSMAP
{ [dim] -> [o0] : o0 = 3 + dim }
DOMAIN
{ [i0] : i0 >= 0 and i0 <= 190 }
ACCESSED
{ [i0] : i0 >= 3 and i0 <= 193 }
INBOUNDS
1
src/external/isl/isl_ctx.c:305: isl_ctx not freed as some objects still reference it
```

ivanradanov:
@wsmoses did you try the latest commit? that should have removed the debug prints and fixed the memory handling

wsmoses:
yeah I did [just running the test that failed CI, though in debug]

ivanradanov:
Test should be fixed now, checking how it works on the full code now.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -796,9 +842,12 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
   if (auto ifOp = dyn_cast<affine::AffineIfOp>(op)) {
     if (!ifOp.hasElse() || ifOp->getNumResults() == 0 ||
         llvm::any_of(*ifOp.getThenBlock(),
-                     [](Operation &op) { return !mlir::isPure(&op); }) ||
-        llvm::any_of(*ifOp.getElseBlock(),
-                     [](Operation &op) { return !mlir::isPure(&op); })) {
+                     [ifOp](Operation &op) {
+                       return !isSafeToSpeculativelyExecuteAtScope(ifOp, &op);
```
wsmoses, line 846: I'm a little worried about correctness here.

Specifically if you have an if of an if.

We want to use the scope of the outermost if (assuming that the outer if recursively calls this utility)

wsmoses (approved):
lgtm modulo if of if concerns

wsmoses:
will let the if concerns be addressed post commit @ivanradanov

PR #447, opened 2025-03-10
Split affine.parallels such that we no longer need a nested affine.if

ivanradanov:


ivanradanov:
Regarding phase ordering. In theory we would like something like this:
```
func.func(affine-cfg{enable-parallel-if-split=true}),func.func(llvm-to-affine-access),...,affine-to-stablehlo,func.func(affine-cfg{enable-parallel-if-split-false}),affine-to-stablehlo
```
I still haven't introduced the `enable-parallel-if-split` option

the idea is that whatever we can raise to stablehlo with affine.ifs intact we should do so. (llvm-to-affine-access will only split the ifs that it needs to enable raising of llvm.{store,load}) 

Then for the remaining things which we werent able to raise , we want to split the parallel if iteration space for everything we can so that we can raise as much as possible to stablehlo

Not sure if stablehlo supports non-rectangular iteration spaces or tensors, @Pangoraw any input here?

cc @Pangoraw @wsmoses

PR #448, opened 2025-03-10
Ensure strict affine expr checking

vimarsh6739:
Multiplication of 2 dimensions ((d1 + 40) * (d0)) was being treated as an `affine.store` / `affine.load`, which is obviously wrong. 
I'm thinking of raising it in the fallback case to a `memref.store` / `memref.load`

### DIFF
--- a/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
@@ -494,7 +494,7 @@ struct AffineExprBuilder {
       default:
         LLVM_DEBUG(llvm::dbgs()
                    << "Unknown affine binary expression" << bOpExpr << "\n");
-      } 
+      }
     }
     // const, dim or symbol
     return true;


src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -466,6 +466,40 @@ struct AffineExprBuilder {
...
+
+    if (AffineBinaryOpExpr bOpExpr = dyn_cast<AffineBinaryOpExpr>(expr)) {
+      bool ret_flag = false;
+      switch (bOpExpr.getKind()) {
+      case mlir::AffineExprKind::Add:
+        // do nothing
+        break;
+      case mlir::AffineExprKind::Mul: {
+        bool lhsConst = isa<AffineConstantExpr>(bOpExpr.getLHS());
+        bool rhsConst = isa<AffineConstantExpr>(bOpExpr.getRHS());
```
Pangoraw, line 479: nit (though I don't think this is relevant for GB) 
```suggestion
        bool lhsConst = bOpExpr.getLHS().isSymbolicOrConstant();
        bool rhsConst = bOpExpr.getRHS().isSymbolicOrConstant();
```

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -466,6 +466,40 @@ struct AffineExprBuilder {
...
+
+        // either LHS or RHS has to be a constant
+        if (!lhsConst && !rhsConst)
```
wsmoses, line 482: the logic here needs to be more complex to be correct

for example, will accidentally fail if you have, say (s1 + 1) * (s2 + 2), since the lhs/rhs aren't symbols / constants

in any case, we can revisit this PR post GB

wsmoses (changes requested):

ivanradanov:
I thought there is already a isPurelyAffine or something like that function in AffineExpr, can we use that?

vimarsh6739:
Hmm, the implementation here seems to check it correctly here - https://github.com/llvm/llvm-project/blob/c8ec8073aa5d8e87a15d101ded149de399518bc1/mlir/lib/IR/AffineExpr.cpp#L212 (I didn't notice this, so thanks for pointing to this!)

Although I believe we can skip the recursive check here because we are building the expressions recursively already here.

PR #450, opened 2025-03-10
Raise memref.load to stablehlo.gather

Pangoraw:


test/lit_tests/raising/affine_to_stablehlo13.mlir
```diff
@@ -0,0 +1,73 @@
+// RUN: enzymexlamlir-opt %s --raise-affine-to-stablehlo --canonicalize --arith-raise --enzyme-hlo-opt | FileCheck %s
```
wsmoses, line 1: maybe so we can tell what's happening, disable the constant propagation beyond size 0?

wsmoses (approved):

PR #454, opened 2025-03-10
Raise `llvm.load/store` to `memref.load/store`

vimarsh6739:
Does this as a fallback in the case:
1. we cannot promote to an affine load/store
2. The defining op for the load/store is a `GEP` whose base is defined by `enzymexla.memref2pointer` (but I believe it is flattened for access)

### DIFF
--- a/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
@@ -331,7 +331,7 @@ static MemRefVal convertToMemref(PtrVal addr) {
           MemRefLayoutAttrInterface{}, origMemrefType.getMemorySpace());
 
       auto ptr2memref = builder.create<enzymexla::Pointer2MemrefOp>(
-          addr.getLoc(), collapsedType, addr);
+          addr.getLoc(), collapsedType, basePtr);
       return cast<MemRefVal>(ptr2memref.getResult());
     }
   }
--- a/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
@@ -59,6 +59,7 @@
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <utility>
 
 #define DEBUG_TYPE "llvm-to-affine-access"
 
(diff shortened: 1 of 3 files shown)


vimarsh6739:
I'll add a smaller lit test to reproduce the behaviour

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -1072,8 +1109,10 @@ convertLLVMToAffineAccess(Operation *op,
     AffineAccessBuilder &aab = *accessBuilders.back();
     auto dl = dataLayoutAnalysis.getAtOrAbove(op);
     auto res = aab.build(dl, addr);
-    if (failed(res))
+    if (failed(res)) {
       accessBuilders.pop_back();
+      remainingMemOps.push_back(std::make_pair(op, addr));
+    }
```
wsmoses, line 1115: instead of storing this list.

Can you just do a second walk of all store/loadops (after doing the conversion to affine where legal)?

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -300,6 +301,40 @@ static MemRefVal convertToMemref(PtrVal addr) {
...
+  if (auto gep = addr.getDefiningOp<LLVM::GEPOp>()) {
+    auto basePtr = gep.getBase();
+
+    // if basePtr is from a memref, then emit a static shaped Pointer2memref
+    if (auto m2p = basePtr.getDefiningOp<enzymexla::Memref2PointerOp>()) {
+      // get the original memref + type
+      Value origMemref = m2p.getSource();
+      auto origMemrefType = cast<MemRefType>(origMemref.getType());
+
+      // collapse dimensions
```
wsmoses, line 317: see the delinearize pass which I imagine is the right place to add support for multidim of memref load/store

vimarsh6739: Let me take a look - but is there a case where collapsing dimensions wouldn't work? I'm assuming it's when you have multiple GEP indices (which I haven't seen in Oceananigans)

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -1217,6 +1256,122 @@ convertLLVMToAffineAccess(Operation *op,
...
+                         load.getLoc(),
+                         MemRefType::get(memrefTy.getShape(), ty,
+                                         MemRefLayoutAttrInterface{},
+                                         memrefTy.getMemorySpace()),
+                         memref)
+                     .getResult();
+      }
+
+      SmallVector<Value> indices;
+      // If this is a 1D memref and we have GEP indices, use them
```
wsmoses, line 1295: we shouldn't limit to 1d memrefs

wsmoses: instead we should have 0, 0, 0, idx as the indices

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -1217,6 +1256,122 @@ convertLLVMToAffineAccess(Operation *op,
...
+      if (memrefTy.getRank() == 1 && addr.getDefiningOp<LLVM::GEPOp>()) {
+        auto gepOp = addr.getDefiningOp<LLVM::GEPOp>();
+        if (!gepOp.getIndices().empty()) {
+          auto gepIndex = gepOp.getIndices()[0];
+          Value indexVal;
+          if (auto intAttr = dyn_cast<IntegerAttr>(gepIndex)) {
+            indexVal = rewriter.create<arith::ConstantOp>(
```
wsmoses, line 1302: we don't need to special case constant or not

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
wsmoses: there's a distinction from the memref being 1d and the gep being 1d. We should support memref of any size, its fine to assume the gep is 1d for now

src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
vimarsh6739: I think there is a need to here - `convertToIndex` needs a `Value` and `getIndices()` returns `array[Union{value,intAttr}]`

wsmoses: Ah I forgot Alex and I added the constant attribute not just value, yeah that’s right we need it

PR #457, opened 2025-03-10
Raise affine.for to stablehlo.while

Pangoraw:
Using scatter/gather for loads since the indices are now dynamic.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -145,19 +145,9 @@ computeExprRange(affine::AffineValueMap map, AffineExpr expr) {
       range.step = rangeDyn->step;
       break;
     case AffineExprKind::Mul:
-      // %i = (0) to (180) step 1
-      // -%i = (-0) to (-180) (step -1)
-      if (const_ < 0) {
-        // (i) in (lb) to (ub)
-        // (-i) in (-ub) to (-lb) (step -1)
-        range.lb = rangeDyn->lb * const_;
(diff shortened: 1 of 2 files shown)


wsmoses:
On reflection, I don't think we need gather here. Can we use instead https://github.com/openxla/stablehlo/blob/main/docs/spec.md#dynamic_slice

wsmoses (approved):
lgtm

we should also potentially do something similar for the other gather/scatters as well, when legal

wsmoses:
@Pangoraw in post commit review, can you check that this works (or fails) on a case with affine.for iter_args ?

snarang181:
```
module {
  func.func @main(%arg0: memref<4x10xf32>, %arg1: memref<16x10xf32>) {

    affine.parallel (%i) = (0) to (10) {
      %init = affine.load %arg1[0, %i] : memref<16x10xf32>  
      affine.for %j = 0 to 4 iter_args(%accum = %init) -> (f32) {
        %0 = affine.load %arg0[%j, %i] : memref<4x10xf32>
        %1 = arith.mulf %0, %0 : f32
        %2 = arith.addf %1, %accum : f32  
        affine.store %1, %arg1[4 * %j, %i] : memref<16x10xf32>
        affine.yield %2 : f32
      }
    }
    return
  }
}
```

> @Pangoraw in post commit review, can you check that this works (or fails) on a case with affine.for iter_args ?

I wrote this test case and I am seeing `enzymexlamlir-opt: external/llvm-project/mlir/include/mlir/IR/IRMapping.h:74: auto mlir::IRMapping::lookup(T) const [with T = mlir::Value]: Assertion `result && "expected 'from' to be contained within the map"' failed.` *(Assuming this is valid syntax on the test case)

PR #464, opened 2025-03-11
feat: add side-effect to jit_call

avik-pal:
Only the side_effect attr part from https://github.com/EnzymeAD/Enzyme-JAX/pull/425/

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -61,7 +61,7 @@ def KernelCallOp: EnzymeXLA_Op<"kernel_call", [DeclareOpInterfaceMethods<SymbolU
   let hasCanonicalizer = 1;
 }
 
-def JITCallOp: EnzymeXLA_Op<"jit_call", [DeclareOpInterfaceMethods<SymbolUserOpInterface>, DeclareOpInterfaceMethods<CallOpInterface>, Pure]> {
+def JITCallOp: EnzymeXLA_Op<"jit_call", [DeclareOpInterfaceMethods<SymbolUserOpInterface>, DeclareOpInterfaceMethods<CallOpInterface>, DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
   let summary = "JIT Call operation";
 
   let arguments = (ins
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -65,8 +65,9 @@ def JITCallOp: EnzymeXLA_Op<"jit_call", [DeclareOpInterfaceMethods<SymbolUserOpI
   let summary = "JIT Call operation";
 
   let arguments = (ins
-    FlatSymbolRefAttr:$fn, 
+    FlatSymbolRefAttr:$fn,
     Variadic<AnyType>:$inputs,
+    DefaultValuedOptionalAttr<BoolAttr, "false">:$has_side_effect,
     DefaultValuedStrAttr<StrAttr, "">:$backend_config,
     OptionalAttr<AnyAttr>:$operand_layouts,
     OptionalAttr<AnyAttr>:$result_layouts,
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -67,7 +67,7 @@ def JITCallOp: EnzymeXLA_Op<"jit_call", [DeclareOpInterfaceMethods<SymbolUserOpI
   let arguments = (ins
     FlatSymbolRefAttr:$fn,
     Variadic<AnyType>:$inputs,
-    DefaultValuedOptionalAttr<BoolAttr, "false">:$has_side_effect,
+    OptionalAttr<UnitAttr>:$has_side_effect,
     DefaultValuedStrAttr<StrAttr, "">:$backend_config,
     OptionalAttr<AnyAttr>:$operand_layouts,
     OptionalAttr<AnyAttr>:$result_layouts,
--- a/src/enzyme_ad/jax/Dialect/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Ops.cpp
@@ -123,6 +123,19 @@ Attribute JITCallOp::removeArgAttrsAttr() { return nullptr; }
 
 Attribute JITCallOp::removeResAttrsAttr() { return nullptr; }
 
+// Follows that of stablehlo.custom_call
+void JITCallOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  auto hasSideEffect = getHasSideEffectAttr();
+  if (hasSideEffect && !hasSideEffect.getValue())
+    return;
+  effects.emplace_back(MemoryEffects::Allocate::get());
+  effects.emplace_back(MemoryEffects::Free::get());
+  effects.emplace_back(MemoryEffects::Write::get());
+  effects.emplace_back(MemoryEffects::Read::get());
+}
+
 /// Replace cast(subindex(x, InterimType), FinalType) with subindex(x,
 /// FinalType)
 template <typename OpTy>
--- a/src/enzyme_ad/jax/Dialect/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Ops.cpp
@@ -371,31 +371,6 @@ public:
   }
 };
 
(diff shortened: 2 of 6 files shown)


mofeing:
great! thanks @avik-pal

wsmoses (approved):
can you doubole check that a side effecting op isn't removed by canonicalize. we may need to remove the pure marking (and add custom side effects, cc @ftynse for pointers on)

mofeing:
> can you doubole check that a side effecting op isn't removed by canonicalize.

@avik-pal if you need some real MLIR for testing this, you can use the MLIR of calling `MPI.Barrier`

```julia
julia> @code_hlo optimize=false MPI.Barrier(MPI.COMM_WORLD)
module @reactant_Barrier attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  llvm.mlir.global external constant @MPI_COMM_WORLD() {addr_space = 0 : i32} : !llvm.ptr
  llvm.func @MPI_Barrier(!llvm.ptr) -> i32
  func.func @enzymexla_wrapper_MPI_Barrier() {
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.call @MPI_Barrier(%0) : (!llvm.ptr) -> i32
    return
  }
  func.func @main() {
    enzymexla.jit_call @enzymexla_wrapper_MPI_Barrier () : () -> ()
    return
  }
}
```

without `side_effect`, the `jit_call` gets deleted
```julia
julia> @code_hlo optimize=true MPI.Barrier(MPI.COMM_WORLD)
module @reactant_Barrier attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @enzymexla_wrapper_MPI_Barrier() {
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.call @MPI_Barrier(%0) : (!llvm.ptr) -> i32
    return
  }
  llvm.mlir.global external constant @MPI_COMM_WORLD() {addr_space = 0 : i32, sym_visibility = "private"} : !llvm.ptr
  llvm.func @MPI_Barrier(!llvm.ptr) -> i32 attributes {sym_visibility = "private"}
  func.func @main() {
    return
  }
}
```

avik-pal:
I added the MemoryEffectsOpInterface, but now 

```mlir
// RUN: enzymexlamlir-opt %s --canonicalize | FileCheck %s

module @nosideeffect {
  llvm.mlir.global external constant @MPI_COMM_WORLD() {addr_space = 0 : i32} : !llvm.ptr
  llvm.func @MPI_Barrier(!llvm.ptr) -> i32
  func.func @enzymexla_wrapper_MPI_Barrier() {
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.call @MPI_Barrier(%0) : (!llvm.ptr) -> i32
    return
  }
  func.func @main() {
    enzymexla.jit_call @enzymexla_wrapper_MPI_Barrier () : () -> ()
    return
  }
}

module @sideeffect {
  llvm.mlir.global external constant @MPI_COMM_WORLD() {addr_space = 0 : i32} : !llvm.ptr
  llvm.func @MPI_Barrier(!llvm.ptr) -> i32
  func.func @enzymexla_wrapper_MPI_Barrier() {
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.call @MPI_Barrier(%0) : (!llvm.ptr) -> i32
    return
  }
  func.func @main() {
    enzymexla.jit_call @enzymexla_wrapper_MPI_Barrier () {has_side_effect = true} : () -> ()
    return
  }
}
```

neither of the calls are being removed

wsmoses:
hm cc @ftynse

src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -123,6 +123,19 @@ Attribute JITCallOp::removeArgAttrsAttr() { return nullptr; }
...
 Attribute JITCallOp::removeResAttrsAttr() { return nullptr; }
 
+// Follows that of stablehlo.custom_call
+void JITCallOp::getEffects(
+    SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
+        &effects) {
+  auto hasSideEffect = getHasSideEffectAttr();
+  if (hasSideEffect && !hasSideEffect.getValue())
+    return;
+  effects.emplace_back(MemoryEffects::Allocate::get());
```
wsmoses, line 133: we shouldn't do it like this (it otherwise says that the call always allocates/read/writes/reads). We probably want to recurse into the callee, if available and get the effects from within

wsmoses: we may also want to add a custom folder as well

src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -356,14 +370,46 @@ class ReadNoneArg final : public OpRewritePattern<OpTy> {
   }
 };
 
+class RemoveUnusedJitCall final : public OpRewritePattern<JITCallOp> {
```
wsmoses, line 373: In this specific case, I would recommend adding a folder: https://mlir.llvm.org/docs/Canonicalization/#canonicalizing-with-the-fold-method

avik-pal: I am not sure how to remove the op with a folder. "Note that this mechanism cannot be used to remove 0-result operations."

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -61,12 +61,13 @@ def KernelCallOp: EnzymeXLA_Op<"kernel_call", [DeclareOpInterfaceMethods<SymbolU
...
 
-def JITCallOp: EnzymeXLA_Op<"jit_call", [DeclareOpInterfaceMethods<SymbolUserOpInterface>, DeclareOpInterfaceMethods<CallOpInterface>, Pure]> {
+def JITCallOp: EnzymeXLA_Op<"jit_call", [DeclareOpInterfaceMethods<SymbolUserOpInterface>, DeclareOpInterfaceMethods<CallOpInterface>, DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
   let summary = "JIT Call operation";
 
   let arguments = (ins
-    FlatSymbolRefAttr:$fn, 
+    FlatSymbolRefAttr:$fn,
     Variadic<AnyType>:$inputs,
+    DefaultValuedOptionalAttr<BoolAttr, "false">:$has_side_effect,
```
ftynse, line 70: Nit: you can use `UnitAttr` the presence of which indicates "true" and the absence "false" instead to avoid the annoying ternary state.

src/enzyme_ad/jax/Dialect/Ops.cpp
ftynse: I'm more worried about having a specification on the _caller_ that the callee has no side effects. I'd rather have the callee have an attribute that we can check here. We could then have an analysis that adds such attributes on all functions (reusing from Enzyme-MLIR work) in a topological way and avoid recursive/expensive jumping over the call graph.

src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -356,14 +370,46 @@ class ReadNoneArg final : public OpRewritePattern<OpTy> {
...
+    if (results.size() == 0) {
+      rewriter.eraseOp(op);
+      return success();
+    }
+
+    bool anyResultUsed = false;
+    for (auto result : results) {
+      if (!result.use_empty()) {
+        anyResultUsed = true;
+        break;
```
ftynse, line 392: Nit: we can just `return failure()` here.

ftynse:
> hm cc @ftynse

The logic removing non side-effecting ops in the canonicalizer is driven by the `isOpTriviallyDead` function, check that out. AFAIR, it may ignore allocate and read effects, but shouldn't ignore writes, for example. Also, with `-debug-only=greedy-rewrier`, you should get logging on what gets removed.

src/enzyme_ad/jax/Dialect/Ops.cpp
avik-pal: If has_side_effect is not specified, we can check the callee and get the memory effects from https://github.com/EnzymeAD/Enzyme-JAX/pull/876

ftynse: I think we should always check the callee. The dangerous situation I see that the caller says "no side effects, trust me bro" while being wrong. This may also happen on the function itself if the body changes without having the attributes updated, but at least is easier to debug (and we can have an expensive verifier).

avik-pal:
https://github.com/EnzymeAD/Enzyme-JAX/pull/876 has the full implementation now. We don't need any special attribute, we are inferring side effects from the MemoryEffectsInterface

PR #480, opened 2025-03-14
Support iter args for affine.for -> stablehlo

ivanradanov:


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -547,24 +547,89 @@ bool isSafeToSpeculativelyExecuteAtScope(Operation *scope, Operation *op) {
...
+    if (CTT.getRank() != 0)
+      return std::nullopt;
+    SmallVector<int64_t> dimsToBroadcast;
+    auto bc = b.create<stablehlo::BroadcastInDimOp>(v.getLoc(), TT, v,
+                                                    dimsToBroadcast);
+
+    AffineMap newMap = AffineMap::getMultiDimIdentityMap(
+        TT.getRank() - CTT.getRank(), b.getContext());
+
+    return Broadcast{bc, affine::AffineValueMap(newMap, ivs)};
```
ivanradanov, line 591: @Pangoraw I need your input here. We use this function to "expand" any tensor that is incoming as an initial value for an affine.for iter_arg to the iteration space of the parallel we are in. I currently only handle the case where it already exactly matches what we need, or it is a scalar. I don't know what else we can encounter wrt to the incoming shapes and maps in here. 

Would you be able to help with this implementation?

(see the test cases for reference for cases we can handle)

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1523,21 +1609,30 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
...
+                .failed())
           return failure();
       }
 
       Value newIvInBody =
           builder.create<stablehlo::AddOp>(forOp.getLoc(), ivInBody, step);
 
       SmallVector<Value> loopCarried = {newIvInBody};
+      for (auto iterArg : forOp.getRegionIterArgs())
+        loopCarried.push_back(mapping.lookup(iterArg));
```
Pangoraw, line 1622: we need to map the yielded values here (also we can check that `maps[mapping.lookup(iterARgs)].getOperands() == maps[mapping.lookup(yieldedIterARgs)].getOperands()` to make sure the hypothesis that every it is dependent on every iv in the loop nest is true):

```suggestion
      for (auto [iterArg, yieldedIterArgs] : llvm::zip(
        forOp.getRegionIterArgs(),
        forOp.getBody()->getTerminator()->getOperands()))
          loopCarried.push_back(mapping.lookup(yieldedIterArgs));

```

test/lit_tests/raising/raise-affine-to-stablehlo-affine-for-iter-args3.mlir
```diff
@@ -0,0 +1,34 @@
...
+// CHECK:           %[[VAL_3:.*]] = stablehlo.constant dense<18> : tensor<i64>
+// CHECK:           %[[VAL_4:.*]] = stablehlo.constant dense<0> : tensor<i64>
+// CHECK:           %[[VAL_5:.*]] = stablehlo.constant dense<5.000000e-01> : tensor<85x180xf64>
+// CHECK:           %[[VAL_6:.*]]:2 = stablehlo.while(%[[VAL_7:.*]] = %[[VAL_4]], %[[VAL_8:.*]] = %[[VAL_0]]) : tensor<i64>, tensor<85x180x18xf64>
+// CHECK:            cond {
+// CHECK:             %[[VAL_9:.*]] = stablehlo.compare  LT, %[[VAL_7]], %[[VAL_3]] : (tensor<i64>, tensor<i64>) -> tensor<i1>
+// CHECK:             stablehlo.return %[[VAL_9]] : tensor<i1>
+// CHECK:           } do {
+// CHECK:             %[[VAL_10:.*]] = stablehlo.add %[[VAL_7]], %[[VAL_2]] : tensor<i64>
+// CHECK:             stablehlo.return %[[VAL_10]], %[[VAL_8]] : tensor<i64>, tensor<85x180x18xf64>
```
Pangoraw, line 31: with mapping the yielded values we should expect that it is not val_8 to be returned but `%add = stablehlo.add %val_8, %ld`

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
Pangoraw: That looks good, I think we can start with this case of only supporting constants and see where that gets us. If that's not sufficient we can still broadcast the multiple init values with the right broadcast dims.

Pangoraw (approved):
Fixed a bug in the load with a for loop iv and updated the test.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1523,21 +1610,41 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
...
 
       Value newIvInBody =
           builder.create<stablehlo::AddOp>(forOp.getLoc(), ivInBody, step);
 
       SmallVector<Value> loopCarried = {newIvInBody};
+      for (auto [iterArg, yieldedIterArgs] :
+           llvm::zip(forOp.getRegionIterArgs(),
+                     forOp.getBody()->getTerminator()->getOperands())) {
+        if (maps[mapping.lookup(iterArg)].getOperands() !=
+            maps[mapping.lookup(yieldedIterArgs)].getOperands()) {
```
ivanradanov, line 1626:
```suggestion
        if (maps[mapping.lookup(iterArg)] !=
            maps[mapping.lookup(yieldedIterArgs)]) {
```

Thank you very much for the fixes.

If the maps themselves must match then this should be better?

PR #484, opened 2025-03-16
Fix tests

giordano:
* Use [`CHECK-DAG`](https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-dag-directive) directive for matching instructions out-of-order
* Use [regex](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-regex-matching-syntax) for matching variables names

CC: @Pangoraw.


giordano:
Tests are finally passing again on all architectures 🙂

test/lit_tests/raising/llvm_to_affine_cst.mlir
```diff
@@ -132,11 +132,11 @@ module {
 // CHECK-NEXT:    scf.if %20 {
 // CHECK-NEXT:      %21 = arith.addi %12, %c-1_i64 : i64
 // CHECK-NEXT:      %22 = arith.muli %21, %c23862_i64 : i64
-// CHECK-NEXT:      %23 = arith.index_cast %22 : i64 to index
+// CHECK-NEXT:      %23 = arith.index_cast %{{[0-9]+}} : i64 to index
```
wsmoses, line 135: The tests passing is obviously good, but I’m still concerned that the behavior is different under the hood. Likely this means that we have something like a(b(), c()) somewhere where on Linux it calls b then c whereas Apple does c then a.

giordano: Minor comment: it's architecture-dependent rather then OS-dependent (aarch64-linux is same as aarch64-darwin)

wsmoses:
superceeded by https://github.com/EnzymeAD/Enzyme-JAX/pull/485

PR #497, opened 2025-03-19
Split parallel iteration space to simplify if statements

ivanradanov:
The question still stands whether we want this enabled by default or only in some cases because it can have implications on code size.


ivanradanov:
@wsmoses @ftynse 

It looks like this results in less gathers and sliiightly less instructions overall???(or roughly the same)
this is after --affine-cfg --raise-affine-to-stablehlo on [this IR
](https://github.com/EnzymeAD/Enzyme-JAX/issues/536#issue-2951927034)```
(ins)$ cat after-raise.split.mlir | grep gather | wc -l
144
(cmd)$ cat after-raise.mlir | grep gather | wc -l
186

(cmd)$ cat after-raise.split.mlir | wc -l
275455
(cmd)$ cat after-raise.mlir | wc -l
276531
```

Issue #499, opened 2025-03-19
Aarch64 issue


wsmoses:
x/ref https://pastebin.com/YGAyHbGH

@giordano

giordano:
Another module: https://gist.github.com/giordano/4f88260f0ac518d5b256b56393d514f0

giordano:
More output using #500: [output.log.gz](https://github.com/user-attachments/files/19339149/output.log.gz)

wsmoses:
Reduced: https://pastebin.com/KCib7891

wsmoses:
run ./bazel-bin/enzymexlamlir-opt -affine-cfg file.mlir on an aarch64 system

giordano:
The pastebin link doesn't work.  https://gist.github.com/ may be more useful 🙂

PR #500, opened 2025-03-19
With affine print

wsmoses:


giordano:
I believe this was only for debugging (which did help for tracking down #499), closing.

PR #502, opened 2025-03-19
Fixaarch64

wsmoses:
@giordano @Pangoraw


giordano:
Ok, I could confirm (but I presume you both did it as well) this is making GB code working on aarch64 now, thanks!

PR #505, opened 2025-03-20
Do not use subscript to assign to the same densemap

### LINKED ISSUE #504: Nondeterministic crash in raise affine to stablehlo
[mwe.mlir.txt](https://github.com/user-attachments/files/19356842/mwe.mlir.txt)

cc @ivanradanov @Pangoraw

ivanradanov:
The [] operator returns a reference and using it to assign to the same densemap can result in invalid memory access if a reallocation of the map happens.

```diff
-      maps[whileRes] = maps[mapping.lookup(forIterArg)];
+      maps[whileRes] = maps.lookup(mapping.lookup(forIterArg));
```

Fix #504, fix https://github.com/PRONTOLab/GB-25/issues/67.


giordano:
Is a test hard to come up with? 🥲

ftynse:
> Is a test hard to come up with? 🥲

it depends on the allocator behavior, so yes

PR #509, opened 2025-03-20
Handle ne neg step with cmp taking the iv

Pangoraw:
Closes #508. cc @arpitj1, I would appreciate your review.

PR #510, opened 2025-03-20
Fix another off by 1

Pangoraw:
the second added test in #509 was wrong since the range is inclusive in this case. The same logic should be applied for other comp predicates.


wsmoses (approved):

wsmoses:
cc @ivanradanov and @arpitj1 for post commit review

PR #511, opened 2025-03-21
Re-land jax update to 5fef4cff7a37d0bb4d7004741189880b357699a2

ftynse:
This reverts commit 0b0e36a3a78a4dcda5f50ca42b4cd4f2f6534fa1.


giordano:
Note that the update also breaks Reactant: https://github.com/openxla/xla/issues/24033

ftynse:
It should be possible to nuke the entire ppc nonsense somewhere here
https://github.com/EnzymeAD/Enzyme-JAX/blob/main/workspace.bzl#L13-L37

ftynse:
Also, our bazel may be very outdated:
https://github.com/bazel-contrib/rules_go/issues/4109

giordano:
Once the revert of the offending commit reaches GitHub I'll test whether the new XLA + this branch works for Reactant, if so we'll be good to go.

giordano:
Just reverted: https://github.com/openxla/xla/commit/821715be3f7ded31723b9b4c0d5de8f0d2a30a6b.  I'm going to test it with this PR.

giordano:
I'm getting
```
API.cpp:867:10: error: cannot initialize return object of type 'xla::PjRtLoadedExecutable *' with an rvalue of type 'pointer' (aka 'xla::PjRtExecutable *')
  867 |   return std::move(exec_err).value().release();
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
in Reactant, but we probably need to adapt the API there as well: https://github.com/EnzymeAD/Reactant.jl/blob/e07bbe109fbf0bb17bf1f91e90f259bd75d58a97/deps/ReactantExtra/API.cpp#L867

giordano:
Now this PR is blocked by #515.

giordano (approved):

PR #518, opened 2025-03-23
Update JAX

giordano:


giordano:
Tested a local build of Reactant with this version of Enzyme-JAX (and without custom XLA in Reactant), all tests pass, so this should be good to go as soon as tests here finish.

PR #522, opened 2025-03-24
Lower cpu into exit

wsmoses:


giordano (approved):
I can't test it because I don't have a reproducer anymore, but looks good to me as far as I understand.

PR #524, opened 2025-03-25
lift memref.load depending on scf.if into that

ftynse:
When a load depends on the conditional value produced by an `scf.if`, lift that load into both conditional branches when possible (no interferring side-effecting operations). This may allow us to turn such loads into affine loads and further raise them.

Additional patterns are needed to fold arithmetic operations into affine forms and further use those to construct and affine load.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -1,3 +1,4 @@
+#include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Affine/Analysis/AffineAnalysis.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/Passes.h"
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -4325,6 +4326,254 @@ struct SinkStoreInIf : public OpRewritePattern<scf::IfOp> {
...
+    rewriter.replaceOp(loadOp,
+                       newConditional->getResults().take_back(numLoadResults));
+    rewriter.replaceOp(conditional,
+                       newConditional->getResults().drop_back(numLoadResults));
+    return success();
+  }
+};
+
+template <typename Derived, typename BinOp>
+struct FoldAffineApplyBase : public OpRewritePattern<BinOp> {
```
wsmoses, line 4417: this should be handled by https://github.com/EnzymeAD/Enzyme-JAX/blob/e4438171e9a5bc5c62e906f2ee0be3b3e035b145/src/enzyme_ad/jax/Passes/AffineCFG.cpp#L1003 no?

ftynse: This goes into the opposite direction: we now fold _users_ of affine apply.

wsmoses: ah got it

ftynse:
It's doing something weird with signs, please hold.

wsmoses (approved):

ftynse:
macOS CI caught a problem with pattern that could have led to non-deterministic failures elsewhere

PR #529, opened 2025-03-26
copy select-of-extract patterns to libdevice-raise

ftynse:
The absence of these patterns prevents raising for certain constants. They also run later as part of scf.for canonicalization.

### DIFF
--- a/src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
+++ b/src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
@@ -12,6 +12,7 @@
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "src/enzyme_ad/jax/Passes/Passes.h"
+#include "src/enzyme_ad/jax/Passes/SelectPatterns.h"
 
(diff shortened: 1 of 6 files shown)


src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
```diff
@@ -500,6 +500,70 @@ struct ReadOnlyAllocaElim : public OpRewritePattern<LLVM::AllocaOp> {
   }
 };
 
+struct SelectExtractElementToExtractElementSelect
```
wsmoses, line 503: can we put this in a common utility function so we don't duplicate it in both libdevice and canonicalize-scf-for (though perhaps it becomes no longer necessary there)

ftynse: Done. Note that the libdevice operates on LLVM selects whereas the canonicalize-for operates on arith selects.

wsmoses: do we need an llvm select one? the libdevice pattern itself should create arith selects so the arith select one should still apply?

wsmoses:
going to merge for now, can address in post commit review

PR #530, opened 2025-03-26
Don't bail merge parallel induction when there are no index cast

Pangoraw:


wsmoses:
@ivanradanov can you review this

Pangoraw:
This breaks raisecmp test so this is not quite right

src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -3833,7 +3833,10 @@ struct MergeParallelInductions
           bool onlyUsedInAdd = addIndices[i] != nullptr;
           if (onlyUsedInAdd)
             for (auto ilOp : illegal) {
-              if (!ilOp || !isa<MulIOp, ShLIOp>(ilOp)) {
+              if (!ilOp)
+                continue;
+
+              if (!isa<MulIOp, ShLIOp>(ilOp)) {
```
wsmoses, line 3839: this is wrong

wsmoses: we need to set onlyUsedInAdd = false if there is an illegal op

Issue #535, opened 2025-03-27
Select iota optimizations


wsmoses:
a couple of relatively critical optimizations of hlo that might reduce compute by a factor of two (cc @hhkit @ivanradanov )

```
select (iota < cst), tensor1, tensor2
->
concat tensor1[0:cst], tensor2[cst:end]
```

```
select (iota == cst), tensor1, tensor2
->
concat tensor2[1:cst-1], tensor1[cst], tensor2[cst+1:end]
```

hhkit:
is this only for iota in 1 dimension, or can we expect higher dimensional tensors?

```
// RUN: enzymexlamlir-opt %s --enzyme-hlo-opt  --split-input-file | FileCheck %s

module {
  func.func @comp_lt(%arg0: tensor<20xi64>, %arg1: tensor<20xi64>) -> tensor<20xi64>{
    %0 = stablehlo.iota dim = 0 : tensor<20xi64>
    %c = stablehlo.constant dense<5> : tensor<20xi64>    
    %1 = stablehlo.compare  LT, %0, %c : (tensor<20xi64>, tensor<20xi64>) -> tensor<20xi1>
    %2 = stablehlo.select %1, %arg0, %arg1: tensor<20xi1>, tensor<20xi64>
    return %2 : tensor<20xi64>
  }
}

//CHECK       module {
//CHECK-NEXT:   func.func @comp_lt(%arg0: tensor<20xi64>, %arg1: tensor<20xi64>) -> tensor<20xi64>{
//CHECK-NEXT:     %0 = stablehlo.slice %arg0 [0:5] : (tensor<20xi64>) -> tensor<5xi64>
//CHECK-NEXT:     %1 = stablehlo.slice %arg1 [5:20]: (tensor<20xi64>) -> tensor<15xi64>
//CHECK-NEXT:     %2 = stablehlo.concatenate %0, %1, dim = 0: (tensor<5xi64>, tensor<15xi64>) -> tensor<20xi64>
//CHECK-NEXT:     return %2 : tensor<20xi64>
//CHECK-NEXT:   }
//CHECK-NEXT: }

// -----

module {
  func.func @comp_eq(%arg0: tensor<20xi64>, %arg1: tensor<20xi64>) -> tensor<20xi64>{
    %0 = stablehlo.iota dim = 0 : tensor<20xi64>
    %c = stablehlo.constant dense<5> : tensor<20xi64>    
[...]

PR #539, opened 2025-03-27
Add Select-Compare-Iota optimization (#535)

hhkit:
Refer to [the unit test i wrote](https://github.com/EnzymeAD/Enzyme-JAX/blob/a6d8cf6c7e107f04f9c98e2e41d8e1929c0c6c5c/test/lit_tests/selectcompiotaconstsimplify.mlir) for everything I tested.

I have tested:
- all six flags
- swapping the const and iota
- higher dimensions (but LT only)
- if the constant is negative or above the iota dimension

I have also changed the `affine_to_stablehlo5.mlir unit` test, because it was further reduced by this optimization.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -6295,7 +6295,13 @@ struct SelectCompIotaConst final : OpRewritePattern<mlir::stablehlo::SelectOp> {
       if (type.getShape().size() != 1)
         return failure();
 
-    auto compareOp = cast<stablehlo::CompareOp>(compare.getDefiningOp());
+    if (!compare)
+      return failure();
+    auto compOp = compare.getDefiningOp();
+    if (!compOp)
+      return failure();
+
+    auto compareOp = dyn_cast<stablehlo::CompareOp>(compOp);
     if (!compareOp)
       return failure();
 
@@ -6335,13 +6341,6 @@ struct SelectCompIotaConst final : OpRewritePattern<mlir::stablehlo::SelectOp> {
 
     auto constValue = inp.getSplatValue<IntegerAttr>().getInt();
     auto endValue = selectOp.getType().getShape().front();
-    if (constValue <= 0) {
-      rewriter.replaceAllUsesWith(selectOp, falseTensor);
-      return success();
-    }
-    if (constValue >= endValue) {
-      rewriter.replaceAllUsesWith(selectOp, trueTensor);
-    }
 
     rewriter.setInsertionPointAfterValue(compare);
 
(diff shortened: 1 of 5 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6279,6 +6279,154 @@ struct CompareExt final : OpRewritePattern<mlir::stablehlo::CompareOp> {
...
+                                    m_Constant(&inp),
+                                    m_Op<mlir::stablehlo::IotaOp>()))) {
+        // incoming: const `op` iota
+        // treat the match as iota `op` const
+        switch (flag) {
+        case stablehlo::ComparisonDirection::LT:
+          flag = stablehlo::ComparisonDirection::GT;
+          break;
+        case stablehlo::ComparisonDirection::LE:
+          flag = stablehlo::ComparisonDirection::GE;
```
wsmoses, line 6325: do you need to store here if its inverted to concatenate things in the proper order?

test/lit_tests/raising/affine_to_stablehlo5.mlir
```diff
@@ -15,9 +15,8 @@ module {
 }
 
 // CHECK:  func.func private @main_raised(%arg0: tensor<180xf32>, %arg1: tensor<180xf32>, %arg2: tensor<180xf32>) -> (tensor<180xf32>, tensor<180xf32>, tensor<180xf32>) {
-// CHECK-NEXT:    %c = stablehlo.constant dense<90> : tensor<180xi64>
-// CHECK-NEXT:    %0 = stablehlo.iota dim = 0 : tensor<180xi64>
-// CHECK-NEXT:    %1 = stablehlo.compare GT, %0, %c, SIGNED : (tensor<180xi64>, tensor<180xi64>) -> tensor<180xi1>
-// CHECK-NEXT:    %2 = stablehlo.select %1, %arg0, %arg1 : tensor<180xi1>, tensor<180xf32>
-// CHECK-NEXT:    return %arg0, %arg1, %2 : tensor<180xf32>, tensor<180xf32>, tensor<180xf32>
+// CHECK-NEXT:    %0 = stablehlo.slice %arg1 [0:91] : (tensor<180xf32>) -> tensor<91xf32>
```
wsmoses, line 18: shouldn't the cutoff here be 90, not 91?

wsmoses:
@Pangoraw mind giving this a review as well?

wsmoses:
and/or @ftynse

test/lit_tests/raising/affine_to_stablehlo5.mlir
hhkit: Let me work this out. My head has been quite toast from staring at the cases for so long.

The `%iota` generates the set [0, 179].
`%iota > 90` is the set [91, 179], which is true for the latter half of the first input.
`%iota <= 90` is the set [0, 90], which is, syntactically, `[0:91]` because stablehlo.slice's end is the exclusive limit.

wsmoses: ah yeah right

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
hhkit: I ran around myself in circles a couple of times because i originally swapped the true and false tensors as well and the test results wouldn't change.

Because I am always "canonicalizing" to the form of "iota op const", I only have to flip the direction of the flag. The i1 tensor stores the same truth value. When we later map it to the true/false tensors, those tensors don't need to be swapped as well.

That is an admittedly terrible explanation and truthfully the only reason I know it works is because it survives the unit test.

wsmoses: oh I understand

wsmoses (approved):

PR #544, opened 2025-03-28
If Index cast optimization 

arpitj1:
affine.if result if yields a constant with one user that does index cast on it. Then the index cast is moved inside the affine.if


src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -5211,6 +5211,156 @@ static bool isLoopParallel(AffineForOp forOp,
...
+    SmallVector<arith::IndexCastOp> indexCastOps(ifOp.getNumResults(), nullptr);
+    bool hasOptimizableResult = false;
+
+    // Check each result to see if it's used in a single index_cast operation
+    for (auto result : llvm::enumerate(ifOp.getResults())) {
+      int resultIdx = result.index();
+      Value value = result.value();
+
+      // Skip if the result is not an integer type that can be index-cast
+      if (!value.getType().isIntOrIndex()) {
```
wsmoses, line 5234: can we add a further restriction here if the yielded results are both constants?

wsmoses: or I suppose if at least one of the yielded results are constants*

wsmoses: [that way it is in some sense work reducing]

arpitj1: Added this check

PR #546, opened 2025-03-28
Lockstep for op raising

ivanradanov:
We don't actually have any loops with this pattern currently

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -5600,7 +5600,7 @@ static bool isLoopMemoryLockStepExecutable(AffineForOp forOp) {
     return false;
 
   // Collect all load and store ops in loop nest rooted at 'forOp'.
-  SmallVector<Operation *, 8> loadAndStoreOps;
+  SmallVector<Operation *> loadAndStoreOps;
   auto walkResult = forOp.walk([&](Operation *op) -> WalkResult {
     if (auto readOp = dyn_cast<AffineReadOpInterface>(op)) {
       // Memrefs that are allocated inside `forOp` need not be considered.
@@ -5611,10 +5611,7 @@ static bool isLoopMemoryLockStepExecutable(AffineForOp forOp) {
       if (!isLocallyDefined(writeOp.getMemRef(), forOp))
         loadAndStoreOps.push_back(op);
     } else if (!isa<AffineForOp, AffineYieldOp, AffineIfOp>(op) &&
-               !hasSingleEffect<MemoryEffects::Allocate>(op) &&
                !isMemoryEffectFree(op)) {
-      // Alloc-like ops inside `forOp` are fine (they don't impact parallelism)
-      // as long as they don't escape the loop (which has been checked above).
       return WalkResult::interrupt();
     }
 
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -5611,7 +5611,7 @@ static bool isLoopMemoryLockStepExecutable(AffineForOp forOp) {
       if (!isLocallyDefined(writeOp.getMemRef(), forOp))
         loadAndStoreOps.push_back(op);
     } else if (!isa<AffineForOp, AffineYieldOp, AffineIfOp>(op) &&
-               !isMemoryEffectFree(op)) {
+               !isReadNone(op)) {
       return WalkResult::interrupt();
     }
 
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -5645,6 +5645,14 @@ static bool isLoopMemoryLockStepExecutable(AffineForOp forOp) {
         return false;
       }
 
+      // I haven't thought through the logic in this case, conservatively fail
+      // for now.
+      bool eitherNestedInNestedFor =
+          dstOp->getParentOfType<affine::AffineForOp>() != forOp ||
+          srcOp->getParentOfType<affine::AffineForOp>() != forOp;
+      if (eitherNestedInNestedFor)
+        return false;
+
       if (srcOp == dstOp) {
         // Since we will be executing different iterations of the same
         // instruction at the same time in lock step fashion, any dependence
@@ -5701,12 +5709,12 @@ struct AffineParallelizePattern : public OpRewritePattern<affine::AffineForOp> {
                                 PatternRewriter &rewriter) const override {
     SmallVector<LoopReduction> reductions;
     if (!::isLoopParallel(forOp, parallelReductions ? &reductions : nullptr))
-      return failure();
+      return rewriter.notifyMatchFailure(forOp, "!isLoopParallel");
 
     // Fail early if there are iter arguments that are not reductions.
     unsigned numReductions = reductions.size();
     if (numReductions != forOp.getNumIterOperands())
-      return failure();
+      return rewriter.notifyMatchFailure(forOp, "reduction num mismatch");
 
     Location loc = forOp.getLoc();
     rewriter.setInsertionPoint(forOp);
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -5414,7 +5414,7 @@ bool isReadNone(Operation *op) {
   return false;
 }
 
-// Vendored
+// ------------------------- Vendored -------------------------
 
 /// Returns true if `v` is allocated locally to `enclosingOp` -- i.e., it is
 /// allocated by an operation nested within `enclosingOp`.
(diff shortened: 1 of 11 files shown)


ivanradanov:
@wsmoses can be used like this
```
./bazel-bin/enzymexlamlir-opt mid3.mlir --raise-affine-to-stablehlo=prefer_while_raising=false -o mid3.raised.while.mlir
./bazel-bin/enzymexlamlir-opt mid3.mlir --raise-affine-to-stablehlo=prefer_while_raising=true -o mid3.raised.unroll.mlir
```

We don't find any lockstep executable loops it looks like

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
   return res;
 }
 
+static LogicalResult tryRaisingForOpToStableHLOUnroll(
+    affine::AffineForOp forOp, IRMapping &mapping, OpBuilder &builder,
+    llvm::DenseMap<Value, affine::AffineValueMap> &maps, ParallelContext pc) {
+  auto tmpBlock = new Block();
+  OpBuilder oldFuncBuilder(builder.getContext());
```
ftynse, line 985: Nit: prefer updating the insertion point of the existing builder, potentially along with `InsertionGuard`. This makes sure the builder/rewriter configuration and listeners is correctly preserved.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+  auto yield = oldFuncBuilder.create<affine::AffineYieldOp>(
+      clonedFor.getLoc(), clonedFor.getResults());
+  if (failed(affine::loopUnrollFull(clonedFor))) {
+    delete tmpBlock;
+    return failure();
+  }
+  for (auto &innerOp : tmpBlock->without_terminator()) {
+
+    if (tryRaisingOpToStableHLO(&innerOp, mapping, builder, maps, pc).failed())
+      return failure();
```
ftynse, line 997: This leaks the block. Just put it in a `unique_ptr` so it's always deleted when exiting the scope and `.release` it when merging below.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+  auto clonedFor = cast<affine::AffineForOp>(oldFuncBuilder.clone(*forOp));
+  auto yield = oldFuncBuilder.create<affine::AffineYieldOp>(
+      clonedFor.getLoc(), clonedFor.getResults());
+  if (failed(affine::loopUnrollFull(clonedFor))) {
+    delete tmpBlock;
+    return failure();
+  }
+  for (auto &innerOp : tmpBlock->without_terminator()) {
+
+    if (tryRaisingOpToStableHLO(&innerOp, mapping, builder, maps, pc).failed())
```
ftynse, line 996: Big warning: this code needs significantly more discipline around mutation. Specifically, the mapping may eventually be mutated before returning failure, which will result in it containing values that were deleted. Such errors are extremely difficult to debug, in particular, when the same address that was used by the deleted value gets reused by a different value (and the allocator tends to do that!). GDB cannot detect this. ASAN cannot detect this. The behavior is non-deterministic and often disappears under instrumentation as that uses different allocation schemes. Without exageration, I have wasted months of my life debugging such bugs. Spend 30 minutes and add `const` whenever the mapping is not mutated and pedantically document expectations. This will save you months at some point.

Note that it doesn't _look_ to me that it is _currently_ being mutated, but it's only a matter of time before somebody adds another failure case when it is.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+      return failure();
+  }
+  for (auto [yielded, res] :
+       llvm::zip_equal(yield.getOperands(), forOp.getResults())) {
+    auto mapped = mapping.lookupOrNull(yielded);
```
ftynse, line 1001: We very likely want to remove the mapping for `yielded` after this. The value will no longer exist after the block is deleted, but the mapping for it will persist, and may be erroneously associated with some other value that just happens to have the same address.

Maybe we can have a local IRMapping that we use for tmpBlock and then update that one with only the values that survive.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+      SplatElementsAttr::get(
+          TT, ArrayRef<Attribute>(IntegerAttr::get(ET, range.ub))));
+  Value step = builder.create<stablehlo::ConstantOp>(
+      forOp.getLoc(), TT,
+      SplatElementsAttr::get(
+          TT, ArrayRef<Attribute>(IntegerAttr::get(ET, range.step))));
+
+  Block *entryBlock = &forOp->getParentOfType<func::FuncOp>().getBody().front();
+
+  Block *cond = new Block(), *body = new Block();
```
ftynse, line 1044: Nit: use OpBuilder/Rewriter to create blocks.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+  return success();
+}
+
+template <class T> static SmallVector<BlockArgument, 6> getIVs(T op);
+template <> SmallVector<BlockArgument, 6> getIVs(affine::AffineParallelOp op) {
+  return {op.getIVs().begin(), op.getIVs().end()};
+}
+template <> SmallVector<BlockArgument, 6> getIVs(affine::AffineForOp op) {
+  return {op.getInductionVar()};
+}
```
ftynse, line 1143: Isn't `LoopLikeInterface::getLoopInductionVars` working  for these?

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1533,150 +1782,32 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
...
-    {
-      OpBuilder::InsertionGuard guard(builder);
-      builder.setInsertionPointToStart(cond);
-      Value cond = builder.create<stablehlo::CompareOp>(
-          forOp.getLoc(), ivInCond, ub, stablehlo::ComparisonDirection::LT);
-      builder.create<stablehlo::ReturnOp>(forOp.getLoc(), cond);
+    if (pc.options.enableLockstepFor &&
+        tryRaisingLockStepForOpToStableHLO(forOp, mapping, builder, maps, pc)
+            .succeeded()) {
+      llvm::errs() << "LOCKSTEP\n";
```
ftynse, line 1794: leftover debug

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,217 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+    assert(mapped);
+    mapping.map(res, mapped);
+
+    auto found = maps.find(yielded);
+    if (found != maps.end()) {
+      affine::AffineValueMap avm = found->second;
+      maps[res] = avm;
+    }
+  }
+  delete tmpBlock;
```
ftynse, line 1011: Don't we want to keep the result on success?

src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -5299,6 +5299,187 @@ static bool isLoopParallel(AffineForOp forOp,
...
+        loadAndStoreOps.push_back(op);
+    } else if (auto writeOp = dyn_cast<AffineWriteOpInterface>(op)) {
+      // Filter out stores the same way as above.
+      if (!isLocallyDefined(writeOp.getMemRef(), forOp))
+        loadAndStoreOps.push_back(op);
+    } else if (!isa<AffineForOp, AffineYieldOp, AffineIfOp>(op) &&
+               !hasSingleEffect<MemoryEffects::Allocate>(op) &&
+               !isMemoryEffectFree(op)) {
+      // Alloc-like ops inside `forOp` are fine (they don't impact parallelism)
+      // as long as they don't escape the loop (which has been checked above).
```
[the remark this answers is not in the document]
ivanradanov: Hmm. I copied this over from the parallelization checks. I removed allocations as allowed for now since we don't need them.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -552,10 +560,21 @@ bool isSafeToSpeculativelyExecuteAtScope(Operation *scope, Operation *op) {
   return inBounds;
 }
 
+// The name is parallel context but a more accurate description would be
+// LockStepContext
 struct ParallelContext {
+  struct Options {
+    bool enableLockstepFor = true;
+    bool preferWhileRaising = true;
+  } & options;
```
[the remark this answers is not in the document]
ivanradanov: Hm. yes certainly a copy is better

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -596,6 +615,17 @@ struct ParallelContext {
...
+    ParallelContext newPc = *this;
+    auto iv = forOp.getInductionVar();
+    auto ivr = getIVRange(iv);
+    if (!ivr)
+      return std::nullopt;
+    newPc.ranges.push_back(*ivr);
+    newPc.ivs.push_back(iv);
+    return newPc;
+  }
+
```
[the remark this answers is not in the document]
ivanradanov: We only call this function when we determine a for loop is "parallelizable" in our slightly loosened condition for legal parallel execution which is not "all iterations can execute in arbitrary order and overlap arbitrarily" but "we can execute all iterations in lockstep at the same time". i.e. we allow write-after-reads to be broken in some cases like suppose we had this
```
      affine.for %arg7 = 0 to 19 {
        %1 = affine.load %arg0[%arg7 + 1] : memref<20xf64, 1>
        affine.store f(%1), %arg0[%arg7 ] : memref<2xf64, 1>
```
Which we currently do not so it's questionable whether we want to merge support for it at all.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
ivanradanov: Oh yes unique_ptr is a great idea. I completely missed this leak....

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
ivanradanov: I wanted to avoid the overhead of copying over the mapping but you are right that this could lead to some obscure bugs so I will do that.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
ivanradanov: I found the std::optional wrapping of the result of that function confusing so I opted to use these.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
ivanradanov: I deliberately created a new builder here because the builder we pass in is for building the raised version of the function and the builder here is only used for building a temporary unrolled version of the loop. None of the operations created by the new builder will make it alive out of this function.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -944,6 +978,223 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+  // Materialize an unrolled version of the loop in a temporary block and
+  // generate the raised version of that. The unrolled version will be deleted
+  // afterwards and the results of the original for loop will be mapped to it.
+
+  // There arises a problem with the affine maps contained in the for loop which
+  // until now correctly identified the loop iv as an affine dim, but will now
+  // take constants as inputs. We need to canonicalize those maps before raising
+  // the operations because we assume dim inputs to be loop ivs. This is the
+  // reason we need to canonicalize all affine maps before we raise them in the
+  // other parts of the code.
```
ivanradanov, line 993: @ftynse I wanted to ask you about this part. We would want to canonicalize the affine exprs in the unrolled version of the loop, but as far as I know we can't run the pattern rewriter in just a block  because it needs an IsolatedFromAbove operation. Is there a nice way to be able to reuse the canonicalizers or would we need to rewrite/vendor them in?

We would like to do the unroll while in the raising sequence because we want the option to first try the for->while raising and only fallback to this if that doesn't succeed.
But perhaps it might be better to attempt the raising and if we fail due to an infeasible while->for unroll that specific loop, try again and repeat until we manage to raise.

Not sure if there is a nicer way to do this and I am overthinking it.

wsmoses:
this case does apply from real code in:
```
  func.func private @"##call__Z39gpu__compute_integrated_ab2_tendencies_16CompilerMetadataI10StaticSizeI8_45__20_E12DynamicCheckvv7NDRangeILi2ES0_I6_3__2_ES0_I8_16__16_EvvEE5FieldI4Face6Centervvvv11OffsetArrayI7Float64Li3E13CuTracedArrayISC_Li3ELi1E11_59__34__1_EESC_vvvES8_ISA_S9_vvvvSB_ISC_Li3ESD_ISC_Li3ELi1E11_59__35__1_EESC_vvvE21LatitudeLongitudeGridISC_8Periodic7BoundedSM_28StaticVerticalDiscretizationISB_ISC_Li1ESD_ISC_Li1ELi1E5_25__EESB_ISC_Li1ESD_ISC_Li1ELi1E5_24__EESP_SR_ESC_SC_SB_ISC_Li1E12StepRangeLenISC_14TwicePrecisionISC_ESV_5Int64EESY_SC_SC_SY_SY_SB_ISC_Li1ESD_ISC_Li1ELi1E5_34__EES10_S10_S10_SC_SC_vSW_EvSB_ISC_Li3ESD_ISC_Li3ELi1E12_59__34__24_EESB_ISC_Li3ESD_ISC_Li3ELi1E12_59__35__24_EES13_S15_SC_#445$par98"(%arg0: memref<1x34x59xf64, 1>, %arg1: memref<1x35x59xf64, 1>, %arg2: memref<24xf64, 1>, %arg3: memref<24x34x59xf64, 1>, %arg4: memref<24x35x59xf64, 1>, %arg5: memref<24x34x59xf64, 1>, %arg6: memref<24x35x59xf64, 1>) {
    %cst = arith.constant 1.600000e+00 : f64
    %cst_0 = arith.constant 6.000000e-01 : f64
    %cst_1 = arith.constant 0.000000e+00 : f64
    affine.parallel (%arg7, %arg8) = (0, 0) to (20, 45) {
      %0 = affine.load %arg2[7] {alignment = 8 : i64, ordering = 0 : i64, tbaa = [#tbaa_tag]} : memref<24xf64, 1>
      %1 = affine.load %arg5[7, %arg7 + 7, %arg8 + 7] : memref<24x34x59xf64, 1>
      %2 = arith.mulf %1, %cst {fastmathFlags = #llvm.fastmath<none>} : f64
      %3 = affine.load %arg3[7, %arg7 + 7, %arg8 + 7] : memref<24x34x59xf64, 1>
      %4 = arith.mulf %3, %cst_0 {fastmathFlags = #llvm.fastmath<none>} : f64
      %5 = arith.subf %2, %4 {fastmathFlags = #llvm.fastmath<none>} : f64
      %6 = arith.mulf %0, %5 {fastmathFlags = #llvm.fastmath<none>} : f64
      affine.store %6, %arg0[0, %arg7 + 7, %arg8 + 7] : memref<1x34x59xf64, 1>
      %7 = affine.load %arg2[7] {alignment = 8 : i64, ordering = 0 : i64, tbaa = [#tbaa_tag]} : memref<24xf64, 1>
      %8 = affine.load %arg6[7, %arg7 + 7, %arg8 + 7] : memref<24x35x59xf64, 1>
      %9 = arith.mulf %8, %cst {fastmathFlags = #llvm.fastmath<none>} : f64
      %10 = affine.load %arg4[7, %arg7 + 7, %arg8 + 7] : memref<24x35x59xf64, 1>
      %11 = arith.mulf %10, %cst_0 {fastmathFlags = #llvm.fastmath<none>} : f64
      %12 = arith.subf %9, %11 {fastmathFlags = #llvm.fastmath<none>} : f64
      %13 = affine.if #set(%arg7) -> f64 {
        affine.yield %12 : f64
      } else {
        affine.yield %cst_1 : f64
      }
      %14 = arith.mulf %7, %13 {fastmathFlags = #llvm.fastmath<none>} : f64
      affine.store %14, %arg1[0, %arg7 + 7, %arg8 + 7] : memref<1x35x59xf64, 1>
      affine.for %arg9 = 0 to 9 {
        %15 = affine.load %arg0[0, %arg7 + 7, %arg8 + 7] : memref<1x34x59xf64, 1>
        %16 = affine.load %arg2[%arg9 + 8] {alignment = 8 : i64, ordering = 0 : i64, tbaa = [#tbaa_tag]} : memref<24xf64, 1>
        %17 = affine.load %arg5[%arg9 + 8, %arg7 + 7, %arg8 + 7] : memref<24x34x59xf64, 1>
        %18 = arith.mulf %17, %cst {fastmathFlags = #llvm.fastmath<none>} : f64
        %19 = affine.load %arg3[%arg9 + 8, %arg7 + 7, %arg8 + 7] : memref<24x34x59xf64, 1>
        %20 = arith.mulf %19, %cst_0 {fastmathFlags = #llvm.fastmath<none>} : f64
        %21 = arith.subf %18, %20 {fastmathFlags = #llvm.fastmath<none>} : f64
        %22 = arith.mulf %16, %21 {fastmathFlags = #llvm.fastmath<none>} : f64
        %23 = arith.addf %15, %22 {fastmathFlags = #llvm.fastmath<none>} : f64
        affine.store %23, %arg0[0, %arg7 + 7, %arg8 + 7] : memref<1x34x59xf64, 1>
        %24 = affine.load %arg1[0, %arg7 + 7, %arg8 + 7] : memref<1x35x59xf64, 1>
        %25 = affine.load %arg2[%arg9 + 8] {alignment = 8 : i64, ordering = 0 : i64, tbaa = [#tbaa_tag]} : memref<24xf64, 1>
        %26 = affine.load %arg6[%arg9 + 8, %arg7 + 7, %arg8 + 7] : memref<24x35x59xf64, 1>
        %27 = arith.mulf %26, %cst {fastmathFlags = #llvm.fastmath<none>} : f64
        %28 = affine.load %arg4[%arg9 + 8, %arg7 + 7, %arg8 + 7] : memref<24x35x59xf64, 1>
        %29 = arith.mulf %28, %cst_0 {fastmathFlags = #llvm.fastmath<none>} : f64
        %30 = arith.subf %27, %29 {fastmathFlags = #llvm.fastmath<none>} : f64
        %31 = affine.if #set1(%arg7, %arg9) -> f64 {
          affine.yield %30 : f64
        } else {
          affine.yield %cst_1 : f64
        }
        %32 = arith.mulf %25, %31 {fastmathFlags = #llvm.fastmath<none>} : f64
        %33 = arith.addf %24, %32 {fastmathFlags = #llvm.fastmath<none>} : f64
        affine.store %33, %arg1[0, %arg7 + 7, %arg8 + 7] : memref<1x35x59xf64, 1>
      }
    }
    return
  }
  ```

PR #549, opened 2025-03-28
fix: build

avik-pal:
cc @giordano. tested locally that build passes on reactant


giordano (approved):

wsmoses (approved):
im surprised we didn't break earlier?

ftynse:
maybe a version bump brought more transitive link dependencies

PR #563, opened 2025-03-30
Moving transpose to before whileOp  + broadcast iota error fix

arpitj1:
Transform this:

out = while(in)
{
y = something(in)
....
x =tr(a)  //inner transpose
return x
}
c = tr(out) //outer transpose

to this:

z = tr(in) //Moved to before while
out = while(z)
{
new = tr(z) //Another moved to beginning of while
y=something(new)
...
return a
}

c = out;


avik-pal:
[test_sharded.mlir.txt](https://github.com/user-attachments/files/19527591/test_sharded.mlir.txt)

we might be hitting an infinite loop of optimizations here

arpitj1:
> [test_sharded.mlir.txt](https://github.com/user-attachments/files/19527591/test_sharded.mlir.txt)
> 
> we might be hitting an infinite loop of optimizations here

I think it's fixed with the latest change. I was able to run this through

avik-pal:
can confirm

wsmoses (approved):

PR #572, opened 2025-03-31
slice(reduce_window) -> reduce

tyb0807:


wsmoses (approved):

wsmoses (commented):
can you also add the td to primitives.py

PR #577, opened 2025-03-31
reshape

chelini:


wsmoses:
@chelini can you also add the transform dialect like in https://github.com/EnzymeAD/Enzyme-JAX/pull/576

avik-pal (approved):

PR #581, opened 2025-03-31
reshape(slice(x)) -> slice(reshape(x)) if reshape adds a unit dimension

jumerckx:
working on tests

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -6257,6 +6257,28 @@ struct ReshapeSlice final : OpRewritePattern<mlir::stablehlo::ReshapeOp> {
     }
     const UnitDimMapping &mapping = *mappingOpt;
 
+    // Check if the reshape removes a unit dimension that was *created* by the
+    // slice.
+    // If so, the transformation is invalid.
+    for (int64_t removedSliceDim : mapping.removedSourceDims) {
+      // removedSliceDim is an index into sliceResultType
+      if (sliceResultType.getDimSize(removedSliceDim) != 1) {
+        // Should not happen if mapping is correct, but double-check
+        return failure();
+      }
+      int64_t originalDim =
+          findOriginalDimForSliceResultDim(sliceOp, removedSliceDim);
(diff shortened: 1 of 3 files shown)


wsmoses:
from @avik-pal 's test I think we need both adding a dimension and also removing a dimension

jumerckx:
I liberally used ai assistance, explaining the verbose comments all over the place.
I haven't yet unified the slicereshapehelper logic with the utility added here but I will.
The tests I added should pass, I'm not sure what the buildkite failures are about?

giordano:
> I'm not sure what the buildkite failures are about?

This is the only command we run in buildkite at the moment: https://github.com/EnzymeAD/Enzyme-JAX/blob/bf4a081615ae4a9d983b3d7387eb17303705b9d7/.buildkite/pipeline.yml#L29 If there's a failure is because the pipeline failed to start at all.

wsmoses:
@Pangoraw can you review this PR

avik-pal:
https://gist.github.com/avik-pal/890a979a20c7614239b4d924f6b31e72

./bazel-bin/enzymexlamlir-opt --enzyme-hlo-generate-td="patterns=transpose_elementwise;slice_reshape_elementwise;reshape_elementwise;reshape_concat;reshape_slice" --transform-interpreter --enzyme-hlo-remove-transform --cse --canonicalize --enzyme-hlo-opt envs/new_sharding_no_dbg_info.mlir > envs/new_sharding2.mlir
envs/new_sharding_no_dbg_info.mlir:153:14: error: number of output elements (556740) doesn't match expected number of elements (79057080)
      %105 = stablehlo.reshape %104 {mhlo.sharding = "{devices=[2,2,2]<=[2,2,2]T(1,0,2) last_tile_dim_replicate}"} : (tensor<1x268x2060xf64>) -> tensor<268x2060xf64>

jumerckx:
oof that's bad, I'll try debug

jumerckx:
The problem was slice making a dimension 1. I now detect this and don't do any transformation in that case.
Though at second thought, it should be possible to be a bit more precise and still handle other size-1 dimensions, not introduced by the slice itself.
If that needs to be supported, I can look into it tomorrow.

wsmoses:
subsumed by other PR

PR #601, opened 2025-04-02
Remove dependent induction variable from stablehlo.while

arpitj1:
Currently supports:
1. Identifies induction variable
2. Addition of constant step value
3. Less than comparision


wsmoses (approved):

wsmoses:
awesome!

very minor, can you add this to transform dialect tablegen?

PR #636, opened 2025-04-05
feat: pad(concat, const) to concat(..., const)

### LINKED ISSUE #626: `concatenate` + `pad` (specific cases) -> `broadcast_in_dim(val) followed by a concatenate`
%256 = stablehlo.concatenate %224, %255, dim = 1 {mhlo.sharding = "{devices=[1,4,8]<=[8,4]T(1,0)}"} : (tensor<1x2031x4080xf64>, tensor<1x1x4080xf64>) -> tensor<1x2032x4080xf64> loc(#loc1194)

      %257 = stablehlo.pad %256, %cst_73, low = [0, 1, 0], high = [0, 0, 0], interior = [0, 0, 0] {mhlo.sharding = "{devices=[1,4,8]<=[8,4]T(1,0)}"} : (tensor<1x2032x4080xf64>, tensor<f64>) -> tensor<1x2033x4080xf64> loc(#loc1194)

avik-pal:
fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/626

note conflicting optimization, cant be in the default options

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -11806,6 +11806,87 @@ struct BroadcastInDimIsReshape final
   }
 };
 
+struct ConstPadConcatToConcat : public OpRewritePattern<stablehlo::PadOp> {
+  using OpRewritePattern<stablehlo::PadOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(stablehlo::PadOp padOp,
+                                PatternRewriter &rewriter) const override {
+    auto concatOp =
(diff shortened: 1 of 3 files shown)


wsmoses:
This is somewhat dangerous on its own, would it be reasonably to make the body a separate method and also thus have the double optimization from the issue (we can still leave this as its own pattern for testing ofc)

wsmoses:
Generally making things pad is preferred to being a constant concat

avik-pal:
I mostly want to run this on the IR (not actually enable it in reactant), it makes it a bit easier to parse

avik-pal:
> also thus have the double optimization from the issue

what do you mean?

wsmoses:
Ah ignore my double optimization comment

wsmoses (approved):

PR #637, opened 2025-04-05
WIP

chelini:


wsmoses (commented):
I think this is too specify in its current form =/

wsmoses:
converted into the more general https://github.com/EnzymeAD/Enzyme-JAX/pull/646

though it also still needs to be further generalized

Issue #632, opened 2025-04-05
while induction replacement of of pad


wsmoses:
```
result = while arg = input  : bigsize {

....

   yield pad(x, ...)
}
```

->

```
result2 = while arg2 = input : smallsize {
    arg = select if first iteration {
        input
    } else {
        pad(arg2, ...)
    }
....

   yield x
}
result = pad(result2, ...)
```

wsmoses:
fairly similar to / shared code with https://github.com/EnzymeAD/Enzyme-JAX/pull/641

wsmoses:
completed via https://github.com/EnzymeAD/Enzyme-JAX/pull/647

PR #647, opened 2025-04-05
while with pad in yield 

arpitj1:
simplified the yield to be without padded values
https://github.com/EnzymeAD/Enzyme-JAX/issues/632

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -10654,6 +10654,9 @@ struct WhilePadInductionReduction
       if (!pad)
         continue;
 
+      if (!definedOutside(pad.getPaddingValue(), whileOp))
+        continue;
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -10543,6 +10543,307 @@ struct WhileRepeatedInductionReduction
...
+    };
+
+    llvm::SmallVector<Candidate> candidates;
+    for (unsigned idx = 0; idx < yieldOp.getNumOperands(); ++idx) {
+
+      // check if operand is a pad operation
+      auto pad = yieldOp.getOperand(idx).getDefiningOp<stablehlo::PadOp>();
+      if (!pad)
+        continue;
+
```
wsmoses, line 10656: can you make sure that definedOutside(pad.getPaddingValue(), whileOp) for it to be a valid candidate

PR #658, opened 2025-04-06
Check for parallel IV in affineMapToSlice

Pangoraw:


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -337,6 +290,54 @@ struct ParallelContext {
...
+  for (unsigned i = 0; i < rank; i++) {
+    auto expr = accessValueMap.getResult(i);
+
+    if (auto constExpr = dyn_cast<AffineConstantExpr>(expr)) {
+      strides.push_back(1);
+      continue;
+    }
+
+    Value iv = getIVForExpr(accessValueMap, expr);
+    if (affine::isAffineForInductionVar(iv) && !pc.isParallelIV(iv)) {
```
Pangoraw, line 319: This is the core of the change (adding `!pc.isParallelIV(iv)`)

ivanradanov (approved):
lgtm

I wonder if we should make all IV queries go through the ParallelContext

wsmoses (approved):

PR #660, opened 2025-04-06
Don't raise top level ops differently & raise arith bitcast

Pangoraw:


wsmoses (commented):
can you add that test?

Pangoraw:
There is a crash in whileDeadResults opt when using prefer_while_raising=true and optimizing, investigating. The test and the reactant raise pipeline is written with prefer_while_raising=false so it should not crash there

PR #672, opened 2025-04-07
Xor simplify

wsmoses:


wsmoses:
cc @ftynse for postcommit review.

We hit the limit of the number of transform op patterns that can be included within a template (default max template args in clang is 256). Trying to work around, but on monday if you have better tools for doing

wsmoses:
```
bazel-out/darwin_arm64-dbg/bin/external/llvm-project/mlir/include/mlir/Dialect/Transform/IR/TransformDialect.h.inc:102:46: fatal error: instantiating fold expression with 257 arguments exceeded expression nesting limit of 256
      (addOperationIfNotRegistered<OpTys>(), ...);
```

PR #674, opened 2025-04-07
feat: reshape select

avik-pal:


avik-pal:
@wsmoses do you know what is the mac failure?

wsmoses:
we have defined more transform passes than the macos limit for template parameters

wsmoses:
I also just hit here: https://github.com/EnzymeAD/Enzyme-JAX/pulls
the answer is to increase a config, but it doesnt exist for gcc, naturally

wsmoses:
@avik-pal can you rebase?

wsmoses (approved):

PR #679, opened 2025-04-07
Fix reverse mode derivative for `stablehlo.broadcast_in_dim`

vimarsh6739:
Should ideally fix the bug in EnzymeAD/Reactant.jl#1117

### DIFF
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -682,7 +682,6 @@ public:
     llvm::SmallVector<int64_t> sorted_bcastDims(bcastDims.begin(),
                                                 bcastDims.end());
     llvm::sort(sorted_bcastDims);
-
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -676,8 +678,20 @@ class AutoDiffBroadcastInDimRev
 
     Value res = red->getResult(0);
     Type resTy = gutils->getShadowType(op.getOperand().getType());
-    if (res.getType() != resTy)
-      res = builder.create<ReshapeOp>(op.getLoc(), resTy, res);
+
+    llvm::SmallVector<int64_t> sorted_bcastDims(bcastDims.begin(),
+                                                bcastDims.end());
+    llvm::sort(sorted_bcastDims);
```
wsmoses, line 684: add a test?

wsmoses (approved):

vimarsh6739:
Need to figure out why this is crashing tests again

giordano:
Would it be useful to revert this PR, since it broke CI in two different repos?

PR #683, opened 2025-04-07
Don't reverse unless necessary

wsmoses:


Pangoraw:
we can do the same with broadcast in dims

PR #718, opened 2025-04-09
feat: rotate op communication pattern

avik-pal:


wsmoses (approved):

wsmoses (approved):
Don’t have cycles for deep review but cursory scan looks good

avik-pal:
I have a feeling the partition id math isn't correct for both the passes (essentially not generally to handle any arbitrary concat dim)

Issue #700, opened 2025-04-09
Rotate of Slice / Wrap of Slice


wsmoses:
If you have an op
  with multiple slices which have rotate users

we should make one rotate on a larger slice thereof, and slice into the rotate (essentially we want to perform the rotate on as large a tensor as possible, thereby reducing comms).

same thing for a wrap

PR #736, opened 2025-04-10
SliceWrap

jumerckx:
Other part of #700

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -15336,6 +15336,146 @@ struct SliceExtend final : OpRewritePattern<enzymexla::ExtendOp> {
   }
 };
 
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -15336,6 +15336,143 @@ struct SliceExtend final : OpRewritePattern<enzymexla::ExtendOp> {
...
+        }
+      }
+    }
+
+    if (candidates.size() <= 1) {
+      return rewriter.notifyMatchFailure(
+          triggerWrapOp,
+          "Rewrite condition not met (only found the trigger candidate)");
+    }
+
```
wsmoses, line 15424: @jumerckx can you apply the similar fix I made in https://github.com/EnzymeAD/Enzyme-JAX/commit/53652ee51583d80d5e016836e86e1ccc649a34ba

PR #738, opened 2025-04-11
Test fixes

ivanradanov:
@wsmoses Couldn't immediately figure out `test/lit_tests/while_induction_reduction.mlir` so I XFAIL'd it for now. Should we merge and get back to it later?


wsmoses (commented):
yeah i can fix while_induction_reduction

PR #739, opened 2025-04-11
Support right to left rotate

ivanradanov:
WIP, looking at whether any changes are needed to the padding part of the code


test/lit_tests/communication/rotate-direction.mlir
```diff
@@ -0,0 +1,39 @@
...
+// CHECK:             %[[VAL_4:.*]] = "stablehlo.collective_permute"(%[[VAL_3]]) <{channel_handle = #stablehlo.channel_handle<handle = 1, type = 0>, source_target_pairs = dense<{{\[\[}}3, 0], [0, 1], [1, 2], [2, 3]]> : tensor<4x2xi64>}> : (tensor<4x8x1xf64>) -> tensor<4x8x1xf64>
+// CHECK:             %[[VAL_5:.*]] = stablehlo.slice %[[VAL_2]] [0:4, 0:8, 0:19] : (tensor<4x8x20xf64>) -> tensor<4x8x19xf64>
+// CHECK:             %[[VAL_6:.*]] = stablehlo.concatenate %[[VAL_4]], %[[VAL_5]], dim = 2 : (tensor<4x8x1xf64>, tensor<4x8x19xf64>) -> tensor<4x8x20xf64>
+// CHECK:             sdy.return %[[VAL_6]] : tensor<4x8x20xf64>
+// CHECK:           } : (tensor<4x8x80xf64>) -> tensor<4x8x80xf64>
+// CHECK:           return %[[VAL_1]] : tensor<4x8x80xf64>
+// CHECK:         }
+
+  func.func @right_to_left(%arg: tensor<4x8x80xf64>) -> (tensor<4x8x80xf64>) {
+    %res = "enzymexla.rotate"(%arg) <{amount = 79 : si32, dimension = 2 : si32}> {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{"z", ?}, {"y", ?}, {"x", ?}]>]>} : (tensor<4x8x80xf64>) -> tensor<4x8x80xf64>
```
wsmoses, line 36: can you also add a case where the rotate isn't entirely within the leftmost or rightmost region. I.e.

|.       |.   *   |.       |.      |

              ^ we rotate from here in the second region. As a result we need to send not left or right by one, but by two

wsmoses: alternatively simply making sure we don't accidentally do the wrong thing in this case and returning failure() is fine for now too

ivanradanov:
@avik-pal Actually I don't understand the padding, doesn't it have to be enzymexla.extend instead of stablehlo.pad so that we get the actual values for the exchange on the edge, and not the padded zeroes? I may be missin
g something.

test/lit_tests/communication/rotate-direction.mlir
ivanradanov: I don't understand, how would the rotate op look like? or do you mean a slice->rotate?

wsmoses: mesh is 4 nodes

rotate(100xf32, amount=30)

wsmoses:
moved into https://github.com/EnzymeAD/Enzyme-JAX/pull/740

avik-pal:
> @avik-pal Actually I don't understand the padding, doesn't it have to be enzymexla.extend instead of stablehlo.pad so that we get the actual values for the exchange on the edge, and not the padded zeroes? I may be missing something.


the padding is mostly because I don't yet understand how shardy deals with non-divisible axes, so I was making those divisible with a padding before. Though you might be right here and we need to handle it with an extended rotate followed by a slice

PR #743, opened 2025-04-11
Extend of Elementwise and Extend of splat and Wrap of elementwise

### LINKED ISSUE #722: Elementwise of extends and/or splats -> extend of elementwise
also applies to wraps

Pangoraw:
Closes #722


wsmoses (approved):

wsmoses (approved):
@Pangoraw as a follow up can you also handle unary elementwise ops generally (look for the Elementwise pattern but add a check it is a unary operator (aka one operand)

PR #746, opened 2025-04-11
Fix rotate test

Pangoraw:
The opt pattern does not match because of:

https://github.com/EnzymeAD/Enzyme-JAX/blob/21e0b857aa26219b0c439e39c80b4b4df59b4706/src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp#L1113-L1115


wsmoses (approved):
we should fix, but this is fine for now

PR #760, opened 2025-04-12
feat: use padding to optimize concat comm

avik-pal:


wsmoses (commented):
can you add the "Transform dialect lite' pass options to enable/disable the current passes.

Just to give us more control without rebuilding the jll

wsmoses (approved):

PR #761, opened 2025-04-12
Split opt comm into pass options

wsmoses:


src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
```diff
@@ -2451,16 +2451,24 @@ struct OptimizeCommunicationPass
...
+      patterns.add<WrapCommOptimize>(context, PatternBenefit(wrap_comm));
+
+    if (extend_comm > 0)
+      patterns.add<ExtendCommOptimize>(context, PatternBenefit(extend_comm));
+
+    if (dus_to_pad_comm > 0)
+      patterns.add<DUSToPadComm>(context, PatternBenefit(dus_to_pad_comm));
+
+    if (concat_to_pad_comm > 0)
+      patterns.add<ConcatToPadCommOptimize>(context,
```
avik-pal, line 2470: add ConcatTwoOperandsCommOptimize as well with a default of 0?

wsmoses: done

PR #810, opened 2025-04-14
Broadcast iota

wsmoses:


wsmoses:
@giordano @Pangoraw since you're both presently up. This breaks the affine raising tests, but causing them to be more optimized. I'm about to fall over, and this is required for large grids, so I'm going to merge and we can resolve tests later.

PR #814, opened 2025-04-14
optimize and(compare, compare)

ftynse:


Pangoraw (approved):
The test in #815 which passes with the same opt looks sensible.

wsmoses (approved):

PR #853, opened 2025-04-21
feat: optimize concat(reshape(slice)...) to transpose

### LINKED ISSUE #852: Slice -> Reshape -> Concat can be optimized to Transpose
https://gist.github.com/avik-pal/dd2b626b06a5027bbf70392e5131aea8

avik-pal:
fixes #852

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -9702,8 +9702,7 @@ struct DUSSliceSimplify final
         });
 
     LLVM_DEBUG(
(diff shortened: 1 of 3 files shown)


wsmoses (commented):
Didn’t yet look at the patch I feel like this might be more easily done by looking at ConcatSlice and making a ConcatReshapeSlice that makes it reshape at the end if it can fuse the slices

avik-pal:
> Didn’t yet look at the patch I feel like this might be more easily done by looking at ConcatSlice and making a ConcatReshapeSlice that makes it reshape at the end if it can fuse the slices

updated to reuse the merging logic. the remaining checks are mostly checking that the reshape is doing an axis swap.

wsmoses (approved):

PR #857, opened 2025-04-23
WhileOp Min cut

Pangoraw:


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -2159,6 +2160,501 @@ class AutoDiffBatchNormTrainingRev
...
+          parent.emplace(v, u);
+        }
+      }
+    }
+  }
+
+  // Whether or not an operation can be moved from the forward region to the
+  // reverse region or vice-versa.
+  static inline bool isMovable(Operation *op) {
+    return mlir::isPure(op) && op->getNumRegions() == 0;
```
wsmoses, line 2252: longer term we may want to turn this into an interface (e.g. because things like reduceop have regions, but should be fine)

wsmoses: or more generally the issue is whether any op within a region uses a value defined outside the region

wsmoses (approved):
we should generalize the ismovable (e.g. so stablehlo.reduce doesn't inadvertantly force a cut), otherwise lgtm.

longer term we should consider fancy cost models

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
Pangoraw: Yes, the code only looks at operands to figure out if an operation can be moved: 

https://github.com/EnzymeAD/Enzyme-JAX/blob/ba778b6dfc540de2c638e0caeff222621727c2d0/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp#L2352-L2357

PR #861, opened 2025-04-25
Err if we can't remove while op gradient push/pop

wsmoses:


wsmoses:
Long story, this is required to resolve a huge compile time issue.

Specifically, if we fail to do all removals on the first pass, we rely on enzyme-hlo-opt to do the removal after raising to stablehlo, combining that with cse causes some catastrophic issues, which just having it emit the better code directly resolves

avik-pal:
> Specifically, if we fail to do all removals on the first pass, we rely on enzyme-hlo-opt to do the removal after raising to stablehlo, combining that with cse causes some catastrophic issues, which just having it emit the better code directly resolves

Can we just run the passes without CSE included? One of the cases where this is needed is while6.mlir

wsmoses:
So that fix is actually just making the while op actually know that it has the constant bounds by the time that remove enzyme ops runs (aka cache optimizations). I’m honestly really surprised that things worked in the current version regardless to be totally honest

PR #864, opened 2025-04-25
do not cache known iter count for whileop

Pangoraw:
Checking for ConstOp would not work in while6 because it was a pop.


Pangoraw:
Ok something is wrong because it is crashing in CI, trying to reproduce locally.

Issue #870, opened 2025-04-27
`'stablehlo.concatenate' op failed to infer returned types` in GB-25 code


giordano:
https://github.com/PRONTOLab/GB-25/actions/runs/14689629084/job/41222987061#step:10:602
```
[...]

Issue #840, opened 2025-04-16
`error: 'stablehlo.concatenate' op failed to infer returned types`


giordano:
```
[...]

PR #875, opened 2025-04-28
Fix one dimensional type check

wsmoses:


wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/870 and perhaps https://github.com/EnzymeAD/Enzyme-JAX/issues/840

PR #876, opened 2025-04-28
feat: memory effect annotations on functions

avik-pal:


src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    module.walk([&](FunctionOpInterface funcOp) {
+      bool hasRead = false;
+      bool hasWrite = false;
+      bool hasAllocate = false;
+      bool hasFree = false;
+
+      funcOp.walk([&](Operation *op) {
+        if (auto memOp = dyn_cast<MemoryEffectOpInterface>(op)) {
+          SmallVector<MemoryEffects::EffectInstance, 4> effects;
```
ftynse, line 36:
```suggestion
          SmallVector<MemoryEffects::EffectInstance> effects;
```

Nit

src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp
```diff
@@ -0,0 +1,70 @@
...
+    OpBuilder builder(ctx);
+
+    module.walk([&](FunctionOpInterface funcOp) {
+      bool hasRead = false;
+      bool hasWrite = false;
+      bool hasAllocate = false;
+      bool hasFree = false;
+
+      funcOp.walk([&](Operation *op) {
+        if (auto memOp = dyn_cast<MemoryEffectOpInterface>(op)) {
```
ftynse, line 35: We should be conservative and assume all possible effects if an operation doesn't implement this interface. We can't reason about it.

ftynse: Also consider the `HasRecursiveSideEffects` trait.

src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+          for (auto &effect : effects) {
+            if (effect.getEffect() == MemoryEffects::Read::get())
+              hasRead = true;
+            else if (effect.getEffect() == MemoryEffects::Write::get())
+              hasWrite = true;
+            else if (effect.getEffect() == MemoryEffects::Allocate::get())
+              hasAllocate = true;
+            else if (effect.getEffect() == MemoryEffects::Free::get())
+              hasFree = true;
```
ftynse, line 47: We should be conservative or assert on other kinds (there are currently none).

src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp
```diff
@@ -0,0 +1,70 @@
...
+          }
+        }
+      });
```
ftynse, line 50: I suppose we want to recurse into function calls / enzyme function calls somehow. For this, we need to build a call graph (there may be helper for that), do a topological sort, reverse and proceed in that order. If there are cycles in the call graph, we may need to iterate until convergence or assume worst case.

avik-pal: added this

src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp
avik-pal: > We should be conservative and assume all possible effects if an operation doesn't implement this interface. We can't reason about it.

Add a flag for this (mostly for testing that we can get the minimal set correctly) with the default being we assume all possible effects.

> Also consider the HasRecursiveSideEffects trait.

I will check it out

avik-pal:
@wsmoses the build seems busted

wsmoses:
ignore that for now, finishing up hopefully a "fix the CI" spree

wsmoses:
test fails:
```
//test/lit_tests:memoryeffects/call.mlir.test                            FAILED in 0.4s
  /home/runner/.bazel/execroot/__main__/bazel-out/k8-opt/testlogs/test/lit_tests/memoryeffects/call.mlir.test/test.log
  ```

**Comment by avik-pal** [comment]
tests should be passing now

File: src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp [hunk 5]
```diff
@@ -0,0 +1,216 @@
...
+      if (dfs(target))
+        return true;
+    }
+    visitState[node] = VisitState::Visited;
+    topoOrder.push_back(node);
+    return false;
+  }
+
+  void insertMemoryEffectsAsStringRefs(
+      llvm::SmallDenseSet<StringRef> &effects,
```
ftynse, line 50: We can use `MemoryEffects::Effect` instead of StringRef here and only convert when creating the attribute. This avoids strings flying around in the logic. A micro-optimization on top: use `uint8_t` with bit flags for each effect, but that may be an overkill.

ftynse (commented):
Looks good to me modulo two places that should be more conservative by default (I've debugged way too many of these, it's worth it).

ftynse (approved):

src/enzyme_ad/jax/Passes/LowerJIT.cpp
```diff
@@ -989,12 +989,15 @@ struct LowerJITPass
           NamedAttribute(rewriter.getStringAttr("attr"), backendstr));
       auto dattr = DictionaryAttr::get(op.getContext(), names);
 
+      BoolAttr hasSideEffectAttr =
+          rewriter.getBoolAttr(!isMemoryEffectFree(op));
```
wsmoses, line 993: so this is too conservative here. In particular as it currently is nearly everything will be marked as side-effecting (even if the code doesn't need to). Specifically here we only need to mark things as side-effective if it could write to data which isn't a function argument (e.g. if we were to write to a function argument, that would be represented as an updated result perfectly fine).

test/lit_tests/lowering/gpujit.mlir
```diff
@@ -37,7 +37,7 @@ module attributes {gpu.container_module} {
 }
 
 // CHECK: func.func @main(%arg0: tensor<64xi64>) -> tensor<64xi64> {
-// CHECK-NEXT:    %0 = stablehlo.custom_call @enzymexla_compile_gpu(%arg0) {api_version = 4 : i32, backend_config = {attr = "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00"}, output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [], operand_index = 0, operand_tuple_indices = []>]} : (tensor<64xi64>) -> tensor<64xi64>
+// CHECK-NEXT:    %0 = stablehlo.custom_call @enzymexla_compile_gpu(%arg0) {api_version = 4 : i32, backend_config = {attr = "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00"}, has_side_effect = true, output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [], operand_index = 0, operand_tuple_indices = []>]} : (tensor<64xi64>) -> tensor<64xi64>
```
wsmoses, line 40: per above, this (and all other existing cases, for example) are indeed side effect free from xla semantics

src/enzyme_ad/jax/Passes/LowerJIT.cpp
ftynse: I'd say let's start conservative and then relax progressively. If we go into the per-value effect territory (the representation supports both and arguably there are "global" effects without value that we may not want to ignore, e.g. synchronizations), we may need to account for aliasing, at which point we should just borrow the enzyme-mlir analyses.

wsmoses: yeah that's fine, but it does mean we need an intermediate attribute to force it to be side effect free (current lowerings ought actually be side effect free but are not deduced as such).

wsmoses (approved):
lgtm

PR #885, opened 2025-04-30
feat: register sparse tensor dialect

avik-pal:


wsmoses (approved):

wsmoses:
You might also need to add dependencies in BUILD files

PR #887, opened 2025-05-01
feat: factorization ops

avik-pal:


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -189,4 +189,17 @@ def CommRegionOp : EnzymeXLA_Op<"comm_region", [
   );
 }
 
+def LUFactorizationOp: EnzymeXLA_Op<"lu_factorization", [Pure]> {
```
avik-pal, line 192: `Pure` is probably wrong here. LU routines always alias the input and output matrix

wsmoses: no that's still right here. This can appear as pure (and is, returning a new tensor and taking in a tensor), even if the underlying implementation does things to generate it via a custom call

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -189,4 +189,22 @@ def CommRegionOp : EnzymeXLA_Op<"comm_region", [
...
```
mofeing, line 192: small proposal: this feels like a small subdialect inside `enxymexla`, so how about calling  the op `enzymexla.linalg.lu`? i know they do this in the IREE dialects

avik-pal:
now it is mostly working. one final change I want to make is to return the permutation as well (as part of the op spec)

avik-pal:
This is good to go on my end, I have tested the cpu integration on julia end for correctness

avik-pal:
I want to eventually get rid of the looped batching stuff for cpu, but that needs linking to MKL and I can do it in a followup

wsmoses (approved):
LGTM, note that in a follow up pr we probably need to define derivative and batching rules

avik-pal:
default batching rule of pre-pending dims should work here.

For derivatives, we should define https://github.com/JuliaDiff/ChainRules.jl/blob/150c7a1d70cd40d1dc73d463ee641e199732e968/src/rulesets/LinearAlgebra/factorization.jl#L23.

But the most common usecase would be for linear solving, so we should be raising `LU + triangular_solve -> linsolve` and differentiate that (the rule for that is also much simpler)

PR #900, opened 2025-05-04
feat: some more passes for batching

avik-pal:


avik-pal:
I will split this one up once its ready

avik-pal:
moved all the passes to respective PRs

PR #903, opened 2025-05-04
feat: insertdim + dropdim

avik-pal:
makes life a little bit easier when dealing with reshape opts.

splitting up of #900. adding tests in bg


wsmoses (commented):
this is fine, but also means all the other interactions have to understand these too

avik-pal:
yes. progressively adding those. most of them are in the other PR just making them self contained with tests

wsmoses:
design question, do we want to generalize this into some sort of insertdrop op where there's a list of inserted indices and list of dropped indices

avik-pal:
I was thinking we should eventually do that. But it shouldn't be too hard to do rn, let me do that in this pr itself

PR #904, opened 2025-05-04
feat: elementwise reshape like

avik-pal:


avik-pal:
cc @wsmoses anything else needed here

wsmoses (approved):
I'm a bit confused by where we want to use this, since we have already push reshapes up, but sure

PR #906, opened 2025-05-04
feat: aggressively push slice(transpose) down if it can be eliminated for sure

avik-pal:


avik-pal:
This enables

image

to go to

image

wsmoses (approved):

wsmoses:
also note that if we don't have already we probably want an elementwise(broadcast) -> broadcast(elementwise)

avik-pal:
> 
also note that if we don't have already we probably want an elementwise(broadcast) -> broadcast(elementwise)

that was in the last pr. these broadcasts are not identical

wsmoses:
ah fair nuff

PR #910, opened 2025-05-05
fix: element type change

avik-pal:
fixing the failures from https://github.com/EnzymeAD/Reactant.jl/pull/1250


wsmoses (commented):
ah yeah classic convert always causing this problem

wsmoses (approved):

PR #921, opened 2025-05-08
Allow for over-approximated domains in affine expr simplification

ivanradanov:


ivanradanov:
Will add test in a second. And still need to rethink whether the logic is correct

wsmoses (commented):
can you add that test?

PR #923, opened 2025-05-08
Fix sort to be a valid partial order

wsmoses:
@ftynse in post commit review, know of a better place for this comparator (or is there one in the wild)


ftynse:
Perhaps as a separate file under `lib/IR`. We never really needed it, `SetVector` _is_ deterministic unless the order of insertion somehow isn't... If you instead care about dominance, you need something like topological sort on blocks, we have that somewhere I believe.

PR #958, opened 2025-05-17
feat: conj(real number) -> real number

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -18951,6 +18951,20 @@ struct ReduceReduce final : OpRewritePattern<stablehlo::ReduceOp> {
   }
 };
 
+struct ConjReal final : public OpRewritePattern<chlo::ConjOp> {
```
wsmoses, line 18954: did we not have something for this already?

wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: the only other conjop pass is ConjComplexNegate

PR #960, opened 2025-05-17
feat: add `CheckedOpRewritePattern` & `CheckedOpTraitRewritePattern`

avik-pal:
- `supportsDynamicShapes()` -- opt in for patterns that support dynamic shapes. Once all patterns do we can switch the default
- `disablePatternAttrName()` -- allows providing a UnitAttr to the enclosing funcOpInterface to disable patterns. defaults to `"enzymexla.disable_hlo_opts"`

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -340,8 +340,8 @@ struct CheckedOpRewritePattern : public OpRewritePattern<OpTy> {
   using Base = OpRewritePattern<OpTy>;
   using Base::Base;
 
-  LogicalResult matchAndRewrite(OpTy op,
-                                PatternRewriter &rewriter) const override final {
+  LogicalResult
+  matchAndRewrite(OpTy op, PatternRewriter &rewriter) const override final {
     LogicalResult res =
         failIfFuncOpInterfaceHasAttr(op, kDisablePatternAttrName, rewriter);
     if (res.failed())
@@ -364,8 +364,9 @@ struct CheckedOpTraitRewritePattern : public OpTraitRewritePattern<TraitType> {
   using Base = OpTraitRewritePattern<TraitType>;
   using Base::Base;
 
-  LogicalResult matchAndRewrite(Operation *op,
-                                PatternRewriter &rewriter) const override final {
+  LogicalResult
+  matchAndRewrite(Operation *op,
+                  PatternRewriter &rewriter) const override final {
     LogicalResult res =
         failIfFuncOpInterfaceHasAttr(op, kDisablePatternAttrName, rewriter);
     if (res.failed())
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -15221,6 +15221,11 @@ struct ZeroExtentTensorCanon final : RewritePattern {
 
   LogicalResult matchAndRewrite(Operation *op,
                                 PatternRewriter &rewriter) const override {
+    auto disabledByAttr = failIfFuncOpInterfaceHasAttr(
+        op, StringRef("enzymexla.disable_hlo_opts"), rewriter);
+    if (disabledByAttr.failed())
+      return disabledByAttr;
+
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -304,11 +304,109 @@ class StaticSlice {
...
+      LogicalResult res = failIfDynamicShape(op, rewriter);
+      if (res.failed())
+        return res;
+    }
+
+    return this->matchAndRewriteImpl(op, rewriter);
+  }
+
+  // funcopinterface level attribute that can be used to disable the pattern
+  virtual std::optional<StringRef> disablePatternAttrName() const {
```
wsmoses, line 361: this can just be a static const char*, not a virtual fn

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -304,11 +304,109 @@ class StaticSlice {
...
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(OpTy op,
                                 PatternRewriter &rewriter) const override {
+    LogicalResult res =
+        failIfFuncOpInterfaceHasAttr(op, disablePatternAttrName(), rewriter);
+    if (res.failed())
+      return res;
+
+    if (!supportsDynamicShapes()) {
```
wsmoses, line 351: change this to OpTy::supportsDynamicShapes, then change `supportsDynamicShapes` to be a static member function [not a virtual one], and overrides also to be new static member function. This means that we don't require virtual function indirection [and this becomes a constant check]

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -304,11 +304,109 @@ class StaticSlice {
...
+      LogicalResult res = failIfDynamicShape(op, rewriter);
+      if (res.failed())
+        return res;
+    }
+
+    return this->matchAndRewriteImpl(op, rewriter);
```
wsmoses, line 357: same comment here with matchAndRewriteImpl, see SumToReductionBase makeReduction for an example


also cc @ftynse overall for a review/thoughts on

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: can you check if the latest changes are what you wanted?

wsmoses (commented):
didn't check all the patterns, but seems fine otherwise.

I'll note that it seems like a lot of the patterns already ought support dynamic, but we can follow up on that in a subsequent PR

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: yeah indeed!

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -304,11 +304,91 @@ class StaticSlice {
...
+  using Base::Base;
+
+  LogicalResult matchAndRewrite(OpTy op,
+                                PatternRewriter &rewriter) const override {
+    LogicalResult res =
+        failIfFuncOpInterfaceHasAttr(op, kDisablePatternAttrName, rewriter);
+    if (res.failed())
+      return res;
+
+    if (!((Child *)this)->supportsDynamicShapes()) {
```
ftynse, line 350: I don't see much value in CRTP here. The classes already have a vtable and we get to patterns through a vitrual call. A couple more won't change much, and I don't think the pattern infra cost is dominated by virtual calls. Maintainability-wise, it will be easier without.

Also, it may be worth it to declare `matchAndRewrite` final.

avik-pal: @wsmoses what do you think?

wsmoses: I'm okay either way, slight pref to leave as CRTP for ease/perf [esp since we already have]

wsmoses (approved):

PR #970, opened 2025-05-19
feat: slice op batch interface

avik-pal:


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -53,16 +53,25 @@ static llvm::ArrayRef<bool> getBoolIter(llvm::ArrayRef<bool> vals) {
   return vals;
 }
 
-static Value makeI64Constant(Location loc, OpBuilder &builder, int64_t val) {
-  auto Ty = builder.getI64Type();
-  auto unrankedTensorType = RankedTensorType::get({}, Ty);
+Value makeIntegerConstant(Location loc, OpBuilder &builder, Type type,
```
wsmoses, line 56: this may crash the auto generated tablegen code that assumes a function of this name/spec, fyi

wsmoses: ah jk you have the thing below

wsmoses (approved):

PR #983, opened 2025-05-22
feat: general concat reshape to batching pass

avik-pal:


wsmoses:
e.g. we can make https://github.com/EnzymeAD/Enzyme/blob/db0181320d6e425ee963bd496ed0d8dbb615be18/enzyme/Enzyme/MLIR/Passes/EnzymeBatchPass.cpp#L130 into a batchOperation utility or something, which we call directly here instead of outlining into a new function

avik-pal:
yeah that sounds like a better thing to do

avik-pal:
somehow I made it segfault only on mac...

wsmoses:
That's probably an issue of f(a(), b()) running a before b on Linux and opposite on Mac

avik-pal:
@wsmoses any idea what https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/17003076741/job/48208359246?pr=983#step:19:133 is?

wsmoses:
er.....wat re the linked failure

wsmoses:
that seems like some sort of use after free undefined memory error

wsmoses:
we can always just change that to
std::String new_name = ("batched_" + fn.getName()).str();

.github/workflows/test-gb-25.yml
```diff
@@ -53,7 +53,8 @@ jobs:
           - 'main'
           # - '0123456789abcdef0123456789abcdef01234567'
         reactant_commit:
-          - 'main'
+          # - 'main'
+          - 'ap/new_batch_passes'
```
avik-pal:
```suggestion
          - 'main'
```

PR #985, opened 2025-05-25
Add `linalg.qr` op

mofeing:
@avik-pal the StableHLO code after the `jit_call` in the `enzymexla.linalg.lu` code I guess is for reconstructing the L and U matrices from the output of the LAPACK routine right?

Also, in the case of the QR, it needs some workspace memory for storing intermediate computations. @wsmoses should we leave the allocation to XLA / EnzymeXLA or should we allocate it ourselves in the wrapper? I feel like reusing the workspace buffer across multiple LAPACK calls would be something that could be optimized in the future, but don't know how to properly express it.

I might also need to call another LAPACK function, `*ORGQR`/`*UNGQR`, to retrieve the Q matrix.

### DIFF
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
@@ -737,9 +737,9 @@ struct QRFactorizationOpLowering
     LLVMTypeConverter typeConverter(ctx);
 
     auto input = op.getOperand();
-    auto inputShape = cast<RankedTensorType>(input.getType()).getShape();
-    auto inputRank = static_cast<int64_t>(inputShape.size());
     auto inputType = cast<RankedTensorType>(input.getType());
+    auto inputShape = inputType.getShape();
+    auto inputRank = static_cast<int64_t>(inputShape.size());
     auto inputElementType = inputType.getElementType();
 
     const int64_t m = inputShape[inputRank - 2];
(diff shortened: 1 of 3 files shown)


mofeing:
> The git history is very messed up

i fucked it up with the rebase. my bad. im fixing it now.

src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
```diff
@@ -721,6 +750,299 @@ struct LUFactorizationOpLowering
...
+
+    rewriter.replaceAllUsesWith(op.getResult(0), cusolver_call_op.getResult(0));
+    rewriter.replaceAllUsesWith(op.getResult(1), cusolver_call_op.getResult(1));
+
+    // Netlib's LAPACK returns `info`, but cuSOLVER doesn't
+    auto type_info =
+        RankedTensorType::get({}, rewriter.getIntegerType(blasIntWidth));
+    auto info_op = rewriter.create<stablehlo::ConstantOp>(
+        op.getLoc(), type_info, cast<ElementsAttr>(makeAttr(type_info, 0)));
+    rewriter.replaceAllUsesWith(op.getResult(2), info_op.getResult());
```
mofeing: @avik-pal couldn't find what's the behavior of JAX for this cases as you did for the LU

src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
```diff
@@ -721,6 +750,299 @@ struct LUFactorizationOpLowering
...
+    const int64_t m = shape_input[rank_input - 2];
+    const int64_t n = shape_input[rank_input - 1];
+    const int64_t numBatchDims = rank_input - 2;
+
+    if (numBatchDims > 0) {
+      return rewriter.notifyMatchFailure(
+          op, "QR factorization with batch dimensions is not yet supported");
+    }
+
+    // emit `stablehlo.custom_call` to `@cusolver_geqrf_ffi` kernel from jaxlib
```
avik-pal: misplaced comment

src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
```diff
@@ -721,6 +750,299 @@ struct LUFactorizationOpLowering
...
+    auto rank_input = static_cast<int64_t>(shape_input.size());
+
+    const int64_t m = shape_input[rank_input - 2];
+    const int64_t n = shape_input[rank_input - 1];
+    const int64_t numBatchDims = rank_input - 2;
+
+    if (numBatchDims > 0) {
+      return rewriter.notifyMatchFailure(
+          op, "QR factorization with batch dimensions is not yet supported");
+    }
```
avik-pal: this check can be removed, for TPU it should be automatically supported

avik-pal:
> @avik-pal @wsmoses right now i'm using the geqrf routine, but there are others like geqrt (used by LinearAlgebra.jl), geqr3, geqrp, ... Should these methods be available as other EnzymeXLA ops or as attribute of the enzymexla.linalg.qr op?

These should be an attribute imo. By default, any lowering pass is free to choose an algorithm (dependent on backend/sharding). If specified we lower to the specific one. This makes it much easier to define batching rules, sharding rules, etc.

src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
```diff
@@ -721,6 +750,299 @@ struct LUFactorizationOpLowering
...
+      fn = *prefix + fn;
+    } else {
+      op->emitOpError() << "Unsupported complex element type: "
+                        << inputElementType;
+      return rewriter.notifyMatchFailure(op,
+                                         "unsupported complex element type");
+    }
+
+    std::string bind_fn = "enzymexla_lapacke_" + fn;
+    std::string wrapper_fn = "enzymexla_wrapper_lapacke_" + fn;
```
avik-pal, line 818: If we are using lapacke, can we still use https://github.com/JuliaLinearAlgebra/libblastrampoline/ to switch implementations on the julia end? that was one of the original motivations of not directly linking a blas/lapack in enzyme-jax and let downstream users register the functions.

If allocating workspace is an issue, you can create a constant and forward the ptr to that.

src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
```diff
@@ -721,6 +750,299 @@ struct LUFactorizationOpLowering
...
+    auto rank_input = static_cast<int64_t>(shape_input.size());
+
+    const int64_t m = shape_input[rank_input - 2];
+    const int64_t n = shape_input[rank_input - 1];
+    const int64_t numBatchDims = rank_input - 2;
+
+    if (numBatchDims > 0) {
+      return rewriter.notifyMatchFailure(
+          op, "QR factorization with batch dimensions is not yet supported");
+    }
```
avik-pal: can be skipped, the jaxlib functions can handle batching

src/enzyme_ad/jax/Passes/LowerEnzymeXLALinalg.cpp
mofeing: > If we are using lapacke, can we still use https://github.com/JuliaLinearAlgebra/libblastrampoline/ to switch implementations on the julia end?

yes, libblastrampoline still generates the LAPACKE symbols https://github.com/JuliaLinearAlgebra/libblastrampoline/blob/b127bc8dd4758ffc064340fff2aef4ead552f386/ext/gensymbol/gensymbol#L1004

I've tested with a small Julia script using the LAPACKE symbols from LinearAlgebra and `Ops.hlo_call` and it works.

mofeing:
Can we get this merged?

PR #987, opened 2025-05-27
fix: batchnorm gradient

### LINKED ISSUE #337: batchnorm gradient is always zero
https://buildkite.com/julialang/lux-dot-jl/builds/5545#0194f760-d7e2-43f7-af3c-30d3b921a8f1/158-1060

avik-pal:
fixes #337


wsmoses (approved):
yeah that would do it

PR #999, opened 2025-05-30
feat: lift common ops from if

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -19672,6 +19672,54 @@ struct TransposeIf final
...
+          break;
+        }
+      }
+    }
+
+    if (opsToLift.empty())
+      return rewriter.notifyMatchFailure(op, "no common ops found");
+
+    for (auto [trueOp, falseOp] : opsToLift) {
+      Operation *liftedOp = rewriter.clone(*trueOp);
```
wsmoses, line 19709: instead could we perhaps do (assuming the relevant ordering)
```
rewriter.modifyOpInPlace(trueOp, [&](){
    trueOp->moveBefore(ifOp);
})
```

avik-pal: done

wsmoses (approved):
longer term we should think about making equality consider the fact that the false operand is replaced with the true one (since if there is a common expression that depends on a previous one getting hoisted, we'll have to run this pattern a second time)

PR #1012, opened 2025-06-02
feat: implement matchers for complex numbers + simplification patterns

avik-pal:
I can upstream these to mlir proper in a followup pr if needed

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -5757,13 +5757,15 @@ struct SubSimplify
                                     PatternRewriter &rewriter) const {
 
     if (matchPattern(op.getRhs(), m_AnyZeroFloat()) ||
-        matchPattern(op.getRhs(), m_Zero())) {
+        matchPattern(op.getRhs(), m_Zero()) ||
+        matchPattern(op.getRhs(), m_AnyZeroComplex())) {
       rewriter.replaceOp(op, op.getLhs());
       return success();
     }
 
     if (matchPattern(op.getLhs(), m_AnyZeroFloat()) ||
-        matchPattern(op.getLhs(), m_Zero())) {
+        matchPattern(op.getLhs(), m_Zero()) ||
+        matchPattern(op.getLhs(), m_AnyZeroComplex())) {
       rewriter.replaceOpWithNewOp<stablehlo::NegOp>(op, op.getRhs());
       return success();
     }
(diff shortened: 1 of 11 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6018,16 +6020,20 @@ struct MulSimplify
   LogicalResult matchAndRewriteImpl(stablehlo::MulOp op,
                                     PatternRewriter &rewriter) const {
 
-    // 0 * x -> x
+    // 0 * x -> 0
```
wsmoses: incidentally do we have this inside of nonan right now

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6018,16 +6020,20 @@ struct MulSimplify
...
+      rewriter.replaceOpWithNewOp<stablehlo::ConstantOp>(
+          op, cast<ElementsAttr>(makeAttr(op.getType(), 0)));
       return success();
     }
-    // x * 0 -> x
+    // x * 0 -> 0
     if (matchPattern(op.getRhs(), m_AnyZeroFloat()) ||
-        matchPattern(op.getRhs(), m_Zero())) {
-      rewriter.replaceOp(op, op.getRhs());
+            matchPattern(op.getRhs(), m_Zero()),
```
wsmoses: we should also add a similar anyzerocomplex to the enzyme-simplify-math infra as well

avik-pal: that's in enzyme proper right? Should I try to get this upstreamed instead of copying the code around

wsmoses: yes, but also implemented as an interface here too

wsmoses: see https://github.com/EnzymeAD/Enzyme/blob/0828c872bd7ceb5cd3ae0a7e8bbea28f26d8ecad/enzyme/Enzyme/MLIR/Implementations/ArithAutoDiffOpInterfaceImpl.cpp#L64 for example (and relevant here)

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: yeah let me update this

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: https://github.com/EnzymeAD/Enzyme/blob/0828c872bd7ceb5cd3ae0a7e8bbea28f26d8ecad/enzyme/Enzyme/MLIR/Implementations/ComplexAutoDiffOpInterfaceImpl.cpp#L50 seems to implement these

wsmoses (approved):

src/enzyme_ad/jax/Utils.h
```diff
@@ -40,5 +40,84 @@ template <> inline Attribute makeAttr(mlir::Type elemType, llvm::APFloat val) {
...
+  bool match(Operation *op) const {
+    APFloat realValue(APFloat::Bogus());
+    APFloat imagValue(APFloat::Bogus());
+    return constant_complex_value_binder(&realValue, &imagValue).match(op) &&
+           predicate(realValue, imagValue);
+  }
+};
+
+inline constant_complex_predicate_matcher m_AnyZeroComplex() {
+  return {[](const APFloat &realValue, const APFloat &imagValue) {
```
wsmoses, line 105: cc @ftynse

PR #1014, opened 2025-06-03
Checkpointing take 2

Pangoraw:
Replaces #914

TODO: 
 - [ ] Support number of iterations not equal to `numIters = nInner * nOuter`

### DIFF
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -611,9 +611,8 @@ class AutoDiffWhileRev
     auto revInner = makeForLoop(builder, orig.getLoc(), 0, nInner, 1, operands);
     Block *revInnerBody = &revInner.getBody().front();
 
-    if (orig->hasAttr("enzymexla.disable_min_cut"))
-      revInner->setAttr("enzymexla.disable_min_cut",
-                       orig->getAttr("enzymexla.disable_min_cut"));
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -458,6 +458,278 @@ class AutoDiffWhileFwd
...
+
+    if (lastCache)
+      builder.setInsertionPointAfterValue(lastCache);
+    else
+      builder.setInsertionPointToStart(revOuterBody);
+
+    operands.assign(revOuterBody->getArguments().slice(1).begin(),
+                    revOuterBody->getArguments().slice(1).end());
+
+    auto revInner = makeForLoop(builder, orig.getLoc(), 0, nInner, 1, operands);
```
wsmoses, line 611: we should just clone all the attributes of the original loop -- excludingf the checkpointing attr

src/enzyme_ad/jax/Implementations/WhileLoopInfo.cpp
```diff
@@ -82,8 +82,16 @@ std::optional<int64_t> WhileLoopInfo::getConstantLimit() {
...
+  int64_t start = getConstantStart().value(),
+          limit = getConstantLimit().value(), step = getConstantStep().value();
+
+  if (limit < start) // cmp is lt
+    return 0;
+
+  if (step < 0 && limit > start)
+    return 0;
+
+  return (limit - start) / getConstantStep().value();
```
wsmoses: use step here, instead of re-running getConstantStep?

PR #1018, opened 2025-06-03
hasSideEffectAttr fix

sbrantq:
Previously an `enzymexla.jit_call` returning nothing is not effective


wsmoses (commented):
Cc @avik-pal 

and can you add a test

wsmoses:
@sbrantq retroactively can you add a test here

PR #1021, opened 2025-06-09
fix: don't inline kernel call / jit call

avik-pal:


src/enzyme_ad/jax/gpu.cc
```diff
@@ -106,10 +111,65 @@ XLA_FFI_Error *execute(XLA_FFI_CallFrame *call_frame) {
...
+
+  size_t numargs = call_frame->args.size;
+  void *ptrs[numargs];
+  for (size_t i = 0; i < numargs; i++) {
+    ptrs[i] =
+        &reinterpret_cast<XLA_FFI_Buffer *>(call_frame->args.args[i])->data;
+  }
+
+  auto *execution_state = reinterpret_cast<xla::ffi::ExecutionState *>(
+      internal_api->XLA_FFI_INTERNAL_ExecutionState_Get(ctx));
```
wsmoses, line 118: maybe we could push all the common code into a common static function (maybe with a template so we could do 
```
if constexpr (with_errors) {
   error code
} else {
   original code
}
```

avik-pal:
I will test it on julia end and then add a test here. dont merge for now

wsmoses (approved):

PR #1026, opened 2025-06-11
build: centralize dialect/pass registration in EnzymeJAX

avik-pal:


src/enzyme_ad/jax/RegistryUtils.cpp
```diff
@@ -195,3 +270,66 @@ void prepareRegistry(mlir::DialectRegistry &registry) {
   mlir::registerLLVMDialectImport(registry);
   mlir::registerNVVMDialectImport(registry);
 }
+
+void initializePasses(mlir::DialectRegistry &registry) {
```
wsmoses: this shouldn't take a registry here (and also be called once during initialization)

avik-pal: during initialization of what?

wsmoses: Of the binary itself. Like enzymexlamliropt ran it during init, same with reactant’s init. These need to be run once per program, not context

avik-pal:
tests seem to be passing

wsmoses (approved):

PR #1036, opened 2025-06-13
feat: refactor pass pipeline options in python

avik-pal:
updated to match how Reactant is setup

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -167,7 +167,7 @@ py_wheel(
     python_requires = "=="+HERMETIC_PYTHON_VERSION+".*",
     strip_path_prefixes = ["src/"],
     summary = "Enzyme automatic differentiation tool.",
-    version = "0.0.9",
+    version = "0.0.10",
     deps = [
         ":enzyme_jax_data",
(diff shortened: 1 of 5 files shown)


test/test_utils.py
```diff
@@ -412,7 +414,7 @@ def pipelines():
             CurBackends,
         ),
         ("PartOpt", JaXPipeline(partialopt), CurBackends),
-        ("DefOpt", JaXPipeline(hlo_opts()), CurBackends),
+        ("DefOpt", JaXPipeline(full_optimization_pass_pipeline()), CurBackends),
```
wsmoses: minor but this defopt doesn't do inlining, whereas IDefopt does do inlining

BUILD
```diff
@@ -167,7 +167,7 @@ py_wheel(
     python_requires = "=="+HERMETIC_PYTHON_VERSION+".*",
     strip_path_prefixes = ["src/"],
     summary = "Enzyme automatic differentiation tool.",
-    version = "0.0.8",
+    version = "0.0.9",
```
wsmoses: technically this should be 0.0.10

wsmoses (approved):

PR #1049, opened 2025-06-16
Add `linalg.svd` op

mofeing:
# To do

- [x] fix S `complex` element types (S is always real)
- [x] ~find TPU kernel name~
- [x] propagate `full` attribute of `enzymexla.linalg.svd` to `full_matrices` attribute of GPU kernel
- [x] test on rectangular matrices


wsmoses:
I'd double check tensirflow and xla proper, but like your other issue it would be wise to add a stablehlo native fallback for all the ops, and we can optimize when something is available

wsmoses (commented):
@avik-pal I'll let you review this one

avik-pal (approved):

mofeing:
@wsmoses failing tests are unrelated to this PR and although @avik-pal has approved it, I can't merge. mind merging it?

PR #1057, opened 2025-06-17
Explicit check for gpu.alloc async token

pengmai:
This fixes a crash where `gpu.alloc` does not return an async token


wsmoses (approved):
Can you add a quick test so we don’t have this recur

PR #1073, opened 2025-06-20
fix: scatter derivative for all constant inputs

avik-pal:
will test this with the Lux case once before merging


avik-pal:
This fixes the lux issue https://github.com/EnzymeAD/Reactant.jl/pull/1411#issue-3161657417

wsmoses (approved):
Can you add the test?

PR #1079, opened 2025-06-21
feat: various passes to fuse mul chains

### LINKED ISSUE #1076: log simplifications
- log(exp)

```
module @reactant_fn attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<f64>) -> tensor<f64> {
    %0 = stablehlo.exponential %arg0 : tensor<f64>
    %1 = stablehlo.log %0 : tensor<f64>
    return %1 : tensor<f64>
  }
}
```

- log(power)

- log(ka) -> log(k) + log(a) // k is a constant

- log(a * a) -> 2 * log(a)

- log(a / k)

- log(k / a)

- log(sqrt)

- log(cbrt)

- log(rsqrt)

### LINKED ISSUE #1077: chained multiply (>= 3) to power
```
module @reactant_fn2 attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<f64>) -> tensor<f64> {
    %0 = stablehlo.multiply %arg0, %arg0 : tensor<f64>
    %1 = stablehlo.multiply %0, %arg0 : tensor<f64>
    %2 = stablehlo.multiply %1, %arg0 : tensor<f64>
    %3 = stablehlo.multiply %2, %arg0 : tensor<f64>
    %4 = stablehlo.log %3 : tensor<f64>
    return %4 : tensor<f64>
  }
}
```

### LINKED ISSUE #1078: div with floating point constant
XLA does it, so we might as well do it to potentially enable other opts

avik-pal:
fixes #1077 

## TODOs

- [x] don't break const prop
- [x] more pow simplify
- [x] div const simplify fixes #1078 
- [x] disable to pow in the default case
- [x] reassoc is not completely working

```
func.func @main2(%arg0: tensor<f64>) -> tensor<f64> {
    %0 = stablehlo.multiply %arg0, %arg0 : tensor<f64>
    %1 = stablehlo.multiply %0, %arg0 : tensor<f64>
    %2 = stablehlo.multiply %1, %arg0 : tensor<f64>
    %3 = stablehlo.multiply %2, %arg0 : tensor<f64>
    %4 = stablehlo.multiply %arg0, %0 : tensor<f64>
    %5 = stablehlo.multiply %4, %3 : tensor<f64>
    return %5 : tensor<f64>
}
```

```
func.func @main2(%arg0: tensor<f64>) -> tensor<f64> {
    %0 = stablehlo.multiply %arg0, %arg0 : tensor<f64>
    %1 = stablehlo.multiply %0, %0 : tensor<f64>
    %2 = stablehlo.multiply %1, %0 : tensor<f64>
    %3 = stablehlo.multiply %0, %2 : tensor<f64>
    return %3 : tensor<f64>
  }
```

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -5963,7 +5963,22 @@ struct DivSimplify
       return success();
     }
 
-    // TODO: div by constant to mul
+    // x / const -> x * (1 / const)
+    if (isa<FloatType>(op.getType().getElementType())) {
+      DenseElementsAttr rhsAttr;
+      if (matchPattern(op.getRhs(), m_Constant(&rhsAttr))) {
+        auto rhsTen = stablehlo::constantOp(rhsAttr);
+        auto oneTen = stablehlo::constantOp(
+            cast<ElementsAttr>(makeAttr(op.getType(), 1)));
+        auto out = fromTensor(stablehlo::divideOp(oneTen, rhsTen, op.getType()));
+
+        rewriter.replaceOpWithNewOp<stablehlo::MulOp>(
+            op, op.getLhs(),
+            rewriter.create<stablehlo::ConstantOp>(op.getLoc(), op.getType(),
+                                                   out));
+        return success();
+      }
+    }
 
     return failure();
   }
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -19879,6 +19879,11 @@ struct CommonAssociativeCommutativeOpReorder final
     if (lhs == rhs)
       return failure(); // already in a good form
 
+    DenseFPElementsAttr constAttr;
+    if (matchPattern(lhs, m_Constant(&constAttr)) &&
+        matchPattern(rhs, m_Constant(&constAttr)))
+      return rewriter.notifyMatchFailure(op, "const prop has higher priority");
+
     auto lhsDefOp = lhs.template getDefiningOp<Op>();
     auto rhsDefOp = rhs.template getDefiningOp<Op>();
 
@@ -19889,12 +19894,22 @@ struct CommonAssociativeCommutativeOpReorder final
       if (lhsLhs == lhsRhs)
         return failure(); // already in a good form
 
+      if (matchPattern(lhsRhs, m_Constant(&constAttr)) &&
+          matchPattern(lhsLhs, m_Constant(&constAttr)))
+        return rewriter.notifyMatchFailure(op,
+                                           "const prop has higher priority");
+
       if (rhsDefOp && isOnlyUsedInOperation(rhsDefOp, op)) {
         auto rhsLhs = rhsDefOp.getLhs();
         auto rhsRhs = rhsDefOp.getRhs();
 
         if (rhsLhs == rhsRhs)
           return failure(); // already in a good form
+
+        if (matchPattern(rhsRhs, m_Constant(&constAttr)) &&
+            matchPattern(rhsLhs, m_Constant(&constAttr)))
+          return rewriter.notifyMatchFailure(op,
+                                             "const prop has higher priority");
 
         Value commonOperand, otherOperand1, otherOperand2;
         bool foundCommonOperand = false;
@@ -19970,6 +19985,11 @@ struct CommonAssociativeCommutativeOpReorder final
 
       if (rhsLhs == rhsRhs)
         return failure(); // already in a good form
+
+      if (matchPattern(rhsRhs, m_Constant(&constAttr)) &&
+          matchPattern(rhsLhs, m_Constant(&constAttr)))
+        return rewriter.notifyMatchFailure(op,
+                                           "const prop has higher priority");
 
       if (rhsLhs == lhs) { // (op x (op x y)) => (op (op x x) y)
         rewriter.replaceOpWithNewOp<Op>(
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -5956,13 +5956,6 @@ struct DivSimplify
   LogicalResult matchAndRewriteImpl(stablehlo::DivOp op,
                                     PatternRewriter &rewriter) const {
 
-    // 0 / x -> 0 [assume non nan here]
-    if (matchPattern(op.getLhs(), m_AnyZeroFloat()) ||
-        matchPattern(op.getLhs(), m_Zero())) {
-      rewriter.replaceOp(op, op.getLhs());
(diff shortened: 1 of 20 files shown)


wsmoses (approved):
My guess is by default well want to disable the tower power one by default since a pow inst is more expensive than N individual ops (for some N that we'll have to figure out).

Later on we should also do the reassociate for dot general thogh obviously that will be harder to write

wsmoses:
Also we should check first that we aren't breaking an op that contains all constant operands

wsmoses:
Which should ideally be first pref => const prop, second => enable cse, third pref => enable "symmetry" (eg prefer squaring like here)

avik-pal:
> My guess is by default well want to disable the tower power one by default since a pow inst is more expensive than N individual ops (for some N that we'll have to figure out).

Shouldn't this be handled by one of the very low level passes (ideally in XLA) depending on the HW?

If we combine to a pow (makes it nicer with logOps and such) and have a pass that is run at the very end to "un-fuse" ops that will solve this issue

wsmoses:
We'd need to check, I'm not confident so 

Regardless we should add a pow const prop for integer -1,0,1,2 for sure (which shouldnt hinder log optimizations since we want to handle the resulting simplification regardless)

wsmoses:
Also +/- 0.5 to (r)sqrt

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -5999,28 +5999,61 @@ struct PowSimplify
...
-          }
+        // pow(X, -0.5) -> rsqrt(X)
+        if (allNegHalf) {
+          rewriter.replaceOpWithNewOp<stablehlo::RsqrtOp>(op, op.getLhs());
+          return success();
+        }
+
+        // pow(X, 0) -> 1
+        if (allZero) {
+          rewriter.replaceOpWithNewOp<stablehlo::ConstantOp>(
```
wsmoses: This might need a no nan flag if the base is zero

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -5999,28 +5999,61 @@ struct PowSimplify
...
+              op, op.getType(), cast<ElementsAttr>(makeAttr(op.getType(), 1)));
```
wsmoses: We also may want an opt for a base of zero or one

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: Actually rather than a flag we could just add a select if base is zero or nan, nan,  one

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: wdym? both cases will still yield 1 no

wsmoses: I need to check, I would’ve expected 0^0 to be nan, and nan^0 to be nan

wsmoses: https://en.m.wikipedia.org/wiki/Zero_to_the_power_of_zero Yeah okay leave it as one then, we should do a check on the zero base then

wsmoses: Ah I thought this was in response to my Nan comment, I meant in addition to the exponent special cases we should also add base special cases too

avik-pal: separated some div patterns into no-nan as well

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -5956,22 +5956,46 @@ struct DivSimplify
...
+      : CheckedOpRewritePattern(context, benefit),
+        allowOnFloatingPointMath(allowOnFloatingPointMath) {}
+
+  LogicalResult matchAndRewriteImpl(stablehlo::DivOp op,
+                                    PatternRewriter &rewriter) const {
+    if (!canApplyNoNanPattern(allowOnFloatingPointMath, op.getType()))
+      return failure();
+
+    // 0 / x -> 0
+    if (matchPattern(op.getLhs(), m_AnyZeroFloat()) ||
```
wsmoses, line 6008: Honestly even with the nan check off we can still do this, but add a select if x is zero, nan, zero. And the no nan essentially lets us skip the check

avik-pal: if x is nan, it should still be a Nan.

If chlo/shlo had a isnan check, we could potentially replace a bunch of the no-nan patterns with a select.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: It might be worth making an explicit isnanop which we lower but you can have one: https://github.com/EnzymeAD/Enzyme-JAX/blob/ba42c5b203ccefd9b1dfe82ffc276d135d5b59c4/src/enzyme_ad/jax/Passes/ArithRaising.cpp#L263

That said I don't think we'd want to remove all the nan check stuff (but add a branch if not enabled)

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6029,6 +6097,41 @@ struct PowSimplify
...
+        allowOnFloatingPointMath(allowOnFloatingPointMath) {}
+
+  LogicalResult matchAndRewriteImpl(stablehlo::PowOp op,
+                                    PatternRewriter &rewriter) const {
+    if (!canApplyNoNanPattern(allowOnFloatingPointMath, op.getType())) {
+      return failure();
+    }
+
+    // pow(x, 0) -> 1 || pow(1, x) -> 1
+    if ((matchPattern(op.getRhs(), m_Zero()) ||
```
wsmoses: I think my reading of ieee spec is that this does not require no nan actually

wsmoses: Pow(0, x) however does require the nan check

wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: the pow(0, x) simplification leads to a ~1.5x slowdown atleast on GPUs.

avik-pal: I will keep the pattern but remove it from defaults

PR #1089, opened 2025-06-22
feat: reshape(bcast) simplify

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -20296,6 +20296,56 @@ struct NegDivConstSimplify final
...
+  using CheckedOpRewritePattern::CheckedOpRewritePattern;
+
+  LogicalResult matchAndRewriteImpl(stablehlo::ReshapeOp op,
+                                    PatternRewriter &rewriter) const {
+    auto bcastInDimOp =
+        op.getOperand().getDefiningOp<stablehlo::BroadcastInDimOp>();
+    if (!bcastInDimOp)
+      return failure();
+
+    if (!isOnlyUsedInOperation(bcastInDimOp, op))
```
wsmoses, line 20311: You should look at the reshape helper functions I made for all the reshapeX ops, which also is worth checking if subsumes this

wsmoses: https://github.com/EnzymeAD/Enzyme-JAX/blob/7277ac177c3547892a91d2397f8f21ddfbb3d1a2/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L614 and friends in particular

avik-pal: will take a look. I am getting these working first to see the effect on the FNO model. Just this set enables a perf gain of 3% over XLA for the primal

wsmoses: Yeah fair nuff, in which case feel free to merge and maybe open an issue. The helper functions should handle both axes added and removed

avik-pal: https://github.com/EnzymeAD/Enzyme-JAX/issues/1090 I will cleanup post deadline

PR #1112, opened 2025-06-27
ci: format bazel build files + format checker

avik-pal:


wsmoses (approved):
sure, but we might want to get a workflow to automatically reformat for us

avik-pal:
that should be easy

avik-pal:
done

PR #1119, opened 2025-06-28
Canonicalize memref.load(pointer2memref(gep))

tyb0807:
Turning `memref.load(pointer2memref(gep(gep(...(x)...))))` into `memref.load(pointer2memref(x))`

### DIFF
--- a/src/enzyme_ad/jax/Dialect/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Ops.cpp
@@ -1,5 +1,4 @@
-//===- EnzymeXLAOps.cpp - EnzymeXLA dialect ops -----------------------*- C++
-//-*-===//
+//===- EnzymeXLAOps.cpp - EnzymeXLA dialect ops -----------------*- C++ -*-===//
 //
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -7,10 +6,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "Ops.h"
 #include "../Utils.h"
 #include "Dialect.h"
 #include "Interfaces/AutoDiffTypeInterface.h"
+#include "Ops.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/LLVMIR/LLVMTypes.h"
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -850,6 +851,107 @@ class Pointer2Memref2PointerCast final
...
+
+      gepOps.emplace_back(gep, gepElemSize);
+      ptr = gep.getBase();
+    }
+
+    if (gepOps.empty())
+      return failure();
+
+    // FIXME: The base should be a function pointer
+    auto base = ptr.getDefiningOp<LLVM::AddressOfOp>();
```
wsmoses: we shouldn't need this? this is always a valid transformation

src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -850,6 +851,107 @@ class Pointer2Memref2PointerCast final
...
+      Value scaledIdx =
+          (scaledGep != 1)
+              ? rewriter.create<arith::MulIOp>(
+                    loc, idx,
+                    rewriter.create<arith::ConstantIndexOp>(loc, scaledGep))
+              : idx;
+
+      // Then divide if needed
+      Value elemOffset =
+          (scaledElement != 1)
```
wsmoses, line 942: we need to check / early fail if this is not an exact division (e.g. offset by one byte in gep)

tyb0807: I think for dynamic indices we need to insert cf.assert to check this at runtime. Do you have any better idea?

src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -850,6 +851,111 @@ class Pointer2Memref2PointerCast final
...
+      PointerUnion<IntegerAttr, Value> rawIdx = gep.getIndices()[0];
+      Value idx = dyn_cast_if_present<Value>(rawIdx);
+      if (!idx)
+        idx = rewriter.create<arith::ConstantIndexOp>(
+            loc, cast<IntegerAttr>(rawIdx).getValue().getSExtValue());
+      // TODO: verify the total byte offset will be element-aligned for dynamic
+      // indices by inserting runtime check
+      if (auto constIdx = idx.getDefiningOp<arith::ConstantIndexOp>()) {
+        // For constant indices, static check and reject unaligned access
+        if ((constIdx.value() * gepElemSize) % elementSize != 0) {
```
wsmoses, line 916: ah yeah here we may want to strengthen this and also support a case where we have e.g. i * 4, or i << 4, but we can do that as a follow up

wsmoses (approved):

PR #1136, opened 2025-07-01
[CI] Build wheel on aarch64-linux using lld as linker

### LINKED ISSUE #1124: Building wheel fails on aarch64-linux
https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/15960137557/job/45011484959#step:7:4193
```
ERROR: /home/runner/work/Enzyme-JAX/Enzyme-JAX/src/enzyme_ad/jax/BUILD:844:17: Linking src/enzyme_ad/jax/enzyme_call.so failed: (Exit 1): gcc failed: error executing CppLink command (from target //src/enzyme_ad/jax:enzyme_call.so) /usr/bin/gcc @bazel-out/aarch64-opt/bin/src/enzyme_ad/jax/enzyme_call.so-2.params

/usr/bin/ld.gold: error: Stub is too far away, try a smaller value for '--stub-group-size'. The current value is 0x7ffbffc.
/usr/bin/ld.gold: error: Stub is too far away, try a smaller value for '--stub-group-size'. The current value is 0x7ffbffc.
collect2: error: ld returned 1 exit status
```
There are _very_ few occurrences of similar errors on the internet, one thing I'd try when I have time (which is, not now) is to link with lld.  I'm even surprised (but maybe I shouldn't?) gold is used in the first place, since it's not maintained by anyone anymore.

giordano:
Should hopefully fix #1124.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,13 +51,19 @@ jobs:
           - '3.12'
     env:
       HERMETIC_PYTHON_VERSION: ${{ matrix.python }}
-      BAZEL_FLAGS : ${{ startsWith(matrix.os, 'macOS-') && '--define using_clang=true' || '' }}
     steps:
       - name: Make space on disk
         run: sudo rm -rf /opt/*
       - uses: actions/checkout@v4
         with:
           submodules: recursive
+      - name: Set environment variables
+        run: |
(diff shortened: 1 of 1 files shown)


giordano:
Superseded by #1173

PR #1151, opened 2025-07-09
Move Werror flags to CI

joelberkeley:
I'm not sure of either the Bazel syntax, nor the github actions syntax. I may need a few attempts to get this working

### DIFF
--- a/src/enzyme_ad/jax/BUILD
+++ b/src/enzyme_ad/jax/BUILD
@@ -487,14 +487,6 @@ cc_library(
     ]) + [
         "Utils.h",
     ],
-    copts = [
-        "-Werror=unused-variable",
-        "-Werror=unused-result",
-        "-Werror=missing-braces",
-        "-Werror=unused-but-set-variable",
-        "-Werror=return-type",
-        "-Werror=unused-result",
-    ],
     visibility = ["//visibility:public"],
     deps = [
         ":EnzymeHLOPatternsIncGen",
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,13 @@ jobs:
           - '3.12'
     env:
       HERMETIC_PYTHON_VERSION: ${{ matrix.python }}
-      BAZEL_FLAGS : ${{ startsWith(matrix.os, 'macOS-') && '--define using_clang=true' || '' }}
+      BAZEL_FLAGS : |
+        --copt=-Werror=unused-variable
+        --copt=-Werror=unused-result
+        --copt=-Werror=missing-braces
+        --copt=-Werror=unused-but-set-variable
+        --copt=-Werror=return-type
+        ${{ startsWith(matrix.os, 'macOS-') && '--define using_clang=true' || '' }}
     steps:
       - name: Make space on disk
         run: sudo rm -rf /opt/*


src/enzyme_ad/jax/BUILD
```diff
@@ -487,14 +487,6 @@ cc_library(
     ]) + [
         "Utils.h",
     ],
-    copts = [
```
wsmoses, line 490: So we shouldn't delete this for sure.

In principle this ought be sufficient to have any Bazel build run with these flags for the relevant code of the target. But apparently not.

Bazel with the -s flag may be helpful to see exactly what commands are run during compile

wsmoses: That said I'm okay also copying into the github actions if needed

wsmoses: The current CI says the GitHub actions level is too agressive as it adds it for all dependencies (like llvm) which we don't have control over, whereas the key here is making sure this repo doesn't introduce any unused variables (and thus potential accidental bugs)

joelberkeley: ok. this sounds like you're going for a different approach - fix CI so that errors in enzyme-jax don't get merged?

note that the enzyme-jax flags create errors when building stablehlo for my project, so getting it running here in CI might not be enough for my needs

joelberkeley: in that case I don't know how to fix it

wsmoses: so the stablehlo errors on your project shouldn't (in principle) be impacted by any of our flags here (and if they are we want to catch them here).

and yeah exactly, the goal here would be to make it such that the errors you're seeing never get merged (as ci fails and they get fixed in advance)

joelberkeley: how would you plan to handle the stablehlo errors? pin against a version of stablehlo in which you've fixed those errors?

I say your flags affect stablehlo because I only see build errors in stablehlo files if I keep your flags. If I delete them, then I see no stablehlo errors

wsmoses: so the error you saw (copying here):

```
external/stablehlo/stablehlo/dialect/Version.h: In constructor 'mlir::vhlo::Version::Version(int64_t, int64_t, int64_t)':
external/stablehlo/stablehlo/dialect/Version.h:77:9: error: missing braces around initializer for 'std::__array_traits<long int, 3>::_Type' {aka 'long int [3]'} [-Werror=missing-braces]
   77 |       : majorMinorPatch({major, minor, patch}) {}
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

makes sense that would be caught by the current flags (since its a header file we include in the target with the flags). Ideally ci would fail equally as your local build (which is what we should fix). and obviously we'll fix the stablehlo header upstream too (cc @GleasonK). we already pin against a stablehlo anyways (via xla pin via jax pin).

my worry from the failure here currently by putting the Werror to happen globally for all dependencies means that it isn't triggered by anything we control in this repo (either our source code, or in the above case a header we choose to include).

```
ERROR: /home/runner/.bazel/external/hwloc/BUILD.bazel:229:11: Compiling hwloc/memattrs.c failed: (Exit 1): gcc failed: error executing CppCompile command (from target @@hwloc//:hwloc) /usr/bin/gcc -U_FORTIFY_SOURCE -fstack-protector -Wall -Wunused-but-set-parameter -Wno-free-nonheap-object -fno-omit-frame-pointer -g0 -O2 '-D_FORTIFY_SOURCE=1' -DNDEBUG -ffunction-sections ... (remaining 37 arguments skipped)

external/hwloc/hwloc/memattrs.c: In function 'hwloc_memattr_get_initiators':
external/hwloc/hwloc/memattrs.c:770:9: error: unused variable 'err' [-Werror=unused-variable]
  770 |     int err = from_internal_location(&imi->initiator, &initiators[i]);
      |         ^~~
```

joelberkeley: ok, thanks. I'm not up for debugging bazel (it's a quick route to sadness), so I'll just wait to see what happens. For now, I'll just patch out those checks

PR #1152, opened 2025-07-10
Add support for non-constant bounds in do-while to for pattern

tyb0807:


test/lit_tests/canonicalizefor/doWhile.mlir
```diff
@@ -323,3 +323,184 @@ module @cmpi_ne_i{
...
+}
+
+// CHECK-LABEL: module @test_dynamic_ub {
+// CHECK:         func.func @do_while(%[[UB:.+]]: index) -> index {
+// CHECK-DAG:       %[[C0:.+]] = arith.constant 0 : index
+// CHECK-DAG:       %[[C1:.+]] = arith.constant 1 : index
+// CHECK-DAG:       %[[POISON:.+]] = ub.poison : index
+// CHECK-NEXT:      %[[COND:.+]] = arith.cmpi sgt, %[[UB]], %[[C1]] : index
+// CHECK-NEXT:      %[[IF:.+]] = scf.if %[[COND]] -> (index) {
+// CHECK-NEXT:        %[[FOR:.+]] = scf.for %[[IV:.+]] = %[[C0]] to %[[UB]] step %[[C1]] iter_args(%{{.*}} = %[[POISON]]) -> (index) {
```
wsmoses: instead of the if statement (which i realize is also a potential upgrade to other cases), here could we instead set the UB of the for loop to be max(1, UB) [prevents duplication and makes things a bit simpler]

tyb0807: Yup, good idea!

wsmoses (approved):

PR #1166, opened 2025-07-19
fix `isLegalConcatToOneDimDUS`

jumerckx:
It seems like `rhs` wasn't properly set.


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -12723,6 +12723,7 @@ bool isLegalConcatToOneDimDUS(stablehlo::ConcatenateOp outer,
         return false;
       }
     }
+    rhs = rhsSlice;
```
wsmoses, line 12726: Add a test?

wsmoses (approved):

PR #1172, opened 2025-07-23
Add a dialect-specific canonicalization pass

ftynse:
This only applies canonicalization patterns for ops from the dialects listed in pass options. Normally those ops list patterns for themselves, but it is not a strict requirement.

This offers significant performance bump compared to general canonicalization that collects patterns from all loaded dialects, many of which may not be present at certain stages of compilation. This wastes cycles trying to match patterns that would never apply.

### DIFF
--- a/src/enzyme_ad/jax/Passes/DialectCanonicalize.cpp
+++ b/src/enzyme_ad/jax/Passes/DialectCanonicalize.cpp
@@ -1,0 +1,56 @@
+//===- DialectCanonicalize.cpp - Dialect-restricted canonicalizer pass ----===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "src/enzyme_ad/jax/Passes/Passes.h"
+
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/DialectCanonicalize.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace {
+struct DialectCanonicalizePass
+    : mlir::enzyme::impl::DialectCanonicalizeBase<DialectCanonicalizePass> {
+  using DialectCanonicalizeBase::DialectCanonicalizeBase;
+
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    for (const std::string &dialectName : dialects) {
+      if (Dialect *dialect = getContext().getLoadedDialect(dialectName)) {
```
wsmoses, line 31: if there's any {llvm, enzymexla, memref, arith, affine} it may be worthwhile to load the whole set (per some patterns generating code in other dialects, and also enzymexla hijacking to add patterns applying to other dialects iirc)

ftynse: You can list those in pass options. As many as you want, there is no hardcoding here. But you definitely don't need tensor and tosa canonicalizers

wsmoses: ohh I read to quickly here, I thought this gathered the options based off of ops in IR, not a pass string, this makes more sense

wsmoses:
cc @avik-pal / @mofeing / @Pangoraw  we'll want to likely swap to this [and figure out the relevant pass set] during raising

PR #1173, opened 2025-07-26
Use new runners

wsmoses:


.github/workflows/build.yml
```diff
@@ -38,14 +38,96 @@ jobs:
...
+        if: ${{ !cancelled() }} 
+        with:
+          name: Enzyme-JAX-tests-${{ matrix.os }}-${{ matrix.python }}
+          path: |
+            bazel-out/**/test.log
+          retention-days: 90
+
+  build-macos:
+    name: Build - ${{ matrix.os }} - python ${{ matrix.python }}
+    runs-on: ${{ matrix.os }}
```
giordano: Why duplicating the workflow?

wsmoses: the macos needs an empty container [and i didnt get it to work otherwise, if you found a fix]

giordano: Yes, I did it in https://github.com/EnzymeAD/Reactant.jl/pull/1478: https://github.com/EnzymeAD/Reactant.jl/blob/3258f95ea931b912b9f653cfed62d6aec56c6e36/.github/workflows/CI.yml#L37

test/test_utils.py
```diff
@@ -26,6 +26,8 @@ def fix_paths():
     runfiles = os.path.dirname(
         os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
     )
+    os.system("find " + runfiles)
+
```
giordano:
```suggestion
```

Issue #1174, opened 2025-07-26
`error: 'stablehlo.multiply' op requires compatible types for all operands and results` in GB-25 code


giordano:
With Reactant v0.2.147 (corresponding diff in this repo: https://github.com/EnzymeAD/Enzyme-JAX/compare/6774f1afb90c377bbf234a7a7dbfab4f7b726481...05b4c93ab2f581eae42f44599a1993311ad72c63), GB-25 code fails to compile with [error](https://github.com/PRONTOLab/GB-25/actions/runs/16537419612/job/46773948966#step:10:593)
```
[...]

PR #1176, opened 2025-07-26
Fix concat to dus

wsmoses:


wsmoses:
fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/1174

PR #1232, opened 2025-08-02
fix: condition direction in compare simplify

avik-pal:


wsmoses (approved):
@avik-pal if this is good to merge [and the correct one], go for it

PR #1234, opened 2025-08-02
[CI] Install `Reactant#main` instead of non-existing commit

giordano:


giordano:
Yes, it's green again. Let's merge it so that this tests can be useful again.

PR #1243, opened 2025-08-02
Run GB-25 tests with larger grid

giordano:
Just an experiment for now, let's see how it goes.


giordano:
I expect this to use ~20 GB of memory on each GPU, this should be a first test to push the grid size up to fill up the memory, if there aren't problems.

PR #1259, opened 2025-08-05
GPU lowering fixes

pengmai:
* fix swapped threadIdx.x/y in kernel launch
* fix SimplyDeadAlloc being too aggressive with enzymexla.memcpy

### DIFF
--- a/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
@@ -1621,7 +1621,7 @@ struct SimplifyDeadAlloc : public OpRewritePattern<T> {
       }
 
       if (auto cpy = dyn_cast<enzymexla::MemcpyOp>(op)) {
-        if (cpy.getTarget() == alloc->getResult(0))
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
```diff
@@ -1621,6 +1621,8 @@ struct SimplifyDeadAlloc : public OpRewritePattern<T> {
       }
 
       if (auto cpy = dyn_cast<enzymexla::MemcpyOp>(op)) {
+        if (cpy.getSource() == alloc->getResult(0))
```
wsmoses, line 1624: we should delete the target one

wsmoses: and do you mind adding a quick test for this to avoid it recurring (and especially the threadid one)

pengmai: Yup, tests have been added for both the source/target cases for memcpy. A test already exists for the threadId lowering, so that has been updated.

Issue #1183, opened 2025-07-27
Scatter split


wsmoses:
a single scatter that results in two independent results, should be split into individual scatter operations (see below)

```mlir
  func.func private @main_2(%arg0: tensor<2xf32> , %arg1: tensor<2xf32>) -> (tensor<3xf32>, tensor<3xf32>) {
    %cst = stablehlo.constant dense<0.000000e+00> : tensor<3xf32>
    %c = stablehlo.constant dense<[[0], [1]]> : tensor<2x1xi32>
    
    %0:2 = "stablehlo.scatter"(%cst, %cst, %c, %arg0, %arg1) <{indices_are_sorted = false, scatter_dimension_numbers = #stablehlo.scatter<inserted_window_dims = [0], scatter_dims_to_operand_dims = [0], index_vector_dim = 1>, unique_indices = true}> ({

    ^bb0(%arg2: tensor<f32>, %arg3: tensor<f32>, %arg4: tensor<f32>, %arg5: tensor<f32>):
      stablehlo.return %arg4, %arg5 : tensor<f32>, tensor<f32>
    }) : (tensor<3xf32>, tensor<3xf32>, tensor<2x1xi32>, tensor<2xf32>, tensor<2xf32>) -> (tensor<3xf32>, tensor<3xf32>)

    return %0#0, %0#1 : tensor<3xf32>, tensor<3xf32>
  }

```

partial (not complete) workaround for https://github.com/openxla/xla/issues/29362

PR #1260, opened 2025-08-06
SplitMultiResultScatter

jumerckx:
fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/1183

Someone that knows the scatter operation better should probably check that my implementation makes sense.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -20426,6 +20426,79 @@ struct SplitConvolutionIntoReverseConvolution final
   }
 };
 
+// partial workaround for: https://github.com/openxla/xla/issues/29362
+struct SplitMultiResultScatter
+    : public CheckedOpRewritePattern<stablehlo::ScatterOp,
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -20426,6 +20426,79 @@ struct SplitConvolutionIntoReverseConvolution final
...
+    if (!returnOp || returnOp.getNumOperands() != op.getNumResults()) {
+      return rewriter.notifyMatchFailure(
+          op, "scatter body does not have a suitable return op");
+    }
+
+    // The arguments to the body are first the operands, then the updates.
+    // We expect the return op to return the update arguments.
+    size_t numInputs = op.getInputs().size();
+    for (size_t i = 0; i < returnOp.getNumOperands(); ++i) {
+      auto returnedVal = dyn_cast<BlockArgument>(returnOp.getOperand(i));
```
wsmoses: we should generalize this a bit. we want to ideally support cases that aren't just returning a blockargument, but where the result doesn't depend on the other operand. Look at forward/backwards slice for some relevant code

PR #1274, opened 2025-08-07
feat: detect and rewrite elementwise as reduce_window/conv

avik-pal:


.github/workflows/test-gb-25.yml
```diff
@@ -53,8 +53,8 @@ jobs:
           - 'main'
           # - '0123456789abcdef0123456789abcdef01234567'
         reactant_commit:
-          - 'main'
-          # - "ap/updated_no_nan"
+          # - 'main'
```
giordano, line 56: Keep main? This commit doesn't exist anymore

avik-pal: testing that this pr doesn't mess up correctness, will revert before merging

giordano: But the commit doesn't exist, this won't work at all. **Edit**: ah, changed to the branch now

avik-pal:
> Relative tolerance is sqrt(eps(T)) (~1e-8 for Float64), it's not particularly strict.

In cudnn non-deterministic mode (default) results don't match to that tolerance across sessions

PR #1275, opened 2025-08-08
Do not create undef init values for variables selected from init vars

tyb0807:


src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1456,6 +1456,15 @@ struct MoveWhileToFor : public OpRewritePattern<WhileOp> {
         if (blockArg.getOwner() == &condOp->getParentRegion()->front()) {
           newArg = loop.getOperand(blockArg.getArgNumber());
         }
+      } else if (auto selectOp = arg.getDefiningOp<arith::SelectOp>()) {
+        auto trueBlockArg = dyn_cast<BlockArgument>(selectOp.getTrueValue());
+        auto falseBlockArg = dyn_cast<BlockArgument>(selectOp.getFalseValue());
+
+        if (trueBlockArg && !falseBlockArg) {
```
wsmoses: we don't need to check the !falseBlockAarg here, replacing poison with something is fine either way

wsmoses: while we're at it, we might as well change Value newArg = nullptr;

and put the poisonop at the end of this with `if (!newArg) newArg = create poison`

wsmoses:
@tyb0807 I think this now needs a rebase

PR #1281, opened 2025-08-08
use addrspacecast insteadof bitcast for pointers

ftynse:
We used to be bitcating pointers because of typed pointers, which are now gone, but we are getting address space mismatches. Repurpose bitcast insertion for address casts. Not sure this is the right approach or whether we should change the function signature to accept pointers in another address spaces.


wsmoses (approved):
if you have a repro we can add as a test to prevent recurring, that owuld be helpful

PR #1282, opened 2025-08-08
bump minimum required default gpu versions

ftynse:
We are using intrinsics that require these.


wsmoses (approved):
fair, though also this should be auto detected previously from the imported IR at some phase? or at least I thought I set that up to happen

Issue #1238, opened 2025-08-02
slice of extend


wsmoses:
```
      %1076 = "enzymexla.extend"(%1047) <{dimension = 0 : i64, lhs = 1 : i64, rhs = 0 : i64}> : (tensor<4x1520x3056xf64>) -> tensor<5x1520x3056xf64> loc(#loc4184)
      %1077 = stablehlo.slice %1076 [0:3, 0:1520, 0:3056] : (tensor<5x1520x3056xf64>) -> tensor<3x1520x3056xf64> loc(#loc4184)
      %2494 = stablehlo.slice %1076 [0:4, 0:1520, 0:3056] : (tensor<5x1520x3056xf64>) -> tensor<4x1520x3056xf64> loc(#loc3406)
```

we can slice 1047 up to [0:3] here before extending

PR #1296, opened 2025-08-11
ExtendSlice

jumerckx:
fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/1238

Should there be a guard for the case where there are multiple slices being taken from the extend?


wsmoses:
actually sorry thought was reviewed, yes there should be such a guard, going to revert for now if you can remake the PR

jumerckx:
> actually sorry thought was reviewed, yes there should be such a guard, going to revert for now if you can remake the PR

Will do once I get back from holiday, Tuesday!

PR #1297, opened 2025-08-11
[CI] Use again `Reactant#main` for GB-25 test

giordano:
With frequent rebases of the branch, using commit hashes is a minefield.


giordano:
I'm merging this PR now that the GB-25 test is finally green again.

PR #1299, opened 2025-08-12
feat: Support for Add operation in AutoDiffScatterRev

JulianTrommer:
This PR implements the Add operation for the reverse diff of the scatter operation.

I also implemented a test case that runs successfully.
To validate if the gradient is working correctly I also created a pull request to Reactant.jl to test the functionality:
https://github.com/EnzymeAD/Reactant.jl/pull/1545
It would be great if someone could confirm if the implementation is correct.


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -2422,9 +2422,11 @@ class AutoDiffScatterRev
                                          MGradientUtilsReverse *gutils,
                                          SmallVector<Value> caches) const {
     auto scatterOp = cast<ScatterOp>(op);
+    Operation &innerOp = scatterOp.getUpdateComputation().front().front();
 
-    if (!stablehlo::isScatterSetindexOp(scatterOp)) {
-      op->emitError("AutoDiffScatterRev only supports Setindex operations")
+    if (!stablehlo::isScatterSetindexOp(scatterOp) && !isa<AddOp>(innerOp)) {
```
wsmoses: sorry for delay -- we were fixing CI

we might also need to check which operands are being added, and that add is the only op

avik-pal: Added the functionality for the check in #1333

JulianTrommer: I will close the PR since the functionality is in #1333.
Thanks for adding it!

PR #1300, opened 2025-08-12
Raise __enzyme_autodiff calls to enzyme.autodiff ops

pengmai:


src/enzyme_ad/jax/Passes/LibDeviceFuncsRaisingPass.cpp
```diff
@@ -179,6 +184,117 @@ class CallToOpRaising : public OpRewritePattern<LLVM::CallOp> {
...
+      return failure();
+    }
+
+    FlatSymbolRefAttr funcToDiffSymbol = targetAddr.getGlobalNameAttr();
+    auto mod = op->getParentOfType<ModuleOp>();
+    auto funcToDiff =
+        cast<FunctionOpInterface>(mod.lookupSymbol(funcToDiffSymbol));
+
+    // Infer the argument activities
+    SmallVector<Value> arguments;
```
wsmoses, line 288: mind moving this to a helper function (that way it can be reused for fwddiff op raising)

PR #1312, opened 2025-08-16
[CI] Pass more options to Julia processes running simulations

giordano:
* Fail immediately if pkgimages aren't precompiled yet
* Don't load startup file
* Use multiple threads (not sure this helps anything, but shouldn't hurt)

### DIFF
--- a/.github/workflows/test-gb-25.yml
+++ b/.github/workflows/test-gb-25.yml
@@ -179,12 +179,12 @@ jobs:
         timeout-minutes: 60
         run: |
           export XLA_FLAGS='--xla_dump_to=${{ env.GB25_DIR }}/xla_dump'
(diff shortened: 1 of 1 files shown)


.github/workflows/test-gb-25.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           - 'main'
           # - '0123456789abcdef0123456789abcdef01234567'
         reactant_commit:
-          - 'main'
+          - '33a3d346572391b31749b172636a38eeb15593dc'
```
giordano: This can be reverted after https://github.com/EnzymeAD/Reactant.jl/pull/1557 is merged.

wsmoses (approved):

giordano:
This is still segfaulting 😕

PR #1307, opened 2025-08-15
Update the xla commit to experiment potential fix for OOM

felixwqp:
Try to fix the large number of zero-broadcast generated from big pad in spmd partitioner. 

for sample HLO, 
```
  HloModule module_0076.reactant_loop_after_spmd_partitioner_cp_pattern

    ENTRY main() -> f64[3056,123] {
      %constant.946 = f64[] constant(0)
      %subtract.1055 = f64[3056,12272]{1,0} parameter(0), sharding={devices=[2,2]<=[2,2]T(1,0)}
      %slice.1057 = f64[3056,1]{1,0} slice(%subtract.1055), slice={[0:3056], [12271:12272]}, sharding={devices=[2,2]<=[2,2]T(1,0)}
      ROOT %pad.1059 = f64[3056,123]{1,0} pad(%slice.1057, %constant.946), padding=0_0x0_122, sharding={devices=[2,2]<=[2,2]T(1,0)}
    }
```

before the fix, the result looks like:  4LjPYjEP2XJGwcn

after the fix, the result looks like: https://screenshot.googleplex.com/
3BZP5MRC5kMpK2L

PR #1318, opened 2025-08-19
Bump jax commit

wsmoses:


giordano:
Does this include https://github.com/openxla/xla/pull/30307? The simulation is as fast as #1307 😅

wsmoses:
yes it does (and I'm checking no other commit broke us, before making a jll)

PR #1319, opened 2025-08-19
ExtendSlice 2

jumerckx:
fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/1238

Now includes a guard to only apply the rewrite if the SliceOp is the only user of the ExtendOp. (https://github.com/EnzymeAD/Enzyme-JAX/pull/1296)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -18759,6 +18759,70 @@ bool isAxisFusible(int dimension, ArrayRef<Value> vals) {
...
+    int64_t new_lhs = std::max((int64_t)0, std::min(limit_d, lhs) - start_d);
+    // new_rhs is the size of the overlap between the slice and the appended
+    // padding.
+    int64_t new_rhs =
+        std::max((int64_t)0, limit_d - std::max(start_d, lhs + size_d));
+
+    // Create the new slice on the original tensor.
+    auto newSlice = rewriter.create<stablehlo::SliceOp>(
+        op.getLoc(), operand, new_starts, new_limits, new_strides);
+
```
wsmoses, line 18824: if new_lhs and new_rhs are 0 [even if multiple users], we can just use the sliced operand directly instead of the extend

Issue #1289, opened 2025-08-09
[CI] Failed jaxmd test


giordano:
https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/16713928087/job/47317194253#step:15:87
```
[7,950 / 7,951] 586 / 587 tests, 1 failed; Testing //test:jaxmd; 64s disk-cache, processwrapper-sandbox
INFO: From Testing //test:jaxmd:
==================== Test output for //test:jaxmd:
Running tests under Python 3.11.10: /root/.bazel/sandbox/processwrapper-sandbox/13802/execroot/__main__/bazel-out/k8-opt/bin/test/jaxmd.runfiles/python_x86_64-unknown-linux-gnu/bin/python3
INFO:2025-08-04 10:00:38,599:jax._src.xla_bridge:830: Unable to initialize backend 'tpu': INTERNAL: Failed to open libtpu.so: libtpu.so: cannot open shared object file: No such file or directory
I0804 10:00:38.599670 133472947390272 xla_bridge.py:830] Unable to initialize backend 'tpu': INTERNAL: Failed to open libtpu.so: libtpu.so: cannot open shared object file: No such file or directory
[ RUN      ] test_utils.EnzymeJaxTest.test
[       OK ] test_utils.EnzymeJaxTest.test
[ RUN      ] JAXMD.test
/root/.bazel/sandbox/processwrapper-sandbox/13802/execroot/__main__/bazel-out/k8-opt/bin/test/jaxmd.runfiles/pypi_jax_md/site-packages/jax_md/space.py:279: DeprecationWarning: invalid escape sequence '\i'
  """Periodic boundary conditions on a parallelepiped.
/root/.bazel/sandbox/processwrapper-sandbox/13802/execroot/__main__/bazel-out/k8-opt/bin/test/jaxmd.runfiles/pypi_dm_haiku/site-packages/haiku/_src/transform.py:341: DeprecationWarning: jax.lib.xla_extension.PjitFunction is deprecated.
  COMPILED_FN_TYPES = (jax.lib.xla_extension.PjitFunction,
/root/.bazel/sandbox/processwrapper-sandbox/13802/execroot/__main__/bazel-out/k8-opt/bin/test/jaxmd.runfiles/pypi_dm_haiku/site-packages/haiku/_src/transform.py:342: DeprecationWarning: jax.lib.xla_extension.PmapFunction is deprecated.
[...]

PR #1326, opened 2025-08-19
Increase tolerance for jaxmd test

giordano:
Ref #1289.

### DIFF
--- a/test/jaxmd.py
+++ b/test/jaxmd.py
@@ -121,7 +121,7 @@ class JAXMD(EnzymeJaxTest):
         self.mlirad_fwd = False
 
         if jax.default_backend() == "tpu":
-            self.tol = 5e-4
+            self.tol = 5e-3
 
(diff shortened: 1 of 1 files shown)


test/jaxmd.py
```diff
@@ -125,7 +125,8 @@ def forward(
         # GPU CI reverse mode needs loose, merits future investigation
         self.tol = 1e-2
 
-        self.tol = 0.04
+        # CPU backend needs loose tolerance, see #1289
```
giordano: Maybe we can have different tolerance for the different backends?

wsmoses (approved):

test/jaxmd.py
wsmoses: fair, that's likely useful so we can keep tests strict when possible  [already some use backend to swap # of iterations]

giordano: Tests are passing with backend-specific tolerance, yay

PR #1329, opened 2025-08-20
Cscf

tyb0807:


src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -314,6 +314,21 @@ struct ForOpInductionReplacement : public OpRewritePattern<scf::ForOp> {
           sameValue |= rattr.zext(maxWidth) == sattr.zext(maxWidth);
         }
 
+      // Helper function to cast value to target type
+      auto castValue = [&](Value value, Value target, Location loc) -> Value {
```
wsmoses: for compile time perf can we make this a non anonymous function?

PR #1331, opened 2025-08-21
fix: check valueAttr is a ElementsAttr

avik-pal:
Not too sure, but this seems to be the source of https://github.com/EnzymeAD/Reactant.jl/pull/1536


wsmoses (approved):
weird, but sure

PR #1333, opened 2025-08-21
feat: extend broadcastreduce to support min/max/add

### LINKED ISSUE #1325: reduce multiply ( bcast_in_dim) to power
```mlir
module @reactant_fn attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<f32>) -> tensor<f32> {
    %cst = stablehlo.constant dense<1.000000e+00> : tensor<f32>
    %0 = stablehlo.broadcast_in_dim %arg0, dims = [] : (tensor<f32>) -> tensor<3x4xf32>
    %1 = stablehlo.reduce(%0 init: %cst) applies stablehlo.multiply across dimensions = [0, 1] : (tensor<3x4xf32>, tensor<f32>) -> tensor<f32>
    return %1 : tensor<f32>
  }
}
```

We already do this for addition

avik-pal:
Some of the older checks seem incorrect, it would match for any reduce region with the first op being `add`.

Also generalized the `broadcastreduce` pattern to support min/max/mul (which at one point we should rename to `reducebroadcast`). Fixes #1325

### DIFF
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -2423,7 +2423,8 @@ public:
                                          SmallVector<Value> caches) const {
     auto scatterOp = cast<ScatterOp>(op);
 
-    auto checkCommonScatterOp = mlir::stablehlo::CheckCommonScatterOp(scatterOp);
+    auto checkCommonScatterOp =
+        mlir::stablehlo::CheckCommonScatterOp(scatterOp);
 
     if (!checkCommonScatterOp.isSetindexScatter) {
       op->emitError("AutoDiffScatterRev only supports Setindex operations")


avik-pal:
now how did I trigger http://github.com/EnzymeAD/Enzyme-JAX/actions/runs/17140660776/job/48627249621?pr=1333#step:18:745??

wsmoses (approved):

wsmoses:
that's extremely weird, but i suppose the same suggestion of finding the path and copying to right place applies

giordano:
Is this a problem only in the [`build_local.jl`](https://github.com/EnzymeAD/Reactant.jl/blob/e7b0a5ae9d3daf45ccd55a33094434adf156b99e/deps/build_local.jl) script? In Yggdrasil [we do copy libdevice](https://github.com/JuliaPackaging/Yggdrasil/blob/1957f2a42fa57eb9462a138eed0517ac81159078/R/Reactant/build_tarballs.jl#L360)

wsmoses:
yeah we might want to similarly copy it to a local directory of similar name

giordano:
Addressed in #1335 (and https://github.com/EnzymeAD/Reactant.jl/pull/1579 for full resolution in Reactant, but I haven't tested that one yet)

PR #1356, opened 2025-08-31
Test BinaryBuilderBase

giordano:
This is to check whether we can run BinaryBuilderBase on the runners.

### DIFF
--- a/.github/workflows/binarybuilder.yml
+++ b/.github/workflows/binarybuilder.yml
@@ -18,6 +18,7 @@ jobs:
 
     container:
       image: ${{ (contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images@sha256:91e1edb7a7c869d5a70db06e417f22907be0e67ca86641d48adcea221fedc674' ) || '' }}
+      options: --privileged
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


giordano:
@staticfloat any clue of what could be causing https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/17363797677/job/49287417608?pr=1356#step:7:71
```
At line 1120, ABORTED (1: Operation not permitted)!
```
?  I'm able to reproduce the error inside the Docker container
```
docker pull ghcr.io/enzymead/reactant-docker-images:main
```
We have full control over the [Docker image](https://github.com/EnzymeAD/reactant-docker-images/blob/156807f0061c1fe4d77a02695c9035b98d7d37ec/Dockerfile), so if we need to fix anything inside that we can do it.

staticfloat:
I suspect the issue is that even though you're running an inner docker instance with `--privileged`, there's an outer docker that is restricting your privileges, or some other kind of privilege restriction.

staticfloat:
You could try exporting `FORCE_SANDBOX_MODE=unprivileged` to force it to use unprivileged mode, which is the only mode that has a hope of working here, I think.

staticfloat:
https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/17388247260/job/49358059371?pr=1356#step:7:31 :(

giordano:
That's expected (and seems to be an upstream problem in Rust): https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/395. I thought we were skipping the test when known to be broken, but I may be wrong

giordano:
Ok, test successful here so far, closing this PR.

PR #1357, opened 2025-09-01
Update JAX to commit a8e73333b8dddb6bd248765e3f7c08f72bb8fcfb

giordano:
Diff: https://github.com/jax-ml/jax/compare/a5e4e80057c3c18c566a4f4079d933386d9bfeb8...a8e73333b8dddb6bd248765e3f7c08f72bb8fcfb


wsmoses:
yeah we should remove that flag from our end here [and just not set it]

PR #1361, opened 2025-09-03
Update JAX to commit 6c4af5f0968fc658356a915c3e63b58468a030f5

giordano:
Diff: https://github.com/jax-ml/jax/compare/a8e73333b8dddb6bd248765e3f7c08f72bb8fcfb...6c4af5f0968fc658356a915c3e63b58468a030f5


giordano:
```
ERROR: Error computing the main repository mapping: cannot load '@@jax//:nvidia_wheel_versions.bzl': no such file
```

giordano:
Can we just delete
```
https://github.com/EnzymeAD/Enzyme-JAX/blob/ecf3bc91fa861f92629ba76dbca147e5c515536c/WORKSPACE#L97
```
?

giordano:
Ah, of course the update overwrote the fix (https://github.com/EnzymeAD/Enzyme-JAX/commit/a38a8f679d99f6fb541470985c0cbd17f1676ac7), however we also need to apply the same fix to Reactant, how do we handle this?

PR #1375, opened 2025-09-07
fix: check for perm and ensure no regressions

### LINKED ISSUE #1374: `transpose_reshape` singleton dims case busted
https://github.com/EnzymeAD/Reactant.jl/actions/runs/17524945837/job/49774098408?pr=1632#step:6:239

will investigate tomorrow morning

avik-pal:
fixes #1374

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -21500,7 +21500,7 @@ struct TransposeReshape final
       int64_t index = std::distance(permutation.begin(), idx);
 
       for (int i = 1; i < dimList.size(); ++i) {
-        if (i + index > permutation.size() ||
+        if (i + index >= permutation.size() ||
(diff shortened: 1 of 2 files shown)


wsmoses (approved):

giordano:
Again
```
         nested task error: MethodError: no method matching String(::Vector{String})
        The type `String` exists, but no method is defined for this combination of argument types when trying to construct it.
        
        Closest candidates are:
          String(::Vector{UInt8})
           @ Base strings/string.jl:72
          String(::SubString{String})
           @ Base strings/substring.jl:75
          String(::Core.Compiler.LazyString)
           @ Core strings/lazy.jl:80
          ...
        
        Stacktrace:
         [1] (::CUDA_Driver_jll.var"#4#11"{Pipe})()
           @ CUDA_Driver_jll ~/.julia/packages/CUDA_Driver_jll/iYAlY/src/wrappers/x86_64-linux-gnu.jl:165
```
I'm so confused by what's going on there.  It works for me on Hydra though.

PR #1381, opened 2025-09-08
Update JAX to commit f74467851b1186b434d4b538d0be419378a47a69

giordano:
Diff: https://github.com/jax-ml/jax/compare/a4c1bee798901d5ee39c9645b090542fed8c5576...f74467851b1186b434d4b538d0be419378a47a69


wsmoses:
officially, no

avik-pal:
> > Benchmark suite 	Current: [a58aa0a](https://github.com/EnzymeAD/Enzyme-JAX/commit/a58aa0a5046818da5a59605551180d0afc40519e) 	Previous: [53bc404](https://github.com/EnzymeAD/Enzyme-JAX/commit/53bc40491de803500811c5e93796de9bbef39642)
> 
> @avik-pal does it make sense to compare two different PRs (and which is the criterion for picking up a specific PR? just whatever happens to be have been last touched?)? Shouldn't the baseline be something more stable like `main`?

It is using the current PR vs the last stored run numbers which will be main (https://github.com/EnzymeAD/Enzyme-JAX/commit/53bc40491de803500811c5e93796de9bbef39642). For PRs we never store the numbers so it wont show up in comparisons

PR #1383, opened 2025-09-11
Dynamic sized while op bound

wsmoses:


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -3373,9 +3373,9 @@ struct WhileOpEnzymeOpsRemover
 
     // TODO: support non-constant loops by using a dynamic dimension
     // ...   should we fail ? i.e. return failure();
```
Pangoraw: This comment can be removed ✅

PR #1328, opened 2025-08-20
ci: upload test timings & add comment

avik-pal:

PR #1394, opened 2025-09-16
Prototype distributed dialect ops

EganJ:


giordano:
@avik-pal the benchmark reporting should try to push when it doesn't have the permissions to do so in the first place (which I also mentioned in https://github.com/EnzymeAD/Enzyme-JAX/pull/1328#issuecomment-3217457044): https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/17768182534/job/50502847013?pr=1394

src/enzyme_ad/jax/Dialect/Distributed/Ops.cpp
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Returns success if the symbol reference attribute refers to an operation of
+ * type T in the given symbol table.
+ */
+template <typename T>
+llvm::LogicalResult checkSymbolIsA(::mlir::SymbolTableCollection &symbol_table,
+                                   mlir::Operation *owning_op,
+                                   mlir::SymbolRefAttr attr) {
+  auto loc = owning_op->getLoc();
+  Operation *symOp = symbol_table.lookupNearestSymbolFrom(owning_op, attr);
```
wsmoses, line 19: x/ref dyn_cast_or_null

test/lit_tests/distributed/parsetest.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: enzymexlamlir-opt %s
+distributed.LeafDevice @myGpu
+distributed.DeviceMesh @gpuMesh @myGpu [2, 2]
```
wsmoses: can you rename to roundtrip.mlir and add checks

PR #1404, opened 2025-09-20
fix: add check to prevent partial slices in batching

avik-pal:


giordano:
Should be fixed now, the cleanup step is now trying a lot harder to remove the existing files.

wsmoses (approved):

PR #1405, opened 2025-09-20
Fix batched abs op

wsmoses:


wsmoses:
yeah as I expected, I think I need to update tablegen under the hood to handle this case

wsmoses:
[this is the correct change here but I need to debug enzyme proper to see why tblgen doesn't like this]

PR #1410, opened 2025-09-22
fix: correct sort ordering of the operations

avik-pal:


avik-pal:
need to see why jaxmd is sad

wsmoses (approved):

PR #1411, opened 2025-09-22
affine-to-stablehlo: stores with multiple values as scatter

Pangoraw:


test/lit_tests/raising/affine_to_stablehlo_store_multi_dims.mlir
```diff
@@ -0,0 +1,30 @@
+// RUN: enzymexlamlir-opt %s --raise-affine-to-stablehlo --canonicalize --enzyme-hlo-opt=max_constant_expansion=0 | FileCheck %s
+
+module {
+  func.func @main(%arg0: memref<100xf32>, %arg1: memref<100xf32>) {
+    affine.parallel (%i, %j) = (0, 0) to (100, 10) step (10, 1) {
+      %0 = affine.load %arg1[%i + %j] : memref<100xf32>
```
wsmoses: Can we do multiplied by some value just to ensure the right index car is modified

wsmoses (approved):

PR #1435, opened 2025-09-28
Remove call to deprecated createTargetMachine overload

boomanaiden154:
The overloads for many of the TargetRegistry functions using string triples are deprecated and will be removed soon. Migrate Enzyme off them to prevent breakage.


boomanaiden154:
@wsmoses for review. This also ideally needs to get imported int google3 reasonably soon so #161053 can land.

wsmoses (approved):

PR #1439, opened 2025-10-02
feat: function argument memory effects for kernels

avik-pal:


src/enzyme_ad/jax/Passes/Passes.td
```diff
@@ -78,6 +78,12 @@ def MarkFunctionMemoryEffectsPass : Pass<"mark-func-memory-effects", "ModuleOp">
       /*description=*/"assume no memory effects for ops not implementing MemoryEffectOpInterface">];
 }
 
+def MarkFunctionArgumentsMemoryEffectsPass : InterfacePass<"mark-func-args-memory-effects",
+                                                           "mlir::FunctionOpInterface"> {
+  let summary = "Mark function arguments with enzymexla.memory_effects attribute";
```
wsmoses: can this just be merged in the same memory effects pass?

src/enzyme_ad/jax/Passes/MarkFunctionArgumentsMemoryEffectsPass.cpp
```diff
@@ -0,0 +1,160 @@
...
+    setAllMemoryEffects(effects);
+    return;
+  }
+
+  SmallVector<MemoryEffects::EffectInstance> memEffects;
+  memInterface.getEffects(memEffects);
+
+  for (const auto &effect : memEffects) {
+    if (effect.getValue() && effect.getValue() == operand->get()) {
+      if (isa<MemoryEffects::Read>(effect.getEffect())) {
```
wsmoses: its probably also worthwhile detecting if something is nocapture

test/lit_tests/memoryeffects/readonly.mlir
```diff
@@ -0,0 +1,49 @@
...
+    %12 = arith.addf %8, %11 : tensor<64xf32>
+    %13 = tt.splat %arg2 : !tt.ptr<f32> -> tensor<64x!tt.ptr<f32>>
+    %14 = tt.addptr %13, %4 : tensor<64x!tt.ptr<f32>>, tensor<64xi32>
+    tt.store %14, %12, %5 : tensor<64x!tt.ptr<f32>>
+    tt.return
+  }
+}
+
+// CHECK: module {
+// CHECK-NEXT:   tt.func @main(%arg0: !tt.ptr<f32> {enzymexla.memory_effects = ["read"]}, %arg1: !tt.ptr<f32> {enzymexla.memory_effects = ["read"]}, %arg2: !tt.ptr<f32> {enzymexla.memory_effects = ["write"]}) attributes {noinline = false} {
```
wsmoses: if we're really lazy we can potentially alternatively set llvm.readonly llvm.writeonly and/or llvm.readnone

src/enzyme_ad/jax/Passes/MarkFunctionMemoryEffectsPass.cpp
```diff
@@ -237,26 +411,34 @@ struct MarkFunctionMemoryEffectsPass
...
-            assert(false && "unknown memory effect");
-          }
+      auto funcEffectInfo = getEffectInfo(builder, effectsSet);
+      funcOp->setAttr("enzymexla.memory_effects",
+                      funcEffectInfo.enzymexlaEffects);
+
+      auto &argEffects = funcArgEffects[symbol];
+      for (unsigned i = 0; i < funcOp.getNumArguments(); i++) {
+        auto argEffectInfo = getEffectInfo(builder, argEffects[i]);
+        funcOp.setArgAttr(i, "enzymexla.memory_effects",
```
wsmoses, line 428: potentially a follow up pr, but nocapture is likely to be immensely useful too [aka, for a pointer/memref input, it isn't stored or returned]

wsmoses (approved):

avik-pal:
If there is an op storing a ptr `llvm.store %arg0, %9 : !llvm.ptr<1>, !llvm.ptr` where `%arg0` is a ptr, is `arg0` still readnone?

wsmoses:
LLVM doesn't auto deduce it in that case: https://godbolt.org/z/K75cfP8qP

but per the IR reference (https://llvm.org/docs/LangRef.html) I think it does apply if provably no possible load from  %9 is itself written into.

Aka readonly/etc only applies to "could this specific function" cause a write/read.

wsmoses:
attributor is slightly more aggressive and says its only possibly writing:

https://godbolt.org/z/sMoKWxd4s

wsmoses:
FYI having both readonly and writeonly may be invalid IR (I know for a fact readonly and readnone is)

PR #1446, opened 2025-10-04
fix: add a separate check for pad_dot_general

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -9667,12 +9667,15 @@ struct PadDotGeneral
       nextOtherArg = slice.getResult();
     }
 
+    auto lhs = otherIsLHS ? nextOtherArg : pad.getOperand();
+    auto rhs = otherIsLHS ? pad.getOperand() : nextOtherArg;
+    if (!areValidDotGeneralInputs(lhs, rhs, op.getDotDimensionNumbersAttr()))
```
wsmoses, line 9672: this feels like the wrong solution. shouldn't we just slice the inputs properly here?

PR #1453, opened 2025-10-05
feat: propagate bounds in whileop

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -23332,6 +23334,55 @@ struct DynamicPadToPad
...
+    auto inductionVar = whileBody.getArgument(0);
+
+    bool anyNoOpRemoved = false;
+
+    auto limit = info.getConstantLimit().value();
+    auto start = info.getConstantStart().value();
+    auto step = info.getConstantStep().value();
+
+    // currently only removes remainder of induction variable
+    whileBody.walk([&](stablehlo::RemOp remOp) -> WalkResult {
```
wsmoses: instead could we just look at all users of the induction var (that likely will be faster than a walk of the body)

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -23332,6 +23334,55 @@ struct DynamicPadToPad
...
+      if (remOp.getLhs() == inductionVar) {
```
wsmoses: we can also then reuse this to also do things like remove comparisions >= start or < limit

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -23332,6 +23334,339 @@ struct DynamicPadToPad
...
+    auto inductionType = inductionVar.getType();
+    unsigned bitWidth = 64;
+    if (auto tensorType = dyn_cast<RankedTensorType>(inductionType)) {
+      if (auto intType = dyn_cast<IntegerType>(tensorType.getElementType())) {
+        bitWidth = intType.getWidth();
+      }
+    }
+
+    // Initialize bounds map with induction variable bounds
+    DenseMap<Value, Bounds> boundsMap;
```
wsmoses, line 23372: min/max only applies if step >= 0, if negative this needs to be changed

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -23332,6 +23334,339 @@ struct DynamicPadToPad
...
+        rhs_bounds.max.sle(lhs_bounds.max)) {
+      auto lhs = remOp.getLhs();
+      rewriter.replaceOp(remOp, lhs);
+      return true;
+    }
+
+    return false;
+  };
+
+  bool rewriteCompareOp(PatternRewriter &rewriter,
```
wsmoses, line 23563: this seems somewhat (but not wholly) similar to: https://github.com/EnzymeAD/Enzyme-JAX/blob/9861ed866a3c9a64c860a50583615d3846ea337f/src/enzyme_ad/jax/Passes/AffineCFG.cpp#L5941

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: the induction var is initialized as 

```cpp
    APInt minBound(bitWidth, std::min(start, limit), true);
    APInt maxBound(bitWidth, std::max(start, limit), true);
    boundsMap[inductionVar] = Bounds(minBound, maxBound);
```

wsmoses: hm, is min/max inclusive or exclusive then. If step > 0, range is [start, limit). If negative it's (limit, start]

avik-pal: inclusive in the subsequent usage, needs to be fixed here

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -23332,6 +23334,361 @@ struct DynamicPadToPad
...
+      if (auto intType = dyn_cast<IntegerType>(tensorType.getElementType())) {
+        bitWidth = intType.getWidth();
+      }
+    }
+
+    // Initialize bounds map with induction variable bounds
+    DenseMap<Value, Bounds> boundsMap;
+    if (step > 0) {
+      APInt minBound(bitWidth, start, true);
+      APInt maxBound(bitWidth, limit - 1, true);
```
wsmoses, line 23375: this isn't necessarily correct, since start == limit is a valid case [where the loop body isn't executed]

wsmoses: though I suppose if the body is never executed, what is done in the after region doesn't matter anyways

avik-pal: added a check for loop count >= 1

wsmoses: I'm not seeing where since if step > limit - start, loop count is still zero [and step only checked for sign]

avik-pal: A bit above

```
    if (!info.isValid() || !info.isConstant() ||
        info.getConstantNumIters() <= 0)
      return failure();
```

wsmoses: ah ok

wsmoses (approved):

PR #1460, opened 2025-10-07
feat: CAPI to create attributes

avik-pal:


wsmoses (approved):

src/enzyme_ad/jax/Integrations/c/EnzymeXLA.h
```diff
@@ -0,0 +1,44 @@
...
+
+MLIR_CAPI_EXPORTED MlirAttribute
+enzymexlaLapackTransposeAttrGet(MlirContext ctx, int32_t mode);
+
+MLIR_CAPI_EXPORTED MlirAttribute enzymexlaLapackSideAttrGet(MlirContext ctx,
+                                                            uint8_t left_side);
+
+MLIR_CAPI_EXPORTED MlirAttribute enzymexlaQRAlgorithmAttrGet(MlirContext ctx,
+                                                             int32_t mode);
+
```
avik-pal, line 32: @snonk you should be able to add the symm specific attributes to this file. On reactant end it will auto-generate the julia code

snonk: Will this mean I don't have to modify `src/enzyme_ad/jax/Dialect/Ops.cpp`? Or do I still have to add things there too

PR #1467, opened 2025-10-11
feat: split variadic scatter

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -7206,40 +7206,44 @@ struct SliceSimplify
...
-          out =
-              DenseIntOrFPElementsAttr::getFromRawBuffer(op.getType(), values);
-        } else {
-          auto ten = stablehlo::constantOp(inp);
-          out = fromTensor(stablehlo::sliceOp(
-              ten, stablehlo::Sizes(op.getStartIndices()),
-              stablehlo::Sizes(op.getStrides()), op.getType()));
-        }
+        // See
+        // https://github.com/EnzymeAD/Reactant.jl/pull/1740#issuecomment-3393612379
```
wsmoses, line 7210: can this be re enabled now?

wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: The commented out part is the offending part. I don't quite understand what this is doing

wsmoses: I think it's just doing const prop of the slice

avik-pal: yeah but why not just call into `stablehlo::sliceOp` (which I updated to make it do)? It seems to handle all the cases

wsmoses: Trying to special case versions where we can just use offset data directly rather than doing the per element copy in the slice app impl, like the splat optimization

wsmoses: In any case consider my approval in total for this Pr as is and we can revisit this optimization being re enabled later

PR #1482, opened 2025-10-16
feat: defer analysis passes + mark ir with analysis results

avik-pal:


src/enzyme_ad/jax/Utils.cpp
```diff
@@ -608,17 +608,27 @@ NoNanResultAnalysis::localGuaranteed(Operation *op,
...
+  if (auto boolAttr = op->getAttrOfType<BoolAttr>(getAttrName())) {
+    if (boolAttr.getValue())
+      return State::GUARANTEED;
+    else
+      return State::NOTGUARANTEED;
+  }
+
   if (auto constantOp = dyn_cast<stablehlo::ConstantOp>(op)) {
     if (guaranteed(constantOp)) {
+      op->setAttr(getAttrName(), BoolAttr::get(op->getContext(), true));
```
wsmoses: cc @ftynse 

I think we can get away with this here but it's techincally not valid since we're modifying an op outside of the pattern rewriter.

Specifically to do so properly we ought say:
```
rewriter.modifyOpInPlace(op, [&](){
  op->setAttr(...)
});
```

which then will also re add it to the worklist queue.

However since arguably the addition of the attributes here won't change any results, but essentially act as an analysis cache I think its okay.

but worth pondering

avik-pal: update to use modifyOpInPlace

wsmoses (approved):

PR #1487, opened 2025-10-18
feat: support cluster dims in kernel call op

avik-pal:
again mostly for triton where we need this if `num-ctas > 1`


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -55,6 +59,9 @@ def KernelCallOp: EnzymeXLA_Op<"kernel_call", [DeclareOpInterfaceMethods<SymbolU
     TensorI64:$blocky,
     TensorI64:$blockz,
     TensorI64:$shmem,
+    Optional<TensorI64>:$clusterx,
```
wsmoses, line 62: why have it be optional, instead just default set to 1?

avik-pal: We will have to update all of the current mlir files to include cluster(x/y/z)

PR #1496, opened 2025-10-21
[CI] Debug GB-25 workflow

giordano:


giordano:
https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/18681894618/job/53264729465#step:13:19
```
bazel-bin/libReactantExtra.so.runfiles/cuda_nvvm/nvvm/libdevice/libdevice.10.bc
```
but https://github.com/EnzymeAD/Enzyme-JAX/blob/bce12a6e157413c6849ddc1b11fe803445015549/.github/workflows/test-gb-25.yml#L133 so the file changed position, was in `bazel-bin/libReactantExtra.so.runfiles/cuda_nvcc`, it's in `bazel-bin/libReactantExtra.so.runfiles/cuda_nvvm` now.  Which means https://github.com/EnzymeAD/Reactant.jl/blob/322bd41cd94804c8cf2985f1bf9bfb7a37f49a86/deps/build_local.jl#L251-L263 is also wrong, it's generating a broken symlink.

giordano:
https://github.com/EnzymeAD/Reactant.jl/pull/1782 seems to be doing its job, with that PR libdevice is actually symlinked correctly.

PR #1500, opened 2025-10-21
Fix MacOS local builds

vimarsh6739:
Currently, MacOS builds of Enzyme-JAX fail without setting these options in bazelrc. Apple clang 17.0.0 defaults to using `-mmacosx-version-min=10.11` , which doesn't play well with `xla` which needs atleast version `10.13`. 

The `using_clang` definition is to increase bracket depth for TransformOps (we have more than 256 patterns, and Apple clang again has a max depth of 256).

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -16,15 +16,19 @@ common --action_env=PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=upb
 common --experimental_repo_remote_exec
 common --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
 common --cxxopt=-w --host_cxxopt=-w
-# common --copt=-fbracket-depth=1024 --host_copt=-fbracket-depth=1024
+
 common --define=grpc_no_ares=true
 common --define=tsl_link_protobuf=true
 common --define open_source_build=true
-common
 common --define framework_shared_object=true
 common --define tsl_protobuf_header_only=true
 common --define=allow_oversize_protos=true
 
+common --enable_platform_specific_config
+
+# macos specific config
+build:macos --macos_minimum_os=11.3
+build:macos --define using_clang=true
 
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -16,15 +16,19 @@ common --action_env=PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=upb
...
-common
 common --define framework_shared_object=true
 common --define tsl_protobuf_header_only=true
 common --define=allow_oversize_protos=true
 
+common --enable_platform_specific_config
+
+# macos specific config
+build:macos --macos_minimum_os=11.3
+build:macos --define using_clang=true
```
giordano, line 31: Technically one _could_ compile with GCC on macOS.

wsmoses: I mean frankly the correct thing to do is to replace the using clang flag we added here with a detection of the compiler type from within Bazel and select based off that

vimarsh6739: `using_clang` is simply bumping up bracket-depth here https://github.com/EnzymeAD/Enzyme-JAX/blob/f5cf0aa4b6121bdc7f930650724a606af2f12abf/src/enzyme_ad/jax/BUILD#L22

vimarsh6739: @giordano @wsmoses  What should I specify here for now then(default toolchain on mac is clang)?

giordano: > default toolchain on mac is clang

"default" != "only option possible" (although perhaps that's what XLA wants you to believe)

wsmoses: @vimarsh6739 so higher level question, what are you trying to do with this here?

macos as a build clearly works since CI says it builds.

if the goal is to make it easier to build, I would replace the `using_clang` variable with an autodetection within bazel that determins if clang is being used, and adds the template copt automatically (so no one needs to manually specify using clang).

PR #1501, opened 2025-10-21
refactor: cleanup linalg/lapack lowering

avik-pal:
~Somehow i managed to trigger a stackoverflow on reactant with this change~


avik-pal:
tested downstream, that this works

wsmoses (approved):

PR #1516, opened 2025-10-26
Statically linked cuda libs

wsmoses:
we need to local jll build test against https://github.com/JuliaPackaging/Yggdrasil/pull/12363

? not sure if the current builder stuff takes a branch


wsmoses:
The reactant Pr forcing the use of lld fixes that

PR #1532, opened 2025-10-30
WIP Correctly handle byval kernel args of gpu.launch_func

ivanradanov:
Test file `byval.cu`:

```

struct StructTy {
    int a[100];
};

__global__ void foo(StructTy a, int *out) {
    //printf("%p\n", &out[0]);
    //printf("%p\n", &a.a[0]);
    //printf("%d\n", a.a[0]);
    out[0] = 100;
    //printf("%d\n", out[0]);
}

extern "C" int printf(const char *, ...);

int main() {
#ifndef TEST_NON_CAPTURED
  const char *res;
  cudaFuncGetName(&res, foo);
  //printf("NAME %s\n", res);
#endif
  StructTy a = {{125}};
  int *d_out;
  int h_out;
  cudaMalloc(&d_out, sizeof(int));
  foo<<<1,1>>>(a, d_out);
  cudaMemcpy(&h_out, d_out, sizeof(int), cudaMemcpyDeviceToHost);
  printf("RES %d\n", h_out);
  return 0;
}
```

We need to preserve the byval attribute on both the call site and function, thus I opted to clone the llvm function and keep it as an llvm function thus keeping the byval attribute, and on the call site add a custom attr `reactant.arg_attrs` that the polygeistToLLVM pass acts on.

For some reason I am getting a launch failure with the LLVM function path (see the `#ifdef` in the source), whereas with the gpu.func path we get illegal access mostl likely due to the missing byval attribute on the kernel side function? Looking into it but for now putting up what I have

### DIFF
--- a/test/lit_tests/lowering/cuda_async_codegen.mlir
+++ b/test/lit_tests/lowering/cuda_async_codegen.mlir
@@ -31,6 +31,7 @@ module attributes {gpu.container_module} {
   }
 }
 
+
 // CHECK:  llvm.func weak_odr local_unnamed_addr @_ZN9AllocatorI8Vector_dIiEE6resizeEPS1_ii(%arg0: i64, %arg1: !llvm.ptr, %arg2: i32, %arg3: i32) -> !llvm.ptr {
 // CHECK-NEXT:    %0 = llvm.mlir.constant(0 : i64) : i64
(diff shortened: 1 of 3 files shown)


wsmoses:
so the arg_attrs attribute should be fine for both llvm and gpu func ops

PR #1538, opened 2025-10-31
fix: guaranteed analysis bfs queue was busted

avik-pal:


src/enzyme_ad/jax/Utils.h
```diff
@@ -354,7 +356,7 @@ template <typename Child> class GuaranteedResultAnalysisBase {
 
     // Map of operations we have seen before. The target of the map[o] is a list
     // of sub-queries, that if all true prove that `o` is no-nan.
-    DenseMap<Operation *, SmallPtrSet<Operation *, 2>> seen;
+    llvm::MapVector<Operation *, llvm::SmallSetVector<Operation *, 2>> seen;
```
wsmoses: looking at the code, this doesnt need to be a mapvector, since seen is not used non determinstically

src/enzyme_ad/jax/Utils.h
```diff
@@ -468,18 +474,29 @@ template <typename Child> class GuaranteedResultAnalysisBase {
     // would invalidate. Therefore all seen operations [including op] are known
     // to be guaranteed.
     for (auto &sval : seen) {
-      opCache[sval.first] = true;
-      rewriter.modifyOpInPlace(sval.first, [&]() {
-        sval.first->setAttr(attrName,
-                            BoolAttr::get(sval.first->getContext(), true));
-      });
+      bool allGuaranteed = true;
```
wsmoses: per above comment, this is the only user of seen, and these shouldnt have different results depending on input

src/enzyme_ad/jax/Utils.h
```diff
@@ -468,18 +474,29 @@ template <typename Child> class GuaranteedResultAnalysisBase {
...
+      for (auto v : sval.second) {
+        bool found = opCache.find(v) != opCache.end();
+        if ((found && !opCache[v]) || !found)
+          allGuaranteed = false;
+      }
+      if (allGuaranteed) {
+        opCache[sval.first] = true;
+        rewriter.modifyOpInPlace(sval.first, [&]() {
```
wsmoses: possible only exception here being the order in which we do the modifications in place, and perhaps add to workloist

avik-pal: I needed the MapVector for only this part else the results were non-deterministic across multiple runs.

src/enzyme_ad/jax/Utils.h
```diff
@@ -454,9 +456,13 @@ template <typename Child> class GuaranteedResultAnalysisBase {
       case State::PENDING: {
         assert(localtodo.size());
         assert(seen.find(cur) == seen.end());
-        SmallPtrSet<Operation *, 2> set(localtodo.begin(), localtodo.end());
-        for (auto v : set) {
+        llvm::SmallSetVector<Operation *, 2> set(localtodo.begin(),
```
wsmoses: so this I don't think needs to be a setvector, the change on 458 of set -> localtodo should keep fix the nondeterminsim on own?

src/enzyme_ad/jax/Utils.h
```diff
@@ -454,9 +456,13 @@ template <typename Child> class GuaranteedResultAnalysisBase {
...
+                                                 localtodo.end());
+        for (auto v : localtodo) {
           reverseSeen[v].push_back(cur);
+          if (opCache.find(v) == opCache.end() && seen.find(v) == seen.end()) {
```
wsmoses: we shouldn't need to bother with these extra checks. sure it can save some memory, but I'm not sure its terribly beneficial since immediately after popping from todo we'll do a check if we've already computed it

src/enzyme_ad/jax/Utils.h
```diff
@@ -475,11 +479,14 @@ template <typename Child> class GuaranteedResultAnalysisBase {
       });
     }
 
-    assert(opCache.find(op) != opCache.end());
-    rewriter.modifyOpInPlace(op, [&]() {
-      op->setAttr(attrName, BoolAttr::get(op->getContext(), true));
-    });
-    return true;
+    if (opCache.find(op) != opCache.end()) {
```
wsmoses: We can do slightly better here,
We can save the result of opcache.find as found or some other variable and then the result is in found->second

wsmoses (approved):

PR #1549, opened 2025-11-02
feat: symmetric tensor detection + transpose(symmetric) simplify

snonk:
Adding symmetric analysis

### DIFF
--- a/src/enzyme_ad/jax/Utils.cpp
+++ b/src/enzyme_ad/jax/Utils.cpp
@@ -611,8 +611,14 @@ SymmetricResultAnalysis::State SymmetricResultAnalysis::localGuaranteed(
   if (outTy.getDimSize(0) != outTy.getDimSize(1))
     return State::NOTGUARANTEED; // quick check and exit
 
-  if (auto constantOp = dyn_cast<stablehlo::ConstantOp>(op)) {
-    if (guaranteed(constantOp, rewriter)) {
+  SplatElementsAttr splatAttr;
+  if (matchPattern(op, m_Constant(&splatAttr))) {
+    return State::GUARANTEED;
+  }
+
+  DenseElementsAttr denseAttr;
+  if (matchPattern(op, m_Constant(&denseAttr))) {
+    if (guaranteedConstantOp(op, denseAttr, rewriter)) {
       return State::GUARANTEED;
     } else {
       return State::NOTGUARANTEED;
@@ -740,8 +746,9 @@ NoNanResultAnalysis::localGuaranteed(Operation *op,
       return State::NOTGUARANTEED;
   }
 
-  if (auto constantOp = dyn_cast<stablehlo::ConstantOp>(op)) {
-    if (guaranteed(constantOp, rewriter)) {
+  DenseElementsAttr denseAttr;
+  if (matchPattern(op, m_Constant(&denseAttr))) {
+    if (guaranteedConstantOp(op, denseAttr, rewriter)) {
       return State::GUARANTEED;
     } else {
       return State::NOTGUARANTEED;
@@ -882,8 +889,9 @@ FiniteResultAnalysis::localGuaranteed(Operation *op,
       return State::NOTGUARANTEED;
   }
 
-  if (auto constantOp = dyn_cast<stablehlo::ConstantOp>(op)) {
-    if (guaranteed(constantOp, rewriter)) {
+  DenseElementsAttr denseAttr;
+  if (matchPattern(op, m_Constant(&denseAttr))) {
+    if (guaranteedConstantOp(op, denseAttr, rewriter)) {
       return State::GUARANTEED;
     } else {
       return State::NOTGUARANTEED;
@@ -994,8 +1002,9 @@ NonNegativeResultAnalysis::State NonNegativeResultAnalysis::localGuaranteed(
       return State::NOTGUARANTEED;
   }
 
-  if (auto constantOp = dyn_cast<stablehlo::ConstantOp>(op)) {
-    if (guaranteed(constantOp, rewriter)) {
+  DenseElementsAttr denseAttr;
+  if (matchPattern(op, m_Constant(&denseAttr))) {
+    if (guaranteedConstantOp(op, denseAttr, rewriter)) {
       return State::GUARANTEED;
     } else {
       return State::NOTGUARANTEED;
--- a/src/enzyme_ad/jax/Utils.h
+++ b/src/enzyme_ad/jax/Utils.h
@@ -494,51 +494,47 @@ public:
     return false;
   }
 
-  bool guaranteed(stablehlo::ConstantOp constOp, PatternRewriter &rewriter) {
-    if (!constOp)
+  bool guaranteedConstantOp(Operation *op, DenseElementsAttr denseAttr,
+                            PatternRewriter &rewriter) {
+    if (!op)
       return false;
 
     auto attrName = ((Child *)this)->getAttrName();
-    if (auto boolAttr = constOp->getAttrOfType<mlir::BoolAttr>(attrName)) {
+    if (auto boolAttr = op->getAttrOfType<mlir::BoolAttr>(attrName)) {
       if (boolAttr.getValue())
         return true;
       else
         return false;
     }
 
-    auto it = opCache.find(constOp);
+    auto it = opCache.find(op);
     if (it != opCache.end())
       return it->second;
 
-    Attribute attr = constOp.getValue();
-
     bool guaranteedResult = false;
-    if (auto denseAttr = dyn_cast<DenseElementsAttr>(attr)) {
-      if (denseAttr.getType().getShape().size() && denseAttr.isSplat()) {
-        denseAttr = denseAttr.resizeSplat(
-            RankedTensorType::get({}, denseAttr.getType().getElementType()));
-      }
+    if (denseAttr.getType().getShape().size() && denseAttr.isSplat()) {
+      denseAttr = denseAttr.resizeSplat(
+          RankedTensorType::get({}, denseAttr.getType().getElementType()));
+    }
 
-      // For floating point values
-      if (isa<FloatType>(denseAttr.getElementType())) {
-        if (((Child *)this)->constantFloatCheck(denseAttr)) {
-          guaranteedResult = true;
-        }
-      }
-
-      // For integer values
-      if (isa<IntegerType>(denseAttr.getElementType())) {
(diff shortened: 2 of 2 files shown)


src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,130 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+  // elementwise ops
+  if (isa<stablehlo::AndOp, stablehlo::OrOp, stablehlo::XorOp, stablehlo::NotOp,
+          stablehlo::AbsOp, stablehlo::ExpOp, stablehlo::ConvertOp,
+          stablehlo::CompareOp, stablehlo::TanhOp, stablehlo::LogisticOp,
+          stablehlo::FloorOp, stablehlo::CeilOp, stablehlo::NegOp, stablehlo::LogOp,
+          stablehlo::SineOp, stablehlo::CosineOp, stablehlo::SqrtOp, stablehlo::RsqrtOp,
+          stablehlo::AddOp, stablehlo::SubtractOp, stablehlo::MulOp, stablehlo::DivOp,
+          stablehlo::MaxOp, stablehlo::MinOp>(op)) {
+    recursiveCheck = true;
+  }
```
avik-pal: you can check with `hasTrait<OpTrait::Elementwise>`

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,130 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+  if (auto constantOp = dyn_cast<stablehlo::ConstantOp>(op)) {
+    if (guaranteed(constantOp, rewriter)) {
+      return State::GUARANTEED;
+    } else {
+      return State::NOTGUARANTEED;
+    }
+  }
+
+  // A + A^T will always be symmetric
+  if (auto addOp = dyn_cast<stablehlo::AddOp>(op)) {
```
avik-pal: restrict the check to 2D tensors for now

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,130 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+
+    if (auto lhsT = dyn_cast_or_null<stablehlo::TransposeOp>(lhs.getDefiningOp())) {
+      auto lhsInput = lhsT.getOperand();
+      if (lhsInput == rhs)
+        return State::GUARANTEED;
+    }
+  }
+
+  // A * A^T will always be symmetric
+  if (auto mulOp = dyn_cast<stablehlo::MulOp>(op)) {
```
avik-pal: same here.

avik-pal: in principle any binary elementwise op with $a$, $a^T$ whould yield a symmetric tensor

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,130 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+      if (rhsInput == lhs)
+        return State::GUARANTEED;
+    }
+
+    if (auto lhsT = dyn_cast_or_null<stablehlo::TransposeOp>(lhs.getDefiningOp())) {
+      auto lhsInput = lhsT.getOperand();
+      if (lhsInput == rhs)
+        return State::GUARANTEED;
+    }
+  }
```
avik-pal: check that the permutation is indeed `[1, 0]`. this will be incorrect if we haven't simplified away a transpose of `[0, 1]` dims

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,130 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+
```
avik-pal, line 627: We also want the dot_general case where `A x A^T` is symmetric

src/enzyme_ad/jax/Utils.cpp
wsmoses: any commutative op, subtract wont

src/enzyme_ad/jax/Utils.cpp
wsmoses: we can start with 2d but I think this may be useful to generalize

I need to think about this more, but I think the general representation of symmetry here is a list of sets of dimensions, where all dimensions in the same set can be arbitrarily transposed without changing the results

specifically we can define symm of a matrix as saying x == transpose(x, [1, 0]). The generalization of this is saying

```
x == transpose(x, [something in transposable[0], something in transposable[1], ...])
```

where transposable[i] is the set of indices that can all be collectivly transposed

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,128 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+    SmallVector<int64_t> perm;
+    for (auto v : tOp.getPermutation())
+      perm.push_back(v);
+    return perm.size() == 2 && perm[0] == 1 && perm[1] == 0;
+  };
+
+  // an elementwise operation with A and A^T will always be symmetric
+  // A(A^T) will also always be symmetric
+  if (isa<stablehlo::DotGeneralOp>(op) ||
+      (op->hasTrait<OpTrait::Elementwise>() && op->hasTrait<OpTrait::IsCommutative>())) {
```
avik-pal: `op->hasTrait<OpTrait::IsCommutative>()` in addition to this mark it with HLOCommutative

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,128 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+  auto isTrueTranspose = [](stablehlo::TransposeOp tOp) -> bool {
+    SmallVector<int64_t> perm;
+    for (auto v : tOp.getPermutation())
+      perm.push_back(v);
+    return perm.size() == 2 && perm[0] == 1 && perm[1] == 0;
+  };
+
+  // an elementwise operation with A and A^T will always be symmetric
+  // A(A^T) will also always be symmetric
+  if (isa<stablehlo::DotGeneralOp>(op) ||
```
avik-pal: separate out dotgeneral

src/enzyme_ad/jax/Passes/EnzymeHLOPatterns.h
```diff
@@ -39,6 +39,8 @@ void addNoNanDivSimplify(RewritePatternSet &patterns,
 void addNoNanZeroBasePowSimplify(RewritePatternSet &patterns,
                                  bool allowOnFloatingPointMath,
                                  MLIRContext &context, PatternBenefit benefit);
+void addSymmetricSimplify(RewritePatternSet &patterns, MLIRContext &context,
+                          PatternBenefit benefit);
```
avik-pal: should not be needed

src/enzyme_ad/jax/TransformOps/TransformOps.cpp
```diff
@@ -53,6 +53,13 @@ void ApplyNoNanZeroBasePowSimplify::populatePatterns(
   addNoNanZeroBasePowSimplify(patterns, getParameter(), *getContext(),
                               PatternBenefit(getBenefit().value_or(1)));
 }
+
+void ApplySymmetricSimplify::populatePatterns(
+    RewritePatternSet &patterns) {
+  addSymmetricSimplify(patterns, *getContext(),
+                              PatternBenefit(getBenefit().value_or(1)));
+}
+
```
avik-pal: should not be needed

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,128 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+    }
+  }
+
+  // check that transpose dimensions are [1,0]
+  auto isTrueTranspose = [](stablehlo::TransposeOp tOp) -> bool {
+    SmallVector<int64_t> perm;
+    for (auto v : tOp.getPermutation())
+      perm.push_back(v);
+    return perm.size() == 2 && perm[0] == 1 && perm[1] == 0;
+  };
```
avik-pal, line 632:
```suggestion
  auto isTrueTranspose = [](stablehlo::TransposeOp tOp) -> bool {
    auto perm = tOp.getPermutation();
    return perm.size() == 2 && perm[0] == 1 && perm[1] == 0;
  };
```

check if this works

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,128 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+  // A(A^T) will also always be symmetric
+  if (isa<stablehlo::DotGeneralOp>(op) ||
+      (op->hasTrait<OpTrait::Elementwise>() && op->hasTrait<OpTrait::IsCommutative>())) {
+    auto lhs = op->getOperand(0);
+    auto rhs = op->getOperand(1);
+
+    if (auto lhsType = dyn_cast_or_null<ShapedType>(lhs.getType());
+        lhsType && lhsType.hasRank() && lhsType.getRank() == 2) {
+
+      if (auto rhsT = dyn_cast_or_null<stablehlo::TransposeOp>(rhs.getDefiningOp())) {
```
avik-pal:
```suggestion
      if (auto rhsT = rhs.getDefiningOp<stablehlo::TransposeOp>()) {
```

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -583,6 +583,128 @@ bool canApplyNoNanPattern(bool allowOnFloatingPointMath, Type outTy, Type inTy,
...
+        auto rhsInput = rhsT.getOperand();
+        if (rhsInput == lhs && isTrueTranspose(rhsT))
+          return State::GUARANTEED;
+      }
+  
+      if (auto lhsT = dyn_cast_or_null<stablehlo::TransposeOp>(lhs.getDefiningOp())) {
```
avik-pal: same here

wsmoses (approved):

PR #1552, opened 2025-11-02
refactor: move broadcast derivative rule to tablegen

avik-pal:


src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -1446,3 +1444,127 @@ def : HLODerivative<"FftOp",
...
+              Reshape
+              (InsertDeletedReduceDimsType),
+              (ReduceAdd (DiffeRet), (ReduceAddNullValue), (BroadcastDimsToReductionDims))
+            ),
+            (BroadcastDimensionsToInversePermutation)
+          )
+        )
+      ],
+      (
+        SelectIfActive $x,
```
wsmoses: select if active shouldn't be required here

however in this case I'm surprised that the hloidentityop part failed, since this is what it should do (it also supports the reverse mode tablegen so we should do that separately as well)

avik-pal: will remove

this doesn't fix the batched ad issue, i am just moving the reverse mode cpp version to tablegen

wsmoses: yeah all that readonlyidentityop does is say that for forward mode, redo the op with corresponding shadow indices [which should be equivalent to this here]

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -1446,3 +1444,158 @@ def : HLODerivative<"FftOp",
...
+              Reshape
+              (InsertDeletedReduceDimsType),
+              (ReduceAdd (DiffeRet), (ReduceAddNullValue), (BroadcastDimsToReductionDims))
+            ),
+            (BroadcastDimensionsToInversePermutation)
+          )
+        )
+      ],
+      (
+        BroadcastInDim (ResultTypeWithBatch), (Shadow $x), (getBroadcastDimensionsWithBatch)
```
wsmoses, line 1600: cam you explicitly add batch tests? In principle the tablegen interface should automatically do batching on top of so we should check it doesn't accidentally conflict

PR #1569, opened 2025-11-04
feat: optimized conv batching

avik-pal:
~doesn't yet work numerically...~


avik-pal:
Validated numerically on julia end

wsmoses (approved):

PR #1578, opened 2025-11-05
Triton forward mode AD

Pangoraw:
Many rules still missing, but setting up the infrastructure.

~~There seems to be a problem with argument attributes which are not correctly mapped when functions are cloned. This would need a fix in Enzyme proper.~~ see https://github.com/enzymead/enzyme/pull/2531

### DIFF
--- a/src/enzyme_ad/jax/Implementations/TritonDerivatives.td
+++ b/src/enzyme_ad/jax/Implementations/TritonDerivatives.td
@@ -21,7 +21,6 @@ def MakeRange : TritonInst<"MakeRangeOp">;
 def : TritonReturnOp<"ReturnOp">;
 
 def : TritonInactiveOp<"AssertOp">;
-def : TritonInactiveOp<"BitcastOp">;
 def : TritonInactiveOp<"MakeRangeOp">;
 def : TritonInactiveOp<"PrintOp">;
 


workspace.bzl
```diff
@@ -1,7 +1,7 @@
 JAX_COMMIT = "e9609ce42f272e0a4e908b9e16ea81239e76385c"
 JAX_SHA256 = ""
 
-ENZYME_COMMIT = "0ce301aedef3ca040c8703cb1b7d340ed4a58271"
```
wsmoses: @sbrantq we're blocked on enzyme bump by https://github.com/EnzymeAD/Enzyme-JAX/pull/1577 if you can take a quick look

wsmoses (approved):

src/enzyme_ad/jax/Implementations/TritonDerivatives.td
```diff
@@ -0,0 +1,42 @@
...
+class TritonReturnOp<string m> : ReturnOp<"triton", m>;
+
+def FpToFp : TritonInst<"FpToFpOp">;
+def PreciseDivF : TritonInst<"PreciseDivFOp">;
+def MakeRange : TritonInst<"MakeRangeOp">;
+
+def : TritonReturnOp<"ReturnOp">;
+
+def : TritonInactiveOp<"AssertOp">;
+def : TritonInactiveOp<"BitcastOp">;
```
wsmoses: this is not inactive, you can bitcast fp to int. Derivative here is another bitcast

wsmoses (approved):

PR #1591, opened 2025-11-07
feat: triton_ext dialect

avik-pal:


wsmoses (approved):
can we include a simple test of some sort, otherwise lgtm

PR #1595, opened 2025-11-08
Fix an assertion in guaranteed nonan

sbrantq:
In debug build the MWE gives

```
enzymexlamlir-opt: src/enzyme_ad/jax/Utils.h:457: bool mlir::enzyme::GuaranteedResultAnalysisBase<Child>::guaranteed(mlir::Operation*, mlir::PatternRewriter&) [with Child = mlir::enzyme::NoNanResultAnalysis]: Assertion `seen.find(cur) == seen.end()' failed.
```

### DIFF
--- a/src/enzyme_ad/jax/Utils.h
+++ b/src/enzyme_ad/jax/Utils.h
@@ -365,6 +365,9 @@ public:
       auto cur = todo.front();
       todo.pop_front();
 
+      if (seen.find(cur) != seen.end())
+        continue;
+
       SmallVector<Operation *, 2> localtodo;
       State status;
 
--- a/test/lit_tests/shared_operand_nonan.mlir
+++ b/test/lit_tests/shared_operand_nonan.mlir
@@ -1,0 +1,21 @@
+// RUN: enzymexlamlir-opt --pass-pipeline="builtin.module(enzyme-hlo-opt{no_nan=false})" %s | FileCheck %s
+
+module {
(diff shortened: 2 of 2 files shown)


src/enzyme_ad/jax/Utils.h
```diff
@@ -365,6 +365,9 @@ template <typename Child> class GuaranteedResultAnalysisBase {
       auto cur = todo.front();
       todo.pop_front();
 
+      if (seen.find(cur) != seen.end())
+        continue;
```
wsmoses, line 369: this isn't the right fix, this gate is required to avoid cycles/redundant work

sbrantq: Oh yes lemme ponder the whole thing for a while

src/enzyme_ad/jax/Utils.h
```diff
@@ -454,7 +454,8 @@ template <typename Child> class GuaranteedResultAnalysisBase {
       }
       case State::PENDING: {
         assert(localtodo.size());
-        assert(seen.find(cur) == seen.end());
+        if (seen.find(cur) != seen.end())
```
wsmoses: ah so the bug is somehow the if (see.find(cur) != seen.end()) at line 367 was dropped

sbrantq: You mean like this (the old fix of adding the check at line 367)?

wsmoses: yeah that was correct, I read too quickly

PR #1601, opened 2025-11-10
pass to raise triton custom call

Pangoraw:
Should triton call also have the shared memory size as operand?

### DIFF
--- a/src/enzyme_ad/jax/BUILD
+++ b/src/enzyme_ad/jax/BUILD
@@ -955,6 +955,7 @@ cc_library(
         "@llvm-project//mlir:NVVMToLLVM",
         "@llvm-project//mlir:OpenMPDialect",
         "@llvm-project//mlir:OpenMPToLLVM",
+        "@llvm-project//mlir:Parser",
         "@llvm-project//mlir:Pass",
         "@llvm-project//mlir:QuantOps",
         "@llvm-project//mlir:ROCDLDialect",
@@ -995,6 +996,7 @@ cc_library(
         "@triton//:TritonToTritonGPUPasses",
         "@xla//xla/mlir/utils:type_util",
         "@xla//xla/mlir_hlo",
+        "@zlib",
     ],
 )
 


src/enzyme_ad/jax/Passes/RaiseTritonCustomCallPass.cpp
```diff
@@ -0,0 +1,175 @@
+//===- PrintPass.cpp - Print the MLIR module                     ------------ //
```
avik-pal: nit: old comment

src/enzyme_ad/jax/Passes/RaiseTritonCustomCallPass.cpp
```diff
@@ -0,0 +1,175 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements a pass to print the MLIR module
```
avik-pal: old comment

avik-pal (commented):
> Should triton call also have the shared memory size as operand?

shared memory size is determined once we lower tt.func (see https://github.com/EnzymeAD/Enzyme-JAX/blob/b6348b25f8bf54e9ddb06bc6ad4b07fec447235a/src/enzyme_ad/jax/Passes/LowerTriton.cpp#L65)

src/enzyme_ad/jax/Passes/RaiseTritonCustomCallPass.cpp
```diff
@@ -0,0 +1,173 @@
...
+      return failure();
+
+    SymbolTable symTable(SymbolTable::getNearestSymbolTable(callOp));
+
+    OpBuilder::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPointToEnd(&symTable.getOp()->getRegion(0).front());
+
+    auto innerMod = cast<mlir::ModuleOp>(mir.release().getOperation());
+    auto outerMod = enzymexla::triton_ext::TritonModuleOp::create(
+        rewriter, callOp.getLoc(), "triton_module");
```
avik-pal, line 100: we might want to make sure there isn't a name clash when we raise multiple custom calls

Pangoraw: `symTable.insert` below will rename to a unique name like `triton_module_0` if it already exists.

PR #1602, opened 2025-11-10
feat: partial support for broadcast elementwise trait

avik-pal:
We need to support BroadcastElementwise properly but this atleast handles the extremely common cases we emit from Reactant


wsmoses (approved):
never seen the other trait before

PR #1604, opened 2025-11-11
feat: add a pass to convert tt to ttg preserving attributes

avik-pal:


src/enzyme_ad/jax/Passes/ConvertTritonToTritonGPUPreservingModuleAttributes.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (mod->hasAttr("enzymexla.ttg.num-warps")) {
+      numWarps =
+          mod->getAttrOfType<IntegerAttr>("enzymexla.ttg.num-warps").getInt();
+    }
+
+    if (mod->hasAttr("enzymexla.ttg.threads-per-warp")) {
+      threadsPerWarp =
+          mod->getAttrOfType<IntegerAttr>("enzymexla.ttg.threads-per-warp")
+              .getInt();
+    }
```
Pangoraw, line 47: nice, we should probably use the same attribute here

https://github.com/EnzymeAD/Enzyme-JAX/blob/7d0b3924ab6ec0bf8417fc2a4b0376a59b9f6207/src/enzyme_ad/jax/Passes/RaiseTritonCustomCallPass.cpp#L111-L112

i hardcoded to 64

avik-pal: This particular attribute appears after tt to ttg lowering.

Let me see if we can completely avoid specifying the grids/blocks/cluster specification to triton_call and automatically construct those from module attributes

PR #1619, opened 2025-11-16
Symm Lowering

snonk:


src/enzyme_ad/jax/Passes/LowerEnzymeXLALapack.cpp
```diff
@@ -29,6 +29,208 @@ namespace enzyme {
...
+  LogicalResult matchAndRewrite(enzymexla::SymmOp op,
+                                PatternRewriter &rewriter) const override {
+    if (backend == "cpu")
+      return matchAndRewriteCPU(op, rewriter);
+
+    // else if (backend == "cuda")
+    //   return matchAndRewriteCUDA(op, rewriter);
+
+    // else if (backend == "tpu")
+    //   return matchAndRewriteTPU(op, rewriter);
```
avik-pal: As a fallback, don't error. Emit a dot_general instead

PR #1621, opened 2025-11-16
add constant int/float check for symm

snonk:


avik-pal (approved):
run clang-format but else looks good. And add a test

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -593,11 +593,75 @@ SymmetricResultAnalysis initSymmetricResultAnalysis() {
...
+  int64_t rows = shape[0];
+  int64_t cols = shape[1];
+
+  if (rows != cols)
+    return false;
+
+  auto values = attr.getValues<APInt>();
+  auto it = values.begin();
+
+  for (int64_t i = 0; i < rows; i++) {
```
avik-pal: add a quick check here that if the attr is a SplatElementsAttr, then it is implicitly a symmetric matrix

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -593,11 +593,82 @@ SymmetricResultAnalysis initSymmetricResultAnalysis() {
...
+  return true;
 }
 
 bool SymmetricResultAnalysis::constantFloatCheck(DenseElementsAttr attr) {
-  return false; // TODO
+  if (!attr)
+    return false;
+
+  auto type = dyn_cast<RankedTensorType>(attr.getType());
+  llvm::errs() << "type: " << type.getRank() << "\n";
```
avik-pal: remove this line

PR #1626, opened 2025-11-17
add recursive symmetry checks + check for existing attr

gaurav-arya:


src/enzyme_ad/jax/Utils.cpp
```diff
@@ -605,6 +605,13 @@ SymmetricResultAnalysis::State SymmetricResultAnalysis::localGuaranteed(
     PatternRewriter &rewriter) {
   assert(op);
 
+  if (auto boolAttr = op->getAttrOfType<BoolAttr>(getAttrName())) {
+    if (boolAttr.getValue())
+      return State::GUARANTEED;
+    else
+      return State::NOTGUARANTEED;
+  }
```
avik-pal: This is probably not the right place, we want the check to be for all analysis passes. Add this https://github.com/EnzymeAD/Enzyme-JAX/blob/aded9a2821774cdfa05c6207afa3b92e16b6d583/src/enzyme_ad/jax/Utils.h#L545 and check

gaurav-arya: Makes sense. It's in all the other `localGuaranteed`s though, e.g., https://github.com/EnzymeAD/Enzyme-JAX/blob/83e2944281fed9db45a0d72ed867b79401836ef7/src/enzyme_ad/jax/Utils.cpp#L758-L763 -- I'll get rid from all and move to `localGuaranteedWithSetAttr`?

avik-pal: yeah that should be cleaner

gaurav-arya: https://github.com/EnzymeAD/Enzyme-JAX/pull/1626/commits/e3cd107b101b777e6dd4f8b754cf149575298a17

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -660,13 +660,22 @@ SymmetricResultAnalysis::State SymmetricResultAnalysis::localGuaranteed(
     }
   }
 
+  if (isa<enzymexla::SymmOp>(op)) {
+    return State::GUARANTEED;
+  }
```
avik-pal: Is this generally true? it is true for syrk but not for symm if B is not symmetric

src/enzyme_ad/jax/Utils.cpp
```diff
@@ -660,13 +660,22 @@ SymmetricResultAnalysis::State SymmetricResultAnalysis::localGuaranteed(
...
+
   bool recursiveCheck = false;
 
   // elementwise ops
   if (stablehlo::hasTraitElementwise(op)) {
     recursiveCheck = true;
   }
 
+  if (isa<stablehlo::TransposeOp>(op), isa<stablehlo::DotGeneralOp>(op)) {
```
avik-pal: for dot general it only holds if the operands commute, not in general I think

wsmoses: even then not necessarily since you could have batched dimensions

src/enzyme_ad/jax/Utils.cpp
gaurav-arya: yup you're right, there's no op for `syrk` yet right?

gaurav-arya:
Okay I just got rid of the wrong checks

src/enzyme_ad/jax/Utils.cpp
avik-pal: correct

PR #1653, opened 2025-11-24
feat: common syrk optimization patterns

### LINKED ISSUE #1489: rewrite `dot_general(A, transpose(A))` to be `syrk(A)`
For the 2d case, we should call into the respective BLAS. For the batched case, I couldn't find a strided syrk, we can emit a triton kernel for this https://github.com/KellerJordan/modded-nanogpt/blob/0e97a1cbe384f624e2a714f5e32710d12faa4f53/train_gpt.py#L139-L255

avik-pal:


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -427,7 +427,11 @@ def SyrkOp: EnzymeXLA_Op<"blas.syrk", [Pure, SameOperandsAndResultElementType]>
   let summary = "Multiplication involving a symmetric matrix";
 
   let description = [{
-    C := alpha*A*A^T + beta*C, or C := alpha*A^T*A + beta*C, where alpha and beta are scalars. C must be a n x n symmetric matrix."
+    C := alpha*A*A^T + beta*C, or C := alpha*A^T*A + beta*C, where alpha and beta are
+    scalars. C must be a n x n symmetric matrix.
+
+    If `fill` is present, then both the upper and lower triangles of the matrix are filled.
+    Otherwise only accessing `uplo` is valid.
```
wsmoses: rather we should describe this as accessing the non-uplo part is undefined behavior.

Reason being that if we want is_symmetric(syrk) == true always, then we need to leverage the fact that syrk(fill=false) could have filled the bottom with the symmetric part (and we chose the undefined behavior which best suited us). Otherwise fill = false being potentially defined means we cannot make that assumption

wsmoses (approved):

wsmoses (approved):

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -427,7 +427,11 @@ def SyrkOp: EnzymeXLA_Op<"blas.syrk", [Pure, SameOperandsAndResultElementType]>
   let summary = "Multiplication involving a symmetric matrix";
 
   let description = [{
-    C := alpha*A*A^T + beta*C, or C := alpha*A^T*A + beta*C, where alpha and beta are scalars. C must be a n x n symmetric matrix."
+    C := alpha*A*A^T + beta*C, or C := alpha*A^T*A + beta*C, where alpha and beta are
+    scalars. C must be a n x n symmetric matrix.
+
+    If `fill` is present, then both the upper and lower triangles of the matrix are filled.
+    Otherwise accessing the non-uplo part of the matrix is undefined behavior.
```
wsmoses: Very minor but can we say the values in that region are undefined. That way loading itself isn’t undefined (aka could be a segfault)

wsmoses (approved):

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
avik-pal: done

PR #1663, opened 2025-11-28
Add partial symmetry detection

gaurav-arya:
- Implement partial symmetry lattice analysis
- Port over + generalize detection logic from 2D symmetry analysis
- Add initial optimization


avik-pal (commented):
Generally LGTM. For future, I would add a small section above the partial symmetry annotation describing what the attribute means mathematically

PR #1668, opened 2025-11-30
feat: generalize loop raising passes to work with affine indexing

### LINKED ISSUE #1675: While Loop raising with multiple induction var indexing
This is relevant only post https://github.com/EnzymeAD/Enzyme-JAX/pull/1668.

- [ ] If we have multiple dims being sliced along, we can hoist the dynamic slice from the loop with is a gather op
- [ ] generalize the dynamic_update_slice hoisting as well with a scatter

avik-pal:
fixes #1675 

- loop info
  - [x] affine index propagation
  - [x] remove the loop offset map
- loop to batch
  - [x] general affine indexing support
  - [x] step != 1
  - [x] negative scaling
  - [x] iota indexing with scaled index
  - [x] constant iteration argument
- while is copy simplify
  - [x] support affine indexing
  - [x] step != 1
  - [x] negative scaling
- Miscellaneous issues
  - [x] fix regression in nested_loop_conv test
  - [x] regression in other testing
  - [x] infinite compile time
- Multiple indices depend on ind var

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -25203,7 +25203,8 @@ private:
     SmallVector<int64_t> inductionVarDimensions;
 
     for (auto [i, startIndex] : llvm::enumerate(startIndices)) {
-      if (info.isConstantAcrossIterations(startIndex))
+      // we could hoist the other dimensions but licm should fix this
+      if (info.isConstantAcrossIterations(startIndex, false))
         continue;
 
       if (!affineIndexInfo.contains(startIndex))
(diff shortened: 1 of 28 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -2237,12 +2239,13 @@ struct DUSPad final
...
+      newDusStartIndexValues.push_back(startVal - lowPad);
+    }
+
+    for (auto val : newDusStartIndexValues) {
       auto newStartConst = stablehlo::ConstantOp::create(
           rewriter, loc, indexScalarType,
-          DenseElementsAttr::get(indexScalarType, newStartAttr));
+          cast<ElementsAttr>(makeAttr(indexScalarType, val)));
       newDusStartIndices.push_back(newStartConst);
     }
```
avik-pal, line 2250: @wsmoses can you check this patch

avik-pal:
looking into them

wsmoses (approved):

PR #1673, opened 2025-12-02
feat: add outside values to while loop argument list

avik-pal:
This will likely fix the issue with tt that @giordano reported some time back.

### DIFF
--- a/src/enzyme_ad/jax/Passes/WhileLoopOutsideValuesAddToArgumentListPass.cpp
+++ b/src/enzyme_ad/jax/Passes/WhileLoopOutsideValuesAddToArgumentListPass.cpp
@@ -1,4 +1,5 @@
 #include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/RegionUtils.h"
 #include "mlir/Transforms/WalkPatternRewriteDriver.h"
 
 #include "src/enzyme_ad/jax/Passes/Passes.h"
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Passes/WhileLoopOutsideValuesAddToArgumentListPass.cpp
```diff
@@ -0,0 +1,141 @@
...
+          if (definedOutside(v, whileOp) && !seen.contains(v)) {
+            seen.insert(v);
+            extraValues.push_back(v);
+          }
+        }
+      });
+    };
+
+    collectExternal(whileOp.getCond());
+    collectExternal(whileOp.getBody());
```
Pangoraw: nit: you can probably use
```suggestion
    SetVector<Value> extraValuesSet;
    getUsedValuesDefinedAbove(whileOp->getRegions(), extraValuesSet);
    SmallVector<Value> extraValues = extraValuesSet.takeVector();

```

GleasonK:
We have:
- [WhileOpImplicitCapture](https://github.com/openxla/stablehlo/blob/85fe3af9ab56e80243299d3d9c7e3000dfa24832/stablehlo/transforms/optimization/StablehloAggressiveSimplification.cpp#L1444) which takes constant arguments and converts them to implicit captures
- [prepareExplicitCapturedConstants](https://github.com/openxla/xla/blob/80b219c0b28616e2f45be88d36d651d8b4bfc6d0/xla/mlir_hlo/stablehlo_ext/transforms/stablehlo_prepare_for_hlo_export.cpp#L117) which takes implicit captures and moves the constant values into the while body as explicit captures

An improvement to the prepareExplicitCapture for whileop would be to only sink smaller constants or single-use constants into the while body, converting larger shared constants into arguments.

PR #1676, opened 2025-12-03
Operation definitions for Bessel functions

jumerckx:
An operation is defined for each function listed here: https://specialfunctions.juliamath.org/stable/functions_overview/#Bessel-Functions

For now, all function arguments are mapped onto operation operands, as opposed to constant attributes.
~~Some of the variants are special cases of more general ones:~~
```jl
besselj(nu,z)   # Bessel function of the first kind of order nu at z

besselj0(z)     # besselj(0,z)
besselj1(z)     # besselj(1,z)
```

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -782,7 +782,7 @@ def GesvjOp : EnzymeXLA_Op<"lapack.gesvj", [Pure]> {
 
 // Special Functions - Bessel Functions
 
-def BesselJ : EnzymeXLA_Op<"special.besselj", [Pure, SameOperandsAndResultType, Elementwise]> {
+def BesselJ : EnzymeXLA_Op<"special.besselj", [Pure, AllTypesMatch<["z", "res"]>, Elementwise]> {
   let summary = "Bessel function of the first kind of order nu at z";
   
   let arguments = (ins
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -780,6 +780,276 @@ def GesvjOp : EnzymeXLA_Op<"lapack.gesvj", [Pure]> {
...
+    HLO_Tensor:$nu,
+    HLO_Tensor:$z
+  );
+  
+  let results = (outs
+    HLO_Tensor:$res
+  );
+}
+
+def BesselJ0 : EnzymeXLA_Op<"special.besselj0", [Pure, SameOperandsAndResultType, Elementwise]> {
```
wsmoses: Is there a reason for us to specialize j0 here? If we have the same implementation it lowers into anyways, no need for now

jumerckx: I imagined those might have custom approximation lowerings but I'm not sure about that. Will see if I can find something

jumerckx: for one, mpfr seems to have separate implementations for the different bessels https://github.com/JuliaMath/SpecialFunctions.jl/blob/1743a8b7ac1565213e87de418765c594720929b6/src/bessel.jl#L682C1-L709C4
I'm not sure yet what the lowering in stablehlo would look like.

wsmoses: in julia right now it just looks like they call custom c library functinos [which we can't do ourselves here anyways]

wsmoses: id probably just start with the more general one, and add the specialized one if we find a lowering for it later

jumerckx:
`SameOperandsAndResultType` I'll have to change this as some arguments can be complex.

PR #1685, opened 2025-12-05
Add raising from memref atomic rmw to affine

ivanradanov:
needs https://github.com/EnzymeAD/Enzyme/pull/2604


wsmoses (approved):

wsmoses (approved):
rebased, which now means we have the required enzyme pr in

PR #1714, opened 2025-12-09
Add detail to "Testing"

romanlee:


DEVDOCS.md
```diff
@@ -77,11 +77,11 @@ enzymexlamlir-opt --enzyme-hlo-generate-td="patterns=and_pad_pad" --transform-in
 
 ## Testing
 
-Run tests with:
+Run all tests with:
 ```bash
-bazel test //test/...
+bazel test //test:all
```
avik-pal: Isn't `...` the same as this?

romanlee: Good point. Actually `...` is recursive while `:all` does just the targets in test/BUILD. Let me fix that, probably `...` is what we want

DEVDOCS.md
```diff
@@ -77,11 +77,11 @@ enzymexlamlir-opt --enzyme-hlo-generate-td="patterns=and_pad_pad" --transform-in
 
 ## Testing
 
-Run tests with:
+Run all tests with:
 ```bash
-bazel test //test/...
+bazel test //test:...
```
wsmoses: the /... should work?

DEVDOCS.md
```diff
@@ -92,3 +92,8 @@ e.g. in `test/lit_tests/if.mlir`:
 ```
 This instructs `lit` to run the `enzyme-hlo-opt` pass on `test/lit_tests/if.mlir`.
 The output is fed to `FileCheck` which compares it against the expected result that is provided in comments in the file that start with `// CHECK`.
+
+Run all lit tests with
+```bash
+bazel test //test/lit_tests:all
```
wsmoses: is all a thing? I thought it was just bazel test //test/lit_tests/...

DEVDOCS.md
wsmoses: shouldnt it be slash?

romanlee: Sorry, yeah, that was a typo. I should never have messed with it in the first place...Fixing it now.

DEVDOCS.md
romanlee: Yeah, like I was saying above, I believe `:all` runs all the targets in the current BUILD file, `/...` does the same plus is recursive. In this case, since there are no further BUILD files in `test/lit_tests` beyond `test/lit_tests/BUILD`, they should do the same thing. Happy to change it to //test/lit_tests/... if you want though.

wsmoses: yeah we should have it also use the ...

romanlee: Changed. Sorry for the trouble on this

PR #1719, opened 2025-12-10
feat: generalize `slice_to_batch` and `concat_to_batch` passes

avik-pal:


wsmoses:
[intentionally cancelled jlls here for the jll build to start]

src/enzyme_ad/jax/Utils.h
```diff
@@ -982,6 +982,33 @@ Value copyTriangularPart(OpBuilder &builder, Value input,
...
+                             stablehlo::SliceOp otherSlice);
+
+stablehlo::ConcatenateOp lowerWrap(enzymexla::WrapOp wrap,
+                                   PatternRewriter &rewriter, bool replace);
+
+LogicalResult concatSliceSimplify(PatternRewriter &rewriter,
+                                  SmallVectorImpl<Value> &operands, int64_t dim,
+                                  SmallVectorImpl<Value> &newOperands);
+
+// these add additional checks that prevent no-op creation
```
wsmoses, line 1004: these might be useful morewidely as well.

It also might be useful to have an optional arg for sharding attr setting [if an op is created]

wsmoses (approved):
nit on some of the helpers, otherwise lgtm

PR #1721, opened 2025-12-11
fix: only raise ops with a single result (for now)

avik-pal:
this will resolve https://github.com/EnzymeAD/Reactant.jl/pull/1957#issuecomment-3639780877


wsmoses (approved):
for the future can you add a test?

PR #1722, opened 2025-12-11
Fix: slice fusion

sbrantq:
(Mentioned this in our progprog meeting earlier today.)

When implementing ThreeFry hashing I had this `(A + B) + A -> A + B` pattern where `A` and `B` are slices. If visited values are slices, it seems they are not added into `extraValues`

### DIFF
--- a/src/enzyme_ad/jax/Passes/AutoBatching.cpp
+++ b/src/enzyme_ad/jax/Passes/AutoBatching.cpp
@@ -959,7 +959,9 @@ LogicalResult GreedyWhileLoopBatchFission::matchAndRewriteImpl(
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


sbrantq:
Oops lemme ponder over the thing again

avik-pal (approved):

avik-pal:
> How do folks feel? I have a sense there may be something I’m overlooking, but I’m not seeing it at the moment.

Is the latest patch not correct? Even if we have visited, we need to still reduce with that value (a minor optimization would be to not add the value into the list for min/max).

sbrantq:
> > How do folks feel? I have a sense there may be something I’m overlooking, but I’m not seeing it at the moment.
> 
> Is the latest patch not correct? Even if we have visited, we need to still reduce with that value (a minor optimization would be to not add the value into the list for min/max).

Yeah I think it looks OK. And we have min/max check there now

PR #1744, opened 2025-12-12
ci: try fixing the benchmark parsing script

avik-pal:


giordano:
For what is worth, the message posted by the bot is now so long that it messes up my email notifications: opening the email isn't sufficient to make the notification go away, and I always need manually mark the notification as read, which is quite annoying.

Issue #1740, opened 2025-12-12
Remaining all-reduce


wsmoses:
```



[...]

PR #1753, opened 2025-12-14
feat: add high-priority for axis swap for contiguous slice

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -20085,17 +20096,40 @@ struct ConcatConcatAxisSwap final
       inners.push_back(concatOp);
     }
 
+    for (auto inner : inners) {
+      if (canFuseIntoSingleSlice(inner.getDimension(),
```
wsmoses, line 20100: we'll need to test on the loop! code from https://github.com/EnzymeAD/Enzyme-JAX/issues/1740#issuecomment-3644993212 to confirm, but I think we may need another pattern that actually breaks the existing extend to force a contiguous slice. In that case ConcatConcat wouldn't see it since its not exclusively concat of concats, it would be concat(concat(...), extend, concat(...))

wsmoses:
going to merge this now, but obviously we're likely to need the follow up mentioned

PR #1756, opened 2025-12-14
Add support to polymer for affine atomic rmw 

ivanradanov:


src/enzyme_ad/jax/polymer/mlir/lib/Conversion/Polymer/Support/IslScop.cc
```diff
@@ -1943,14 +1944,33 @@ std::unique_ptr<IslScop> IslScopBuilder::build(Operation *f) {
...
-                   memref::DeallocOp>(op)) {
-            LDBG() << "Unexpected op " << op;
-            llvm_unreachable("");
-          }
+        } else if (auto rmw = dyn_cast<enzyme::AffineAtomicRMWOp>(op)) {
+          // TODO I am not sure how isl schedule etc analysis will interact with
+          // a read and must_write to the same memory location in the same
+          // statement. Does it make sense to split the operation in two
+          // statements? However, then rescheduling may result in the two
+          // operations no longer being scheduled together which is a problem.
```
wsmoses: @ftynse

ftynse: The same way it is capable of handling a `A[i] = 2*A[i];` statement. Conceptually, you can think that all reads happen before any write, but it doesn't matter for the dependency analysis since statement is an atom for that one. If there are writes before this statement, there is both a true (read-after-write) and an output (write-after-write) dependency with this statement. If there are reads before this statement, there is an anti-dependence (write-after-read) with this statement. Similarly if there is something after it.

You don't really want to split it into two statements because then one of the statements would have _only_ reads and is trivially dead. This is why we had to come up with the statement reconstruction logic in polygeist v1. You may be able to work around it by adding a constraint in the schedule that the schedules of the two newly introduced statements must be equal and only adding a sequence node at the very end, but it sounds more pragmatic to just keep everything in one statement.

ivanradanov: I see, that makes sense, thanks.

I was worried that the a read and write in the same statement might get interpreted as a write and then read of the just written value, but indeed that wouldn't make sense.

wsmoses (approved):

PR #1760, opened 2025-12-14
feat: generalize concat reshape slice

avik-pal:


wsmoses (approved):
Can we also add a relevant test from the gb trigger ?

avik-pal:
done

wsmoses (approved):

PR #1773, opened 2025-12-15
Speed up data dependency check

wsmoses:


wsmoses (commented):
hm so tests fail for what I assume is an equivalent implementation. in that case I'll leave this to you @avik-pal

src/enzyme_ad/jax/Passes/AutoBatching.cpp
```diff
@@ -48,73 +48,71 @@ bool anyOpsAreDataDependent(ArrayRef<Operation *> ops) {
...
-    };
-
-    for (Value operand : op->getOperands()) {
-      if (earlierDefinedValues.contains(operand)) {
+    endCheck:;
+      if (!legal)
         return true;
-      }
-      enqueueDef(operand.getDefiningOp(), op);
+      return true;
```
avik-pal: @wsmoses this one is accidental right?

wsmoses: Er in what sense.

If the parent block differs, the operation is defined outside the block containing all the operations we are checking for dependencies. If that's the case an operation defined outside the block cannot have any dependency on an op inside the block so we can end early

avik-pal: there are 2 returns, so we are always exiting with a true if we enter than condition?

wsmoses: ah yeah sorry that was a bug

src/enzyme_ad/jax/Passes/AutoBatching.cpp
```diff
@@ -677,50 +670,106 @@ SliceToBatchBase::matchAndRewriteImpl(stablehlo::SliceOp sliceOp,
...
+  for (Operation *op : nonRelatedOps) {
+    if (dependsOnRelated.contains(op)) {
+      postOps.insert(op);
+    } else {
+      preOps.insert(op);
     }
+  }
+
+  // Sort each group topologically
+  auto sortedPreOps = mlir::topologicalSort(preOps);
```
wsmoses, line 751: toposort will actually move the ops iirc, in addition to returning the result

avik-pal: there seems to be another function for that https://mlir.llvm.org/doxygen/namespacemlir.html#adb95f9dc749d6a604340f83fd16b19f9 but that requires a block::iterator

wsmoses: ah sorry fair, fine as is for sure

PR #1780, opened 2025-12-16
fix: inf compile time for bcast_in_dim

avik-pal:


wsmoses (approved):
can you add a test for the future?

PR #1781, opened 2025-12-16
feat: syrk simplifications and fixes

avik-pal:
- updates the syrk op to have an output_uplo attr (replacing the old fill attr)
- simplifies output_uplo to minimize copies
- improves the cuda lowering to allow minimize layout changes

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -431,11 +431,11 @@ def SyrkOp: EnzymeXLA_Op<"blas.syrk", [Pure, SameOperandsAndResultElementType]> 
   let summary = "Multiplication involving a symmetric matrix";
 
   let description = [{
-    C := alpha*A*A^T + beta*C, or C := alpha*A^T*A + beta*C, where alpha and beta are
-    scalars. C must be a n x n symmetric matrix.
+    C_out := alpha*A*A^T + beta*C, or C_out := alpha*A^T*A + beta*C, where alpha and beta
+    are scalars. C must be a n x n symmetric matrix.
 
-    If `fill` is present, then both the upper and lower triangles of the matrix are filled.
-    Otherwise the values in the non-uplo part of the matrix are undefined.
+    `output_uplo` determines which part of `C_out` is populated. Accessing the values in
+    the non-`output_uplo` part of the matrix is undefined behavior.
   }];
 
   let arguments = (ins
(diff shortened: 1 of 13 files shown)


avik-pal:
After fixing the layouts

```
Dict{Any, Any} with 6 entries:
  "NewtonSchulz/[1024 x 1024] primal/CUDA/StructuredTensors" => 0.00164716
  "NewtonSchulz/[256 x 256] primal/CUDA/StructuredTensors"   => 0.000392565
  "NewtonSchulz/[256 x 256] primal/CUDA/Default"             => 0.000325339
  "NewtonSchulz/[4096 x 4096] primal/CUDA/StructuredTensors" => 0.0436702
  "NewtonSchulz/[4096 x 4096] primal/CUDA/Default"           => 0.0479543
  "NewtonSchulz/[1024 x 1024] primal/CUDA/Default"           => 0.00164716
```

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -431,11 +431,11 @@ def SyrkOp: EnzymeXLA_Op<"blas.syrk", [Pure, SameOperandsAndResultElementType]>
...
 
   let description = [{
-    C := alpha*A*A^T + beta*C, or C := alpha*A^T*A + beta*C, where alpha and beta are
-    scalars. C must be a n x n symmetric matrix.
+    C_out := alpha*A*A^T + beta*C, or C_out := alpha*A^T*A + beta*C, where alpha and beta
+    are scalars. C must be a n x n symmetric matrix.
 
-    If `fill` is present, then both the upper and lower triangles of the matrix are filled.
-    Otherwise the values in the non-uplo part of the matrix are undefined.
+    `output_uplo` determines which part of `C_out` is populated. Accessing the values in
```
wsmoses, line 437: I don't get the distinction between uplo and output uplo atm (the lapack call itself just has one): https://www.netlib.org/lapack/explore-html/d4/d6e/group__herk_ga09ec411a6b845c47efb0bc8bd3283b03.html

will keep reading, but probably description needs more info

avik-pal: How about this

```
  let description = [{
    C_out := alpha*A*A^T + beta*C, or C_out := alpha*A^T*A + beta*C, where alpha and beta
    are scalars. C must be a n x n symmetric matrix.

    `output_uplo` determines which part of `C_out` is populated. Accessing the values in
    the non-`output_uplo` part of the matrix is undefined behavior.

    LAPACK/BLAS routines typically require a single `uplo` attribute and it is implicitly
    assumed that the output `uplo` corresponds to the input `uplo`. This means the burden
    lies on the user to manually copy data if they need to access the other half of the
    matrix. By specifying the `output_uplo` we can perform transformations that analyze the
    entire dataflow, and avoid computing/copying half of the tensor all together. Generally,
    it is recommended to set this attribute to `enzymexla::LapackUplo::F`, and our passes
    will automatically refine this to minimize data copies.
  }];
```

wsmoses: Sgtm

wsmoses (approved):

PR #1782, opened 2025-12-16
Add pass to convert atomic rmw to non-atomic ops when legal

ivanradanov:


src/enzyme_ad/jax/Passes/RemoveAtomics.cpp
```diff
@@ -0,0 +1,461 @@
+//===---------------------------------------------------------------------===//
```
wsmoses, line 1: can we get rid of the pollydebug.cpp file?

PR #1783, opened 2025-12-16
feat: remove redundant dot_general batching dimensions

avik-pal:


avik-pal:
need to debug the python failures

PR #1785, opened 2025-12-17
Use mlir utils from Enzyme

ivanradanov:


wsmoses:
actually I'm going to be rude and reject this PR as I realize we have equally copied the same function from polygeist into Enzyme proper: https://github.com/EnzymeAD/Enzyme/blob/0fdea7861949cdc9e2866e6b6f34c55fdcca2922/enzyme/Enzyme/MLIR/Interfaces/Utils.cpp#L136

Can we update the enzyme version, then remove the version here (and have it use the single version in enzyme proper)

ivanradanov:
Moved here https://github.com/EnzymeAD/Enzyme/pull/2635 and this PR now makes us use the Enzyme utils

PR #1786, opened 2025-12-17
feat: reduce size of iteration argument

avik-pal:


avik-pal:
not needed while_dus + https://github.com/EnzymeAD/Enzyme-JAX/pull/1804 gets the job done

PR #1790, opened 2025-12-17
fix(python): properly run all benchmarks

avik-pal:
- [x] bench_vs_xla
- [x] jaxmd
- [x] llama
- [x] neuralgcm
- [x] testffi

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -542,8 +542,8 @@ def recursive_check(tester, lhs, rhs, atol=1e-8, rtol=1e-5, pname=None):
             max_err = jnp.max(jnp.abs(leaf1 - leaf2))
             if tester.skip_test_assert:
                 print(
-                    f"Skipping test assert for {pname} | {tester.name} but test"
-                    + f"failed with max error {max_err}."
+                    f"Skipping test assert for {tester.name} {pname} but test"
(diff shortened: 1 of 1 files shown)


test/bench_vs_xla.py
```diff
@@ -194,109 +197,52 @@ def f(x, y):
...
-                                < 1e-6
-                            ).all()
-                        )
-
-                        self.assertTrue(
-                            (
-                                jnp.abs(res[1] - to_backend(prevres[1], backend)) < 1e-6
-                            ).all()
-                        )
+        self.mlirad_fwd = False
```
wsmoses, line 205: tangent from current PR, but with a lot more work in place for mlir ad than we had originally, we may be able to enable a lot more of these tests

wsmoses:
somewhat of a tangent I wonder if we can use the xprof apis to get actually nice low-noise performance results

avik-pal:
exactly what I have been wondering :sweat:. the bench_vs_xla results make zero sense, those are running the exact same hlos, yet results are way off.

once this is merged, I can port some of the code I had in a reactant pr to try that.

wsmoses:
yeah in spite of the name the bench_vs_xla was originally written for correctness rather than performance testing, which is why they tend to be small

PR #1792, opened 2025-12-17
Fix: bitwise simplification

sbrantq:
A few bitwise simplification patterns are invalid when elemtype is not `i1`. Types like i64 would require the matched operand to be all ones (~not~ handled in this PR)

### DIFF
--- a/test/lit_tests/bitwise_simplify.mlir
+++ b/test/lit_tests/bitwise_simplify.mlir
@@ -36,6 +36,42 @@ module {
     %0 = stablehlo.xor %true, %arg0 : tensor<i1>
     return %0 : tensor<i1>
   }
+
+  func.func @and_i64_with_all_ones(%arg0: tensor<i64>) -> tensor<i64> {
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6919,10 +6919,13 @@ struct AndSimplify
     }
 
     // true & x -> x
-    for (int i = 0; i < 2; i++) {
-      if (matchPattern(op.getOperand(i), m_One())) {
-        rewriter.replaceOp(op, op.getOperand(1 - i));
-        return success();
+    auto elemType = cast<RankedTensorType>(op.getType()).getElementType();
+    if (elemType.isInteger(1)) {
```
wsmoses: Can this be fixed to instead be a check for all ones instead of forcing the type to be an i1

sbrantq: Yeah for sure. How's this?

PR #1804, opened 2025-12-18
feat: dynamic_slice licm + loop reduce of constant args

avik-pal:


test/lit_tests/autobatching/whileslicesizereduction.mlir
```diff
@@ -0,0 +1,70 @@
...
+// WITHOUTLICM-NEXT:   %cst_1 = stablehlo.constant dense<0.000000e+00> : tensor<12x32xf32>
+// WITHOUTLICM-NEXT:   %c_2 = stablehlo.constant dense<1> : tensor<i32>
+// WITHOUTLICM-NEXT:   %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<12x32xf32>) -> tensor<32x12xf32>
+// WITHOUTLICM-NEXT:   %1 = stablehlo.convert %arg1 : (tensor<i64>) -> tensor<i32>
+// WITHOUTLICM-NEXT:   %2 = stablehlo.subtract %1, %c_2 : tensor<i32>
+// WITHOUTLICM-NEXT:   %3 = stablehlo.dynamic_slice %0, %2, %c, sizes = [1, 6] : (tensor<32x12xf32>, tensor<i32>, tensor<i32>) -> tensor<1x6xf32>
+// WITHOUTLICM-NEXT:   %4 = stablehlo.multiply %3, %3 : tensor<1x6xf32>
+// WITHOUTLICM-NEXT:   %5 = stablehlo.reshape %4 : (tensor<1x6xf32>) -> tensor<6x1xf32>
+// WITHOUTLICM-NEXT:   %6 = stablehlo.broadcast_in_dim %5, dims = [1, 2] : (tensor<6x1xf32>) -> tensor<6x6x1xf32>
+// WITHOUTLICM-NEXT:   %7 = stablehlo.reduce(%6 init: %cst) applies stablehlo.add across dimensions = [0] : (tensor<6x6x1xf32>, tensor<f32>) -> tensor<6x1xf32>
```
wsmoses: at first glance I don't see much of a difference between the two cases, you can potenetially have overlapping checks for both with --check-prefix=WITHOUTLICM,EITHER. and use EITHER-NEXT for both lines where they match and WITHOUTLICM / WITHLICM where not

avik-pal: it was indeed testing that both give the same result. did not realize I could just use the same prefix. fixed

wsmoses: If they both give the same result you can just use check and not a custom prefix

wsmoses (approved):

test/lit_tests/autobatching/whileslicesizereduction.mlir
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+// CHECK: func.func @main(%arg0: tensor<12x32xf32> {enzymexla.memory_effects = []}, %arg1: tensor<i64> {enzymexla.memory_effects = []}) -> tensor<12x32xf32> attributes {enzymexla.memory_effects = []} {
+// CHECK-NEXT:   %cst = stablehlo.constant dense<0.000000e+00> : tensor<f32>
+// CHECK-NEXT:   %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<6x1xf32>
+// CHECK-NEXT:   %c = stablehlo.constant dense<2> : tensor<i32>
+// CHECK-NEXT:   %cst_1 = stablehlo.constant dense<0.000000e+00> : tensor<12x32xf32>
+// CHECK-NEXT:   %c_2 = stablehlo.constant dense<1> : tensor<i32>
+// CHECK-NEXT:   %0 = stablehlo.transpose %arg0, dims = [1, 0] : (tensor<12x32xf32>) -> tensor<32x12xf32>
```
wsmoses, line 40: as a tangent for the future, the dynamic slice from here means that this transpose itself isn't necessary

PR #1806, opened 2025-12-18
Try new slice dus optimizations

wsmoses:
cc @rosiezou


wsmoses:
cc @giordano could we fix the CI to exclude the regular build tests in this case? [eg where only updating the gb25 yml]

wsmoses:
why are they showing up in this PR then?

wsmoses:
ah got it

PR #1809, opened 2025-12-19
fix: dot_general batch dims removal pass

avik-pal:


avik-pal:
failure only on the arm runner, fun...

wsmoses (approved):

PR #1828, opened 2025-12-23
debug gpu

wsmoses:
Removed unsupported OS configurations from the build matrix and added environment variables for debugging during tests.


avik-pal:
we probably need cuda13 https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/20447693401/job/58754335737?pr=1828#step:3:12

avik-pal:
yeah

PR #1829, opened 2025-12-23
MPI Ops and lowering passes

romanlee:
Adds EnzymeXLA Ops and lowering to LLVM passes for the following MPI commands:

* MPI_Comm_rank

* MPI_Comm_size
* MPI_Barrier

* MPI_Send
* MPI_Recv
* MPI_Isend
* MPI_Irecv
* MPI_Wait
* MPI_Allreduce

Note: Currently MPI_COMM_WORLD is the only communicator supported. However, I did come up with a solution to handle the Datatype (and Op) for Isend/Irecv/Send/Recv (Allreduce), which we had previously had some discussions about. See comment below.

### DIFF
--- a/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.h
+++ b/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.h
@@ -48,9 +48,9 @@ enzymexlaGeluApproximationAttrGet(MlirContext ctx, int32_t mode);
 //===----------------------------------------------------------------------===//
 
 MLIR_CAPI_EXPORTED MlirAttribute enzymexlaMPIDatatypeAttrGet(MlirContext ctx,
-                                                            int32_t mode);
+                                                             int32_t mode);
 
-MLIR_CAPI_EXPORTED MlirAttribute enzymexlaMPIOpAttrGet(MlirContext ctx, 
+MLIR_CAPI_EXPORTED MlirAttribute enzymexlaMPIOpAttrGet(MlirContext ctx,
                                                        int32_t mode);
 
 //===----------------------------------------------------------------------===//
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
@@ -1,8 +1,8 @@
 // #include "mhlo/IR/hlo_ops.h"
 #include "mlir/Conversion/LLVMCommon/TypeConverter.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "src/enzyme_ad/jax/Dialect/Dialect.h"
 #include "src/enzyme_ad/jax/Dialect/Ops.h"
 #include "src/enzyme_ad/jax/Passes/LinalgUtils.h"
@@ -128,8 +128,8 @@ struct MPICommRankOpLowering
 
       // Create a constant tensor to hold the result
       auto tensorType = llvm::cast<RankedTensorType>(op->getResultTypes()[0]);
-      auto constantAttr = DenseIntElementsAttr::get(tensorType,
-          ArrayRef<int32_t>{-1});
+      auto constantAttr =
+          DenseIntElementsAttr::get(tensorType, ArrayRef<int32_t>{-1});
       Value constantTensor = rewriter.create<stablehlo::ConstantOp>(
           op.getLoc(), tensorType, constantAttr);
 
@@ -263,8 +263,8 @@ struct MPICommSizeOpLowering
 
       // Create a constant tensor to hold the result
       auto tensorType = llvm::cast<RankedTensorType>(op->getResultTypes()[0]);
-      auto constantAttr = DenseIntElementsAttr::get(tensorType,
-          ArrayRef<int32_t>{-1});
+      auto constantAttr =
+          DenseIntElementsAttr::get(tensorType, ArrayRef<int32_t>{-1});
       Value constantTensor = rewriter.create<stablehlo::ConstantOp>(
           op.getLoc(), tensorType, constantAttr);
 
@@ -912,8 +912,8 @@ struct MPIIsendOpLowering : public OpRewritePattern<enzymexla::MPIIsendOp> {
       // Create a constant tensor to hold request
       auto i64Type = rewriter.getI64Type();
       auto tensorType = RankedTensorType::get({}, i64Type);
-      auto constantAttr = DenseIntElementsAttr::get(tensorType,
-          ArrayRef<int64_t>{-1});
+      auto constantAttr =
+          DenseIntElementsAttr::get(tensorType, ArrayRef<int64_t>{-1});
       Value constantTensor = rewriter.create<stablehlo::ConstantOp>(
           op.getLoc(), tensorType, constantAttr);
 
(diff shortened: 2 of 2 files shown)


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1058,4 +1058,141 @@ def AffineStoreVar : EnzymeXLA_Op<"store_var", [Pure]> {
...
+def MPIAllreduceOp : EnzymeXLA_Op<"allreduce", []> {
+  let summary = "Equivalent to "
+                "`MPI_Allreduce(&sendbuf, &recvbuf, count, datatype, op, MPI_COMM_WORLD)`";
+
+  let arguments = (
+    ins AnyTensor : $sendbuf,
+    AnyTensor : $inbuf,
+    TensorOf<[I32]> : $count,
+    StrAttr:$datatype,
+    StrAttr:$op
```
romanlee: Thoughts on this solution to handle Datatype (and Op) for Isend/Irecv/Send/Recv (Allreduce)? We still register the symbol and get it's name on the Reactant side like before, then we just pass the symbol name in to the Op via an StrAttr. This gives us all the support for the various Datatypes x Ops combos that we had before.

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1058,4 +1058,141 @@ def AffineStoreVar : EnzymeXLA_Op<"store_var", [Pure]> {
...
+def MPISendOp : EnzymeXLA_Op<"send", []> {
+  let summary = "Equivalent to "
+                "`MPI_Send(&buf, count, datatype, dest, tag, comm)`";
+
+  let arguments = (
+    ins AnyTensor : $buf,
+    TensorOf<[I32]> : $count,
+    TensorOf<[I32]> : $dest,
+    TensorOf<[I32]> : $tag,
+    StrAttr:$datatype
```
wsmoses: cc @ftynse We can have this be a TypeAttr [and use the elementtype]

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1058,4 +1058,141 @@ def AffineStoreVar : EnzymeXLA_Op<"store_var", [Pure]> {
...
+    TensorOf<[I32]> : $count,
+    TensorOf<[I32]> : $source,
+    TensorOf<[I32]> : $tag,
+    TensorOf<[I64]> : $inrequest,
+    StrAttr:$datatype
+  );
+
+  let results = (
+    outs AnyTensor : $outbuf,
+    TensorOf<[I64]> : $outrequest
```
romanlee: I'm unsure about this `inbuf`/`outbuf` `inrequest`/`outrequest` design here and in other Ops. It allows us to reproduce exactly the same IR that we had before (ie, via manual lowering from Reactant). But the more I've thought about this the more it seems like a weird design for certain Ops/arguments. Eg, here it maybe makes sense for the buffers, but not the requests? And for `Comm_rank`, it seems like it would make more sense for it to be a pure function, where we don't take in a rank, only output one. However, doing that would require some changes to the way we call this on the Reactant side.

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
wsmoses: commented on datatype above, for op we can create an enum and put add/max/min/etc in it [and do the corresponding lowering]

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: We could do it that way too. Although that would mean duplicating a lot of logic that we currently rely on MPI.jl for. Would the ementtype approach be better?

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1058,4 +1058,141 @@ def AffineStoreVar : EnzymeXLA_Op<"store_var", [Pure]> {
   let summary = "Fake store an SSA value for conversion to ISL";
 }
 
+// MPI Ops
+
+def MPICommRankOp : EnzymeXLA_Op<"comm_rank", []> {
+  let summary = "Equivalent to " "`MPI_Comm_rank(MPI_COMM_WORLD, &rank)`";
+
+  let arguments = (
+    ins AnyTensor : $inrank
```
wsmoses: yeah this shouldn't require an inrank, it can just return the rank, same with size

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
wsmoses: this should only return a request, not take as an input [since its write only]

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1058,4 +1058,141 @@ def AffineStoreVar : EnzymeXLA_Op<"store_var", [Pure]> {
   let summary = "Fake store an SSA value for conversion to ISL";
 }
 
+// MPI Ops
+
+def MPICommRankOp : EnzymeXLA_Op<"comm_rank", []> {
```
avik-pal:
```suggestion
def MPICommRankOp : EnzymeXLA_Op<"mpi.comm_rank", []> {
```

we should mimic the lapack/blas convention here

romanlee: Will do, thanks

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: Ok, I removed the inrank arg from comm rank and updated the lowering pass. In the new lowering pass, I now create a constant tensor to hold the result which I then pass into the wrapper function, which allows me to still use output operand aliases to get the result back. 

https://github.com/EnzymeAD/Enzyme-JAX/blob/9358d053e98f4b1607dc9b998cc7362e70437dbd/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp#L129-L134

Is this an ok approach? If so, I'll go ahead and change all the other ops similarly

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: Could you expand on what you mean here? What do you want the op arg to be instead?

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: Could you expand on what you mean here as well?

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
wsmoses: something like this https://github.com/EnzymeAD/Enzyme-JAX/blob/a83eae480f2c9450ad774e9432fad222c33a5411/src/enzyme_ad/jax/Dialect/EnzymeXLAAttrs.td#L55

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: Ok, I removed inrank, insize, inrequest from comm_rank, comm_size, isend, irecv. 

Now, I still have `recv`, `irecv` and `allreduce` taking in an `inbuf` and outputting an `outbuf`. Is this design ok, or should I similarly remove the `inbufs` from those too?

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: Ok, this should be good

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
romanlee: Ok, this should be good

PR #1853, opened 2025-12-26
fix: permutation of DUS start indices

avik-pal:
source of the OOB https://github.com/EnzymeAD/Reactant.jl/issues/2024


wsmoses:
mildly concerning we didn't see this earlier, but glad we caught

Issue #1865, opened 2025-12-28
break up `reduce(add, add(mul(), mul()))`


avik-pal:
Enables fusing the ops into a `op(dot_general, dot_general)`.

```mlir
module @reactant_syr2k attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<f32> {enzymexla.memory_effects = []}, %arg1: tensor<f32> {enzymexla.memory_effects = []}, %arg2: tensor<64x64xf32> {enzymexla.memory_effects = []}, %arg3: tensor<64x64xf32> {enzymexla.memory_effects = []}, %arg4: tensor<64x64xf32> {enzymexla.memory_effects = [], tf.aliasing_output = 0 : i32}) -> tensor<64x64xf32> attributes {enzymexla.memory_effects = []} {
    %cst = stablehlo.constant dense<0.000000e+00> : tensor<f32>
    %0 = stablehlo.reshape %arg3 : (tensor<64x64xf32>) -> tensor<64x64x1x1xf32>
    %1 = stablehlo.broadcast_in_dim %arg0, dims = [] : (tensor<f32>) -> tensor<64x64x1x1xf32>
    %2 = stablehlo.multiply %0, %1 : tensor<64x64x1x1xf32>
    %3 = stablehlo.broadcast_in_dim %arg0, dims = [] : (tensor<f32>) -> tensor<64x64xf32>
    %4 = stablehlo.multiply %arg2, %3 : tensor<64x64xf32>
    %5 = stablehlo.broadcast_in_dim %arg2, dims = [1, 0] : (tensor<64x64xf32>) -> tensor<64x64x64x1x1xf32>
    %6 = stablehlo.broadcast_in_dim %2, dims = [1, 2, 3, 4] : (tensor<64x64x1x1xf32>) -> tensor<64x64x64x1x1xf32>
    %7 = stablehlo.multiply %6, %5 : tensor<64x64x64x1x1xf32>
    %8 = stablehlo.broadcast_in_dim %4, dims = [1, 2] : (tensor<64x64xf32>) -> tensor<64x64x64x1x1xf32>
    %9 = stablehlo.broadcast_in_dim %arg3, dims = [1, 0] : (tensor<64x64xf32>) -> tensor<64x64x64x1x1xf32>
    %10 = stablehlo.multiply %8, %9 : tensor<64x64x64x1x1xf32>
[...]

PR #1870, opened 2025-12-29
feat: remove more intermediate reshape operations

avik-pal:
Mostly experimenting.

- Add reduce to fusible with reshape (delete dims case)
- create helper functions for the fusions so that we can process those in one go, else phase ordering creates issues
- instead of using the worklist to process in isfusible, we keep that only for immediate cases
- for reshape_elementwise, we process a full chain of elementwise and immediately eliminate all of the reshapes

fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/1862 #1865

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -24090,6 +24090,10 @@ LogicalResult dotGeneralDistributiveSimplify(OpTy op,
   if (rhsDotGeneralOp && !isOnlyUsedInOperation(rhsDotGeneralOp, op))
     return failure();
 
+  if (lhsDotGeneralOp == rhsDotGeneralOp) {
+    return failure(); // only a single op
+  }
+
   if (lhsDotGeneralOp && rhsDotGeneralOp)
     return dotGeneralDistributiveSimplify<OpTy>(op, lhsDotGeneralOp,
                                                 rhsDotGeneralOp, rewriter);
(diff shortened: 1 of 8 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -11195,35 +11265,85 @@ struct ReshapeElementwise final
...
 
-    SmallVector<Value> ops;
-    for (auto v : elem->getOperands()) {
-      // doing this prevents a potential infinite loop of patterns (and
-      // generally produces nicer intermediate IR)
-      DenseElementsAttr cstAttr;
+    // Topological sort: we need to process ops in dependency order
+    // Since we collected via DFS, we need to reverse and ensure dependencies
+    // come first
+    auto sortedOps = mlir::topologicalSort(chainOps);
```
wsmoses: the toposort here still worries me a bit here, in contrast to a linear scan through the block [at least on large blocks]

avik-pal: let me ponder a bit, we might be able to get away without sorting if we ensure a proper insertion order into the chain

avik-pal: should now avoid the topological sort

avik-pal: @wsmoses is this good to go?

PR #1874, opened 2025-12-31
feat: raise stencils to reduce_window

avik-pal:


wsmoses (commented):
this seems to share a lot of functionality with https://github.com/EnzymeAD/Enzyme-JAX/blob/426a717b0f6c246c205f72ebeaf67727b028acf7/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L18867, is it possible to share some code as well [or perhaps extend the latter to support dynamic slice?]

PR #1875, opened 2026-01-04
add rocdl target attribute support and calling AMDGPUSerializer

xys-syx:
https://github.com/EnzymeAD/Enzyme-JAX/blob/a6af9b3e7ad5d56fde968895e2e420ef085062a2/test/lit_tests/lowering/gpu-recognize2.mlir#L61
@wsmoses @vimarsh6739 
Hello, I have several questions that block the progress:
1. Could you tell where I can find an example for AMD GPU's corresponding to this?
  I only can find something like this:
  https://github.com/llvm/llvm-project/blob/c97de4387b076176d3dbd428229a03b8909941af/clang/test/Driver/amdgpu-mcpu.cl
  https://github.com/llvm/llvm-project/blob/c97de4387b076176d3dbd428229a03b8909941af/clang/test/CodeGenOpenCL/amdgpu-features.cl
  They are not exactly match and I cannot find an example with abiVersion according to [https://github.com/llvm/llvm-project/blob/c97de4387b076176d3dbd428229a03b8909941af/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td#L2147](here)

2. And, according to this: https://github.com/llvm/llvm-project/blob/ccb47d0fb9d01d44764fa4ca5c6dcf239ab76ed2/mlir/include/mlir/Dialect/LLVMIR/NVVMOps.td#L6184
The default parameter for NVVMTargetAttr is not the same comparing to llvm-project, should I modify or not?


xys-syx:
Or because I set as draft, so you cannot see the @ ?

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
```diff
@@ -2510,6 +2511,14 @@ gdgo->erase();
               continue;
             if (s0.getValue() == "target-cpu")
               sm = s1.getValue();
+            if (backend == "rocm") {
+              if (sm.find("sm_") != std::string::npos) {
+                llvm::errs() << "Warning: Found NVIDIA architecture (" << sm
+                             << ") while targeting ROCm. Overriding to default "
+                                "AMD chip.\n";
+                sm = "gfx900";
```
ivanradanov: Maybe just abort(), as this could lead to some hard to figure out bugs

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
```diff
@@ -2526,15 +2535,25 @@ gdgo->erase();
...
+                /*abiVersion=*/"600");
+          } else {
+            auto chip = sm;
+            if (chip.size() == 0)
+              chip = "sm_80";
+            auto features = feat;
+            if (features.size() == 0)
+              features = "+ptx73";
+            target = NVVM::NVVMTargetAttr::get(
+                gmod.getContext(), /*optLevel*/ 2,
```
ivanradanov: I think the default opt level for ptxas is O3, I would go for that. Maybe we should also make the amd one O3 to match.

src/enzyme_ad/jax/Passes/GPULaunchRecognition.cpp
```diff
@@ -54,24 +55,37 @@ struct GPULaunchRecognitionPass
...
-        /*optLevel=*/2, /*triple=*/"nvptx64-nvidia-cuda", chip, features,
-        /*flags=*/nullptr,
-        /*linkLibs=*/nullptr);
+    Attribute target;
+    if (backend == "rocm") {
+      // Here temporarily set as "" for ROCm backend
+      auto chip = "";
+      auto features = "+wavefrontsize64";
+
+      target = moduleBuilder.getAttr<ROCDL::ROCDLTargetAttr>(
```
ivanradanov, line 71: Opt levels 3 here as well?

PR #1886, opened 2026-01-05
Improve caching

wsmoses:


giordano (approved):
That's ok-ish, there will be conflicts between different jobs, but that's unavoidable.  The real problem is that GitHub cache is almost unusable now for this project, with aggressive eviction as soon as we get over the threshold, we should consider other caching options altogether (what are the chances of getting GCP working?)

PR #1899, opened 2026-01-07
feat: missing licm and cse passes

avik-pal:


wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -29303,8 +29315,9 @@ struct EnzymeHLOOptPass
                    CSE<stablehlo::ConcatenateOp>, CSE<stablehlo::MaxOp>,
                    CSE<stablehlo::NegOp>, CSE<stablehlo::AbsOp>,
                    CSE<enzymexla::RotateOp>, CSE<enzymexla::WrapOp>,
-                   CSE<enzymexla::ExtendOp>, CSEIota, CSE<stablehlo::GatherOp>,
-                   CSE<stablehlo::ScatterOp>>(context, PatternBenefit(65000));
+                   CSE<enzymexla::ExtendOp>, CSEIota, CSE<stablehlo::CompareOp>,
```
wsmoses, line 29324: @avik-pal to double check, does our default CSE also check the corresponding attribute for the comparison direction?

Also does it check the region for scatter/gather?

avik-pal: we check via `!OperationEquivalence::isEquivalentTo(op, nop, OperationEquivalence::IgnoreLocations)`. This should be enough right?

wsmoses: I think so, cc @ftynse 

but maybe add a quick test just in case, for something it shouldn't cse [and also something it should]

avik-pal: confirmed that it works

wsmoses:
also 10s but related, can you add licm_iota ?

PR #1901, opened 2026-01-07
[CI] Further improve gb25 caching

wsmoses:


.github/workflows/test-gb-25.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         run: |
           # Artifact names cannot include forward slashes and some other characters, here we
           # do some sanitisation.
-          ARTIFACT_INDEX=${{ matrix.julia_version }}-${{ matrix.xla_commit }}-${{ matrix.gb25_commit }}-${{ matrix.reactant_commit }}
+          ARTIFACT_INDEX=${{ matrix.julia_version }}
```
giordano, line 68: This doesn't have anything to do with caching, and the artifact names _must_ be fully disambiguated.

wsmoses: doesn't this set the julia env, and in particular the reason why it improves ci caching is that downloading cuda_runtime_jll takes like 30 mins on ci [and we are not hitting the previous cache, last I checked]. Or is this independent from that?

giordano: This is an env var, but it's used only for the names of the artifacts uploaded during the build.  The caching system doesn't use this variable at all.

PR #1902, opened 2026-01-07
feat: fold constants above certain threshold

avik-pal:


wsmoses (approved):
Maybe rename it RecognizeFromConstant?

wsmoses (approved):
something for another PR, but also broadcasted constant would be useful too

avik-pal:
we probably need a faster check here https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/20791545056/job/59714656562?pr=1902 (or a way to bail out quickly)

PR #1910, opened 2026-01-08
Implementation of periodic checkpointing

sriharikrishna:
Added an attribute `enzymexla.checkpoint_period` to specify a checkpoint period. I am not sure how this attribute will be set from the user's code.

Here is how I test the code:
```
bazel-bin/enzymexlamlir-opt test/lit_tests/diffrules/stablehlo/while_periodic_checkpointing.mlir --enzyme-wrap="infn=with_periodic_checkpointing retTys=enzyme_dup argTys=enzyme_dup mode=ReverseModeCombined" --canonicalize --remove-unnecessary-enzyme-ops --enzyme-simplify-math
```

```
bazel-bin/enzymexlamlir-opt test/lit_tests/diffrules/stablehlo/while_sqrt_vs_periodic_checkpointing.mlir --enzyme --canonicalize --remove-unnecessary-enzyme-ops --enzyme-simplify-math --arith-raise --canonicalize | bazel-out/k8-opt/bin/external/stablehlo/stablehlo-translate --interpret
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -224,7 +224,7 @@ jobs:
           
       - name: Update Nightly Release
         uses: ncipollo/release-action@v1
-        if: ${{ !cancelled() && github.ref == 'refs/heads/main' }} 
+        if: ${{ !cancelled() && github.ref == 'refs/heads/main' && matrix.os == 'linux-x86-n2-32' }} 
         with:
           tag: nightly
           name: Nightly Release
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
@@ -915,7 +915,7 @@ struct MPIIsendOpLowering : public OpRewritePattern<enzymexla::MPIIsendOp> {
       // Create a constant tensor to hold request
       auto tensorType = RankedTensorType::get({}, i32Type);
       auto constantAttr =
-          DenseIntElementsAttr::get(tensorType, ArrayRef<int64_t>{-1});
+          DenseIntElementsAttr::get(tensorType, ArrayRef<int32_t>{-1});
       Value constantTensor = rewriter.create<stablehlo::ConstantOp>(
           op.getLoc(), tensorType, constantAttr);
 
@@ -1101,7 +1101,7 @@ struct MPIIrecvOpLowering : public OpRewritePattern<enzymexla::MPIIrecvOp> {
       // Create a constant tensor to hold request
       auto tensorType = RankedTensorType::get({}, i32Type);
       auto constantAttr =
-          DenseIntElementsAttr::get(tensorType, ArrayRef<int64_t>{-1});
+          DenseIntElementsAttr::get(tensorType, ArrayRef<int32_t>{-1});
       Value constantTensor = rewriter.create<stablehlo::ConstantOp>(
           op.getLoc(), tensorType, constantAttr);
 


test/lit_tests/diffrules/stablehlo/while_periodic_checkpointing.mlir
```diff
@@ -0,0 +1,83 @@
...
+    // Both should produce the same results
+    check.expect_almost_eq %diffe_periodic#0, %diffe_no_checkpointing#0 : tensor<f64>
+    check.expect_almost_eq %diffe_periodic#1, %diffe_no_checkpointing#1 : tensor<f64>
+
+    return
+  }
+}
+
+// CHECK-LABEL: func.func private @diffewith_periodic_checkpointing
+// CHECK: stablehlo.while
```
wsmoses: can you add more lines of check to confirm it is as expected?

wsmoses:
user code side we need to modify https://github.com/EnzymeAD/Reactant.jl/blob/6a952c1b5ffd49b0860e0125030568e4252e3e1e/src/Ops.jl#L2149 and https://github.com/EnzymeAD/Reactant.jl/blob/6a952c1b5ffd49b0860e0125030568e4252e3e1e/src/ControlFlow.jl#L17 and https://github.com/EnzymeAD/Reactant.jl/blob/6a952c1b5ffd49b0860e0125030568e4252e3e1e/lib/ReactantCore/src/ReactantCore.jl#L364

adding that, you should be able to test julia side, and even play with fun profiling utilities:

https://enzymead.github.io/Reactant.jl/dev/tutorials/profiling

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -951,102 +1008,140 @@ class AutoDiffWhileRev
...
-                << nInner << " nOuter=" << nOuter
-                << " iters=" << info.getConstantNumIters() << "\n";
+          int64_t numIters = info.getConstantNumIters();
+          int64_t nInner, nOuter;
+
+          if (revModeInfo.mode == CONSTANT_CHECKPOINTING) {
+            // sqrt scheme
+            nInner = std::sqrt(numIters);
+            nOuter = nInner;
+            if (nInner * nOuter != numIters) {
```
Pangoraw: can we use the same `limit = min(nInner, numIters - outerIV)` if this is false. This error is more of a TODO since this was not yet supported (it should be supported for scf.for in Enzyme proper)

sriharikrishna:
> from what I understand, we can make constant checkpointing just a case of periodic where the inner is `sqrt(numIters)`?

Yes. This is correct. One option is to use `sqrt(numIters)` as the default value when the user does not specify a checkpointing period.

wsmoses:
> > from what I understand, we can make constant checkpointing just a case of periodic where the inner is `sqrt(numIters)`?
> 
> Yes. This is correct. One option is to use `sqrt(numIters)` as the default value when the user does not specify a checkpointing period.

I think this would be wise/helpful

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
sriharikrishna: Yes, we can do that to generalize both cases.

wsmoses:
@sriharikrishna first thing, the current test fails:

```

Command Output (stderr):
--
enzymexlamlir-opt /root/.bazel/sandbox/processwrapper-sandbox/6734/execroot/__main__/bazel-out/k8-opt/bin/test/lit_tests/diffrules_stablehlo_while_checkpointing.mlir.test.runfiles/__main__/test/lit_tests/diffrules/stablehlo/while_checkpointing.mlir --enzyme --canonicalize --remove-unnecessary-enzyme-ops --enzyme-simplify-math --arith-raise --canonicalize | FileCheck /root/.bazel/sandbox/processwrapper-sandbox/6734/execroot/__main__/bazel-out/k8-opt/bin/test/lit_tests/diffrules_stablehlo_while_checkpointing.mlir.test.runfiles/__main__/test/lit_tests/diffrules/stablehlo/while_checkpointing.mlir # RUN: at line 1
+ enzymexlamlir-opt /root/.bazel/sandbox/processwrapper-sandbox/6734/execroot/__main__/bazel-out/k8-opt/bin/test/lit_tests/diffrules_stablehlo_while_checkpointing.mlir.test.runfiles/__main__/test/lit_tests/diffrules/stablehlo/while_checkpointing.mlir --enzyme --canonicalize --remove-unnecessary-enzyme-ops --enzyme-simplify-math --arith-raise --canonicalize
+ FileCheck /root/.bazel/sandbox/processwrapper-sandbox/6734/execroot/__main__/bazel-out/k8-opt/bin/test/lit_tests/diffrules_stablehlo_while_checkpointing.mlir.test.runfiles/__main__/test/lit_tests/diffrules/stablehlo/while_checkpointing.mlir
/root/.bazel/sandbox/processwrapper-sandbox/6734/execroot/__main__/bazel-out/k8-opt/bin/test/lit_tests/diffrules_stablehlo_while_checkpointing.mlir.test.runfiles/__main__/test/lit_tests/diffrules/stablehlo/while_checkpointing.mlir:120:16: error: CHECK-NEXT: expected string not found in input
// CHECK-NEXT: %10 = stablehlo.subtract %c, %iterArg_8 : tensor<i64>
               ^
<stdin>:94:8: note: scanning from here
 } do {
       ^
<stdin>:96:2: note: possible intended match here
 %11 = stablehlo.subtract %10, %iterArg_8 : tensor<i64>
 ^

Input file: <stdin>
Check file: /root/.bazel/sandbox/processwrapper-sandbox/6734/execroot/__main__/bazel-out/k8-opt/bin/test/lit_tests/diffrules_stablehlo_while_checkpointing.mlir.test.runfiles/__main__/test/lit_tests/diffrules/stablehlo/while_checkpointing.mlir

-dump-input=help explains the following input dump.

Input was:
<<<<<<
            .
            .
            .
           89:  %5 = stablehlo.multiply %c_0, %4 : tensor<i64> 
           90:  %6:3 = stablehlo.while(%iterArg_8 = %c_1, %iterArg_9 = %3, %iterArg_10 = %cst) : tensor<i64>, tensor<f64>, tensor<3xf64> 
           91:  cond { 
           92:  %10 = stablehlo.compare LT, %iterArg_8, %c_0 : (tensor<i64>, tensor<i64>) -> tensor<i1> 
           93:  stablehlo.return %10 : tensor<i1> 
           94:  } do { 
next:120'0            X error: no match found
           95:  %10 = stablehlo.subtract %c_0, %c_2 : tensor<i64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           96:  %11 = stablehlo.subtract %10, %iterArg_8 : tensor<i64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
next:120'1      ?                                                       possible intended match
           97:  %12 = stablehlo.add %5, %11 : tensor<i64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           98:  %13 = stablehlo.multiply %c_2, %12 : tensor<i64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
           99:  %14 = stablehlo.add %13, %c_2 : tensor<i64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          100:  %15 = stablehlo.convert %14 : (tensor<i64>) -> tensor<f64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          101:  %16 = stablehlo.reshape %15 : (tensor<f64>) -> tensor<1xf64> 
next:120'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            .
            .
            .
>>>>>>

--

********************
********************
Failed Tests (1):
  Enzyme-JaX :: lit_tests/diffrules/stablehlo/while_checkpointing.mlir
```

sriharikrishna:
@wsmoses The tests all pass now.

src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -3060,6 +3151,11 @@ struct WhileOpEnzymeOpsRemover
       auto newType =
           cast<ShapedType>(cast<AutoDiffTypeInterface>(cinfo.cachedType())
                                .getShadowType(numIters));
+      // dynamic_update_slice requires operand rank >= 1. For scalar cache use
+      // 1D.
+      if (newType.getRank() == 0) {
```
Pangoraw, line 3156: could you explain how you hit this?

newType being from `getShadowType(numIters)` should be a `tensor<numItersx?x...>`

sriharikrishna: This is needed for the edge case where checkpoint period equals trip count (N=4, period=4).
Then `nOuter=1`, so the scalar cache has size `1`. `getShadowType(1)` for `tensor<f64>`
returns 0-D (tensor<f64>), which is invalid for `dynamic_update_slice`. This change fixes this by using a 1D cache type for these cases. 

Given that we are allowing users to set the checkpointing period, it is possible that this issue will be triggered. 

My guess is that the sqrt checkpoint test did not check the edge case of checkpointing for a single iteration, which is the only time this issue would be triggered.

Pangoraw: oh right, that makes sense

PR #1931, opened 2026-01-12
Introduce patterns provenance-based DUS simplification

ftynse:
- Provide an analysis that computes, for each element in a tensor, the
  operation or block argument from which that element originates and its
  coordinates in the original tensor. This analysis is based on the
  Presburger library and supports slice/pad/DUS, more operations may be
  added as long as their manipulation of tensor elements is
  representable by a Presburger formula.

- Use this analysis to move slices appearing between two DUS operations
  to use the result of the second DUS when provenance of the sliced
  value remains the same. This minimizes the live range of the slice
  and thus positively impacts on bufferization due to lower overall
  memory footprint.

- Define a new `piecewise_select` operation that constructs a tensor
  from hyperrectangular blocks in different tensors. It can be seen as a
  compact, analyzable representaiton of a chain of selects.

- Define a pattern reconstructing a DUS transitively fed by another DUS
  from the tensors that serve as provenance sources of its values using
  the provenance analysis, thus removing the DUS and potentially the
  source DUS in favor of cheaper operations.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -469,6 +469,27 @@ def UpdateWithoutCornersOp: EnzymeXLA_Op<
   let results = (outs HLO_Tensor:$result);
 }
 
+def PiecewiseSelectOp : EnzymeXLA_Op<"piecewise_select", [Pure, SameOperandsAndResultType]> {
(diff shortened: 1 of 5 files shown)


wsmoses (commented):
can you add the requisite test, and add to transform ops?

ftynse:
done

wsmoses:
hm tests seem to crash

PR #1941, opened 2026-01-19
add lapack.HeevOp for eigendecomposition

mofeing:
WIP

### DIFF
--- a/src/enzyme_ad/jax/Dialect/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Ops.cpp
@@ -1166,15 +1166,15 @@ LogicalResult enzymexla::HeevOp::verify() {
     return emitOpError("Input matrix must be square");
   }
 
-  auto jobz = getJobz();
-  if (jobz != enzymexla::Jobz::N && jobz != enzymexla::Jobz::V) {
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Passes/LowerEnzymeXLALapack.cpp
```diff
@@ -2708,6 +2708,201 @@ struct GesvjOpLowering : public OpRewritePattern<enzymexla::GesvjOp> {
...
+        return rewriter.notifyMatchFailure(op,
+                                           "unsupported input element type");
+      }
+    } else {
+      op->emitOpError() << "Unsupported element type: " << inputElementType;
+      return rewriter.notifyMatchFailure(op, "unsupported input element type");
+    }
+
+    std::string bind_fn = "enzymexla_lapacke_" + fn;
+    std::string wrapper_fn = "enzymexla_wrapper_lapacke_" + fn;
```
avik-pal, line 2930: Can we use the fortran ones instead of lapacke? else we will have to load multiple lapack implementations, and LA.jl ships by default with the fortran ones

mofeing: yeah, no problem. I will need to change the QR but on another PR.

Issue #1949, opened 2026-01-21
MultiRotate Op and MultiSliceOp


wsmoses:
A MultiRotate op has the following semantics

rotleft2, rotleft1, rotate0, rotright1, rotright2 = multi_rotate %x, dimension=, left_amount=2, right_amount=2

This has the semantics that rotleft2 = rotate x left by 2, rotleft2 = rotate x left by 1, etc

This should have an EnzymeHLOOpt for the case where the leftmost or rightmost operands are not used, in which case the operand is replace it with a smaller multi_rotate. If only one result is used, it can be replaced with a single rotate op.

A MultiSlice op has the following semantics

sliceleft2, sliceleft1, slice, sliceright2, sliceright2 = multi_slice %x[start0:end0:step0, start1:end1:step1, ....], dimension=, left_amount=2, right_amount=2

Suppose dimension = 0, this has the semantics that sliceleft2 = x[start0:end0-4:step0, start1:end1:step1, ....]; sliceleft1 = x[start0+1:end0-3:step0, start1:end1:step1, ....]; 

This similarly should have an EnzymeHLOOpt for the case where the leftmost or rightmost operands are not used, in which case the operand is replace it with a smaller multi_slice. If only one result is used, it can be replaced with a single slice op.

For both new operands, also write patterns to lower the multi versions of op into the corresponding single versions (as above), preserving sharding. Additionally write versions to recognize as large a width multi-op as there are available users [aka all results must be used]. Furthermore, implement constant propagation, folders, and CSE for the operations. Finally add an additional CSE for slice and rotate that will re-use an existing multi-op result, if it already computes the result.

wsmoses:
cc @avik-pal @giordano @jumerckx @ftynse

jumerckx:
> A MultiSlice op has the following semantics
> 
> sliceleft2, sliceleft1, slice, sliceright2, sliceright2 = multi_slice %x[start0:end0:step0, start1:end1:step1, ....], dimension=, left_amount=2, right_amount=2
> 
> Suppose dimension = 0, this has the semantics that sliceleft2 = x[start0:end0-4:step0, start1:end1:step1, ....]; sliceleft1 = x[start0+1:end0-3:step0, start1:end1:step1, ....];

What happens if left_amount != right_amount? Shouldn't the offsets be centered around start and end, like:
```
start-2:end-2
start-1:end-1
start+0:end+0
start+1:end+1
start+2:end+2
```

[...]

PR #1953, opened 2026-01-21
Add multirotate and multislice operations

jumerckx:
ref: #1949


src/enzyme_ad/jax/Dialect/Ops.cpp
```diff
@@ -2487,3 +2487,118 @@ OpFoldResult ExtendOp::fold(FoldAdaptor adaptor) {
...
+  }
+
+  // Verify number of results
+  int64_t expectedNumResults = leftAmount + rightAmount + 1;
+  if ((int64_t)getNumResults() != expectedNumResults)
+    return emitOpError("expected ")
+           << expectedNumResults
+           << " results (left_amount + right_amount + 1), got "
+           << getNumResults();
+
```
wsmoses, line 2583: we should also check the start/limit indices are in bounds [especially relevant with the right/left amount]

PR #1954, opened 2026-01-21
Add multirotate simplification

jumerckx:
As described in https://github.com/EnzymeAD/Enzyme-JAX/issues/1949

### DIFF
--- a/test/lit_tests/reduce_unused_multi_rotate.mlir
+++ b/test/lit_tests/reduce_unused_multi_rotate.mlir
@@ -1,7 +1,7 @@
 // RUN: enzymexlamlir-opt --enzyme-hlo-generate-td="patterns=reduce_unused_multirotate" --transform-interpreter --enzyme-hlo-remove-transform %s | FileCheck %s
 
 // CHECK:      func.func @multirotate(%arg0: tensor<20x24x80xf64>) -> (tensor<20x24x80xf64>, tensor<20x24x80xf64>, tensor<20x24x80xf64>, tensor<20x24x80xf64>) {
-// CHECK-NEXT:   %0 = "enzymexla.rotate"(%arg0) <{amount = -1 : si32, dimension = 2 : si32}> : (tensor<20x24x80xf64>) -> tensor<20x24x80xf64>
+// CHECK-NEXT:   %0 = "enzymexla.rotate"(%arg0) <{amount = 79 : si32, dimension = 2 : si32}> : (tensor<20x24x80xf64>) -> tensor<20x24x80xf64>
 // CHECK-NEXT:   %1:3 = "enzymexla.multi_rotate"(%arg0) <{dimension = 2 : si32, left_amount = 1 : si32, right_amount = 1 : si32}> : (tensor<20x24x80xf64>) -> (tensor<20x24x80xf64>, tensor<20x24x80xf64>, tensor<20x24x80xf64>)
 // CHECK-NEXT:   return %arg0, %0, %1#0, %1#2 : tensor<20x24x80xf64>, tensor<20x24x80xf64>, tensor<20x24x80xf64>, tensor<20x24x80xf64>
 


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30279,6 +30279,98 @@ struct ReduceWindowWrapSimplify final
...
+      }
+    }
+
+    // If all results are used, nothing to optimize
+    if (usedCount == totalResults)
+      return failure();
+
+    // If no results are used, this should be handled by dead code elimination
+    if (usedCount == 0)
+      return failure();
```
wsmoses: eh we might as well just do eraseop here

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30279,6 +30279,98 @@ struct ReduceWindowWrapSimplify final
...
+      rewriter.eraseOp(op);
+      return success();
+    }
+
+    // Otherwise, create a smaller MultiRotateOp
+    if (newLeftAmount != leftAmount || newRightAmount != rightAmount) {
+      auto newOp = rewriter.create<enzymexla::MultiRotateOp>(
+          op.getLoc(),
+          SmallVector<Type>(newLeftAmount + newRightAmount + 1,
+                            op.getOperand().getType()),
```
wsmoses, line 30373: also make sure to keep shardings

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: we still need to return success if we did something

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30279,6 +30279,109 @@ struct ReduceWindowWrapSimplify final
...
+      int amount = centerIdx -
+                   usedIdx; // Positive = rotate left, negative = rotate right
+
+      auto rotateOp = rewriter.create<enzymexla::RotateOp>(
+          op.getLoc(), op.getOperand().getType(), op.getOperand(),
+          rewriter.getSI32IntegerAttr(amount), op.getDimensionAttr());
+      // Propagate sharding if present
+      if (auto shard = sdy::getShardingPerValue(op)) {
+        sdy::setShardings(rotateOp, shard);
+      }
```
wsmoses, line 30361: we also should preserve shardings here.

note also more simply we can just do rewriter.replaceOpWithNewOp<...>(op, ....);

note that you need to save the sharding before the replaceOp, since the op gets deleted, and the sharding attr gets added after

jumerckx: > we also should preserve shardings here.

This might have referred to an outdated version? or should other sharding information be carried over?
Anyhow, I've updated with `rewriter.replaceOpWithNewOp `, thank you!

src/enzyme_ad/jax/Passes/EnzymeHLOPatterns.h
```diff
@@ -134,5 +134,9 @@ void addSelfMulToConvolutionLike(RewritePatternSet &patterns,
                                  MLIRContext &context, PatternBenefit benefit);
 void addEnzymeHLOUnroll(RewritePatternSet &patterns, int64_t maxNumIterations,
                         MLIRContext &context, PatternBenefit benefit);
+void addMultiRotateOpt(RewritePatternSet &patterns, MLIRContext &context,
```
wsmoses: Last thing otherwise good to merge, we don’t need this transform add multi rotate right ?

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30293,6 +30293,109 @@ struct ReduceWindowWrapSimplify final
...
+
+    // Calculate new left and right amounts
+    int centerIdx = leftAmount;
+    int newLeftAmount = centerIdx - firstUsed;
+    int newRightAmount = lastUsed - centerIdx;
+
+    // If only one result is used, replace with a single RotateOp
+    if (usedCount == 1) {
+      int usedIdx = firstUsed;
+      int amount = centerIdx -
```
wsmoses, line 30348: I don’t think our op supports negative rotates, so you’ll need to check if negative and if so += by the length of the axis

PR #1956, opened 2026-01-21
RecognizeMultiRotate

jumerckx:
ref: #1949


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30279,6 +30279,82 @@ struct ReduceWindowWrapSimplify final
...
+          amounts.push_back(rotate.getAmount());
+        }
+      }
+    }
+
+    // Need at least 2 rotations to combine
+    if (rotates.size() < 2)
+      return failure();
+
+    // To avoid processing the same set of rotations multiple times,
```
jumerckx: Is this necessary?

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30279,6 +30279,82 @@ struct ReduceWindowWrapSimplify final
...
+    //   result[leftAmount] = rotate by 0 (center)
+    //   result[leftAmount + rightAmount] = rotate right by rightAmount
+    int32_t leftAmount = maxAmount > 0 ? maxAmount : 0;
+    int32_t rightAmount = minAmount < 0 ? -minAmount : 0;
+    int32_t totalResults = leftAmount + rightAmount + 1;
+
+    // Check density - only combine if utilization is reasonable
+    // Avoid creating a MultiRotateOp with many unused intermediate results
+    if (totalResults > 2 * static_cast<int32_t>(rotates.size()))
+      return failure();
```
jumerckx, line 30412: Do we need such a check? This one might be too lenient?

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: I think this is useful

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: I think we should throw if there's any gaps [also checking if an op is use_empty], though we don't need to return failure, we can just resize [and select the relevant subset].

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30279,6 +30279,79 @@ struct ReduceWindowWrapSimplify final
...
+
+    // Create the MultiRotateOp
+    auto newOp = rewriter.create<enzymexla::MultiRotateOp>(
```
wsmoses, line 30427: we need to move the newOp to be defined right after the operand, otherwise we risk a dominace errror

jumerckx:
I've been confusing myself. 
Do you consider a positive rotation amount to be a leftwards or rightwards rotation?
i.e.
```
a = rotate(x, amount=0)
b = rotate(x, amount=1)
c = rotate(x, amount=2)
-->
c, b, a = multirotate(x, left=2, right=0)

or

a, b, c = multirotate(x, left=0, right=2)

```
Currently it's implemented as the former but I now think the latter might be more logical?

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
jumerckx: I added filtering for a contiguous sequence that neighbors or includes rotate(x, 0)

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30293,6 +30293,151 @@ struct ReduceWindowWrapSimplify final
...
+        groupEnd = sortedAmounts[i];
+      } else {
+        // Save current group and start new one
+        contiguousGroups.push_back({groupStart, groupEnd});
+        groupStart = sortedAmounts[i];
+        groupEnd = sortedAmounts[i];
+      }
+    }
+    contiguousGroups.push_back({groupStart, groupEnd});
+
```
wsmoses, line 30357: we should do the contiguous group check before the 2 rotates [in case we reduce the number because not contiguous]

jumerckx: There's a second check as well:
```cpp
    // No qualifying groups found
    if (qualifyingAmounts.size() < 2)
      return failure();
```

wsmoses: Ah okay

PR #1960, opened 2026-01-22
fix: properly set insertion point for lowerExtend

avik-pal:
- [x] fixes https://github.com/EnzymeAD/Reactant.jl/issues/2151
- [x] add a test


wsmoses (commented):
we should also check the other lower* helpers too

wsmoses (approved):

PR #1972, opened 2026-01-24
fix: check dominance in dynamic_slice elementwise

avik-pal:
fixes part of https://github.com/EnzymeAD/Reactant.jl/issues/2061

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -3294,6 +3294,7 @@ struct DynamicSliceElementwise final
     }
 
     llvm::SetVector<stablehlo::DynamicSliceOp> allDSUsers;
+    stablehlo::DynamicSliceOp firstDSOp;
 
     for (auto user : elem->getUsers()) {
       auto dsOp = dyn_cast<stablehlo::DynamicSliceOp>(user);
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -3311,6 +3312,11 @@ struct DynamicSliceElementwise final
         continue;
       }
       allDSUsers.insert(dsOp);
+      if (!firstDSOp) {
+        firstDSOp = dsOp;
+      } else if (dsOp->isBeforeInBlock(firstDSOp)) {
```
wsmoses, line 3317: This is still a bit expensive of a query, see cse for some examples we can re use for speeding up

wsmoses (approved):

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
avik-pal: there we are also using isBeforeInBlock

```
        if (nop->isBeforeInBlock(op)) {
          rewriter.replaceOp(op, nop);
          return success();
        } else {
          rewriter.replaceOp(nop, op);
          return success();
        }
```

wsmoses: Hm I thought we did some number of checks before or behind the current first and then did the more expensive one, but maybe not

wsmoses (approved):

PR #1980, opened 2026-01-27
Separate MultiRotateSpmdOptimize into multirotate_spmd pass argument

wsmoses:
LLM-generated


src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
```diff
@@ -2278,6 +2278,341 @@ struct RotateToPadCommOptimize : public OpRewritePattern<enzymexla::RotateOp> {
...
+    int64_t padding = shard_size * total_mesh_size - full_size;
+
+    int64_t left_amount = rotate.getLeftAmount();
+    int64_t right_amount = rotate.getRightAmount();
+
+    if (left_amount > shard_size || right_amount > shard_size) {
+      return rewriter.notifyMatchFailure(rotate,
+                                         "Rotation amount > shard size, not "
+                                         "handled by this optimization yet.");
+    }
```
jumerckx, line 2336: Is this ok?

wsmoses: yeah, one of the many reasons we have the existing lowering [though we should also add your lowering here, at a lesser priority]

Issue #1970, opened 2026-01-23
Reverse-mode AD fails for stablehlo.reduce with multiply (product accumulation in raised kernels)


KookiesNKareem:
`error: Unsupported operation in reduction rev autodiff(1): ` `Julia 1.10.10, Reactant v0.2.160, Enzyme v0.13.70, KernelAbstractions v0.9.38, CUDA v5.8.3`

```bash
%6 = "stablehlo.reduce"(%5, %0) <{dimensions = array<i64: 0>}> ({
^bb0(%arg2: tensor<f32>, %arg3: tensor<f32>):
  %8 = "stablehlo.multiply"(%arg2, %arg3) : (tensor<f32>, tensor<f32>) -> tensor<f32>
  "stablehlo.return"(%8) : (tensor<f32>) -> ()
}) : (tensor<9xf32>, tensor<f32>) -> tensor<f32>
```

```julia
using Reactant: @compile, CompileOptions, to_rarray, set_default_backend
using KernelAbstractions: @kernel, @index, @Const, get_backend
import Enzyme, CUDA

set_default_backend("gpu")

@kernel function loop_no_array_kernel!(y, @Const(x), N_iters)
    i = @index(Global)
    if i <= length(x)
        val = x[i]
        iter = Int32(1)
        while iter <= N_iters
            val = val * 0.9f0
            iter += Int32(1)
        end
        y[i] = val
    end
end

@kernel function loop_with_array_kernel!(y, @Const(x), @Const(coeffs), N_iters)
    i = @index(Global)
    if i <= length(x)
        val = x[i]
        iter = Int32(1)
        while iter <= N_iters
            val = val * coeffs[iter]
            iter += Int32(1)
        end
        y[i] = val
    end
end

function apply_loop_no_array(x, n)
    y = similar(x)
    kernel! = loop_no_array_kernel!(get_backend(x), 256)
    kernel!(y, x, Int32(n); ndrange=length(x))
    return y
end

function apply_loop_with_array(x, coeffs)
    y = similar(x)
    kernel! = loop_with_array_kernel!(get_backend(x), 256)
    kernel!(y, x, coeffs, Int32(length(coeffs)); ndrange=length(x))
    return y
end

loss_no_array(x) = sum(apply_loop_no_array(x, 10))
loss_with_array(x, coeffs) = sum(apply_loop_with_array(x, coeffs))

x = to_rarray(randn(Float32, 100))
[...]

PR #1998, opened 2026-01-28
diffrule for reduce with mul operation

Pangoraw:
Closes https://github.com/EnzymeAD/Enzyme-JAX/issues/1970


wsmoses (commented):
lgtm, but @avik-pal should review

avik-pal (approved):

PR #2005, opened 2026-01-28
add value for amd chips and abi version

xys-syx:
For issue 2004, I temporarily add value for amd target attributes, chips as gfx1030, abi version as 600, now can fix the issues

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -27160,6 +27160,9 @@ private:
             extraValues.push_back(current);
             continue;
           }
+        } else {
+          extraValues.push_back(current);
+          continue;
         }
       } else if (auto reduce =
                      current.template getDefiningOp<stablehlo::ReduceOp>()) {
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/GPULaunchRecognition.cpp
```diff
@@ -65,12 +65,12 @@ struct GPULaunchRecognitionPass
     Attribute target;
     if (backend == "rocm") {
       // Here temporarily set as "" for ROCm backend
```
ivanradanov:
```suggestion
```

src/enzyme_ad/jax/Passes/GPULaunchRecognition.cpp
ivanradanov, line 63: Can we obtain the ABI and chip information like this for AMD as well?

xys-syx: For CUDA backend, we retrieve the chip info from the target-cpu attribute that Clang compiling CUDA source to LLVM IR.
For ROCm backend, the input is still a CUDA program, so Clang generates NVIDIA target attributes (sm_XX) in the LLVM IR  while not AMD GPU specs. We can't extract AMD chip info from the passthrough attributes the same way.
Do we need to implement something for adding more pass argument like "-reactant-rocm-arch=gfx1030" and "-reactant-rocm-abi=600" in the command line and can get value from there?

ivanradanov: Ah yes, makes sense. Yeah we need a way to specify chips and if possible features. Similar to how we (I think) have it in the parallel->gpu pass

ivanradanov: Although not immediately necessary because when we do the cuda->hip translation, we will first transform to parallel and then transform to hip, so if the parallel->hip part has the option then we should be good.

ivanradanov (approved):

wsmoses (approved):

vimarsh6739 (approved):

PR #2009, opened 2026-01-29
feat: dot_general to symm + fusion passes

snonk:


src/enzyme_ad/jax/primitives.py
avik-pal, line 383: add the fusion passes where syrk ones are defined

avik-pal (commented):
patterns look good, but tests need to be updated with the new IR annotatiosns

avik-pal (approved):

wsmoses (approved):

PR #2023, opened 2026-02-02
Add RecognizeMultiSlice

jumerckx:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30638,6 +30638,207 @@ struct ReduceWindowWrapSimplify final
...
+      SmallVector<Type> resultTypes(totalResults, resultType);
+
+      rewriter.setInsertionPointAfterValue(input);
+      auto newOp = rewriter.create<enzymexla::MultiSliceOp>(
+          op.getLoc(), resultTypes, input, adjustedStartIndices,
+          adjustedLimitIndices, strides, (int32_t)dim, amount);
+
+      // Propagate sharding if present
+      if (auto shard = sdy::getShardingPerValue(op)) {
+        sdy::setShardings(newOp, shard);
```
wsmoses: @avik-pal can you support here I think this [and also recognizemultirotate, in retrospect] may be wrong. Specifically the original slice has only one result, whereas this has multiple, so we need a sharding which replicates it across all values

avik-pal: Yes we need to set for each of the results  https://openxla.org/shardy/sdy_dialect#tensorshardingpervalueattr

jumerckx: Thanks, will try this in a bit!

jumerckx: I tried here but unsure if this makes sense: [`a02f53d` (#2023)](https://github.com/EnzymeAD/Enzyme-JAX/pull/2023/commits/a02f53d09b4748941f2a0d91693ee446c4ac445b)

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -30638,6 +30638,231 @@ struct ReduceWindowWrapSimplify final
...
+          adjustedLimitIndices, strides, (int32_t)dim, amount);
+
+      // Propagate sharding if present (all slices have the same sharding)
+      if (commonSharding.has_value() && commonSharding.value()) {
+        auto shardings = commonSharding.value().getShardings();
+        if (!shardings.empty()) {
+          sdy::TensorShardingAttr singleShard = shardings[0];
+          SmallVector<sdy::TensorShardingAttr> newShardings(totalResults,
+                                                            singleShard);
+          sdy::setShardings(newOp, sdy::TensorShardingPerValueAttr::get(
```
wsmoses, line 32056: Looks good, @jumerckx can you also port this to recognizemultirotate?

jumerckx: done

PR #2033, opened 2026-02-05
Transpose-like slice and elemwise

wsmoses:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -2884,6 +2957,9 @@ struct SliceBroadcast final
     if (!bcast)
       return failure();
 
+    if (isTransposeReshapeLikeBroadcast(bcast))
+      return failure();
+
     SmallVector<int64_t> nbcast_idx;
```
jumerckx, line 2963: unrelated to this pr, but `nbcast_idx` seems dead code?

PR #2037, opened 2026-02-06
Make if_rocm_is_configured to be true and add AMD support for generating binary code

xys-syx:


BUILD
```diff
@@ -89,6 +89,8 @@ cc_library(
         "@llvm-project//mlir:NVVMToLLVMIRTranslation",
         "@tsl//tsl/platform:env",
         "@tsl//tsl/platform:env_impl",
+        "@llvm-project//llvm:AMDGPUAsmParser",
```
wsmoses: this should only be done if available

xys-syx: Yeah, but I did not find one exist for AMD GPU https://github.com/openxla/xla/blob/1427d1b67e3c092ef912358ec2ff5aa6292522cf/xla/tsl/platform/build_config_root.bzl#L8 , is there any other way to achieve this? would add a custom .bzl a good way to solve this?

workspace.bzl
```diff
@@ -175,6 +175,9 @@ sed -i.bak0 "s/patch_cmds = \\[/patch_cmds = \\[\\\"find . -type f -name config.
 
     sed -i.bak0 "s/urls = /patch_cmds = \\[\\\"sed -i.bak -e 's\\/return PACKET_TYPE(0) == PACKET_TYPE(0);\\/return (PACKET_TYPE)(PACKET_TYPE(0) == PACKET_TYPE(0));\\/g' -e 's\\/return CAST_FROM_INT(CAST_TO_INT(a) == CAST_TO_INT(a));\\/return CAST_FROM_INT((decltype(CAST_TO_INT(a)))(CAST_TO_INT(a) == CAST_TO_INT(a)));\\/' Eigen\\/src\\/Core\\/arch\\/clang\\/PacketMath.h\\\"\\],urls = /g" third_party/eigen3/workspace.bzl
     """,
+    """
+sed -i.bak0 "s/patch_cmds = \\[/patch_cmds = \\[\\\"find . -type f -name BUILD.bazel -path '*\\/mlir\\/*' -exec sed -i.bak0 's\\/MLIR_ENABLE_ROCM_CONVERSIONS 0\\/MLIR_ENABLE_ROCM_CONVERSIONS 1\\/g' {} +\\\",/g" third_party/llvm/workspace.bzl
```
xys-syx: I think for merging to main, here also need to add the logic like if_rocm_is_confirgued = true -> transform MLIR_ENABLE_ROCM_CONVERSIONS from 0 to 1, but I keep like this now as I have not achieve this "Set TF_NEED_ROCM=1, bazel auto-detect the complete rocm toolkit, including path, targets, libs, headers -> generate @local_config_rocm and if_rocm_is_configured=true"

PR #2051, opened 2026-02-08
Bessel lowering

jumerckx:
Putting this up as a WIP draft mainly to hear if it's possible to get constprop to work in lit tests? I'd like to test the output instead of only the generated code, but simply running `enzyme-hlo-opt` doesn't inline and constprop as I'd hoped.


src/enzyme_ad/jax/Passes/LowerEnzymeXLABessel.cpp
```diff
@@ -0,0 +1,405 @@
+#include "Enzyme/MLIR/Dialect/Dialect.h"
+#include "Enzyme/MLIR/Passes/EnzymeBatchPass.h"
+#include "src/enzyme_ad/jax/Dialect/Dialect.h"
+#include "src/enzyme_ad/jax/Dialect/Ops.h"
+#include "src/enzyme_ad/jax/Passes/Passes.h"
```
wsmoses, line 5: there are some examples within enzymehloopt.cc where essentially if the input is a constant we will lower the chlo op immediately [and the same would apply here]

PR #2064, opened 2026-02-10
feat: rewrite if with binary op to a select

### LINKED ISSUE #2057: Rewrite conditional binary op with binary op with init value
```mlir
module @reactant_add_cond attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<f64> {enzymexla.memory_effects = []}, %arg1: tensor<f64> {enzymexla.memory_effects = []}) -> tensor<f64> attributes {enzymexla.memory_effects = []} {
    %cst = stablehlo.constant dense<4.000000e+00> : tensor<f64>
    %0 = stablehlo.compare  GT, %arg1, %cst : (tensor<f64>, tensor<f64>) -> tensor<i1>
    %1 = "stablehlo.if"(%0) ({
      %2 = stablehlo.add %arg1, %arg0 : tensor<f64>
      stablehlo.return %2 : tensor<f64>
    }, {
      stablehlo.return %arg1 : tensor<f64>
    }) : (tensor<i1>) -> tensor<f64>
    return %1 : tensor<f64>
  }
}
```

This can be rewritten as a `add(%arg1, select(%0, %arg0, 0))`. Might not be beneficial always but we should atleast always do this inside loops

avik-pal:
fixes #2057

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -14918,8 +14918,9 @@ struct IfBinaryOpToSelectBinaryOp final
       }
     }
 
-    if (!anyHoisted)
+    if (!anyHoisted) {
       return failure();
+    }
 
     // If all results can be hoisted, replace the entire if op
     bool allHoisted = llvm::all_of(hoistedResults,
@@ -14988,8 +14989,9 @@ private:
                          Region &branchRegion, bool isTrueBranch,
                          PatternRewriter &rewriter) const {
     auto binOp = branchVal.getDefiningOp<BinaryOpType>();
-    if (!binOp)
+    if (!binOp) {
       return nullptr;
+    }
 
     Value base = nullptr;
     Value other = nullptr;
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -14852,6 +14852,210 @@ struct IfToSelect final
   }
 };
 
+// Transforms conditional binary operations into binary op with select.
+// Pattern: if(pred) { binop(base, other) } else { base } => binop(base,
```
wsmoses, line 14856: I'm not sure this is the way we'd want to do this.

For example we could instead rewrite this by sinking the binop below the if, and replacing the if result with the select. This is more general (and doesn't require the identity), or even knowing its a binary operator.

However the one thing that does apply in both cases is that it needs to be legal & profitable to speculate the if. In that case this might just be better done as a general "SpeculateIf" pattern. We actually already have one example within enzymehloopt (https://github.com/EnzymeAD/Enzyme-JAX/blob/c36c22dec1c89ce24dfbc12c1a72db59aeb7dc3b/test/lit_tests/speculate_if_pad_to_select.mlir), and several in other mlir dialects (particularly during raising)

PR #2065, opened 2026-02-10
fix: use single scatter and gather after splitting complex

avik-pal:
stablehlo::BitcastConvertOp doesnt allow reinterpretting complex -> real

### DIFF
--- a/test/lit_tests/split_complex_gather.mlir
+++ b/test/lit_tests/split_complex_gather.mlir
@@ -16,22 +16,21 @@ func.func @test_basic_gather(%input: tensor<4xcomplex<f32>>, %indices: tensor<2x
 }
 // CHECK: %[[REAL:.*]] = stablehlo.real %arg0
 // CHECK: %[[IMAG:.*]] = stablehlo.imag %arg0
-// CHECK: %[[REAL_R:.*]] = stablehlo.reshape %[[REAL]] : (tensor<4xf32>) -> tensor<4x1xf32>
-// CHECK: %[[IMAG_R:.*]] = stablehlo.reshape %[[IMAG]] : (tensor<4xf32>) -> tensor<4x1xf32>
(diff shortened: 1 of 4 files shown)


test/lit_tests/split_complex_gather.mlir
```diff
@@ -14,25 +14,25 @@ func.func @test_basic_gather(%input: tensor<4xcomplex<f32>>, %indices: tensor<2x
...
 // CHECK-SAME:   collapsed_slice_dims = [0],
 // CHECK-SAME:   start_index_map = [0],
 // CHECK-SAME:   index_vector_dim = 1
 // CHECK-SAME: >,
-// CHECK-SAME: slice_sizes = array<i64: 1>
-// CHECK-SAME: (tensor<4xf32>, tensor<2x1xi64>) -> tensor<2xf32>
-// CHECK: %[[RESULT:.*]] = stablehlo.complex %[[REAL_GATHER]], %[[IMAG_GATHER]]
+// CHECK-SAME: slice_sizes = array<i64: 1, 2>
+// CHECK-SAME: (tensor<4x2xf32>, tensor<2x1xi64>) -> tensor<2x2xf32>
+// CHECK: %[[REAL_S:.*]] = stablehlo.slice %[[GATHER]] [0:2, 0:1] : (tensor<2x2xf32>) -> tensor<2x1xf32>
```
wsmoses: we should test whether its better to have the extra dimension at the start or end.

At the end is more similar to what the original data may have been stored as. However for TPU, we need the last dim to be a factor of 128

wsmoses:
just curious what was the BitcastConvertOp error that was thrown

test/lit_tests/split_complex_gather.mlir
avik-pal: flipped to be the first dim

wsmoses (approved):

test/lit_tests/split_complex_gather.mlir
wsmoses: this is fine to land for now, but we should attempt to test this later [and potentially turn it into a flag for the pass on first vs last dim]

PR #2066, opened 2026-02-11
 Correct atan2 gradient

ptiede:
There was a minus sign offset. 

https://en.wikipedia.org/wiki/Atan2

Gradient of `arctan2(y, x)` is

```
d arctan2(y, x) = x/(x^2 + y^2) dy - y/(x^2 + y^2) dx
```
annoying part is that y usually comes first...

### DIFF
--- a/test/lit_tests/diffrules/stablehlo/atan2.mlir
+++ b/test/lit_tests/diffrules/stablehlo/atan2.mlir
@@ -7,13 +7,11 @@ func.func @main(%a : tensor<2xf32>, %b : tensor<2xf32>) -> tensor<2xf32> {
 }
 
 // FORWARD:  func.func @main(%arg0: tensor<2xf32>, %arg1: tensor<2xf32>, %arg2: tensor<2xf32>, %arg3: tensor<2xf32>) -> (tensor<2xf32>, tensor<2xf32>) {
-// FORWARD-NEXT:    %0 = stablehlo.multiply %arg0, %arg3 : tensor<2xf32>
-// FORWARD-NEXT:    %1 = stablehlo.multiply %arg2, %arg1 : tensor<2xf32>
+// FORWARD-NEXT:    %0 = stablehlo.multiply %arg2, %arg1 : tensor<2xf32>
+// FORWARD-NEXT:    %1 = stablehlo.multiply %arg0, %arg3 : tensor<2xf32>
 // FORWARD-NEXT:    %2 = stablehlo.subtract %0, %1 : tensor<2xf32>
-// FORWARD-NEXT:    %cst = stablehlo.constant dense<2.000000e+00> : tensor<2xf32>
-// FORWARD-NEXT:    %3 = stablehlo.power %arg0, %cst : tensor<2xf32>
-// FORWARD-NEXT:    %cst_0 = stablehlo.constant dense<2.000000e+00> : tensor<2xf32>
-// FORWARD-NEXT:    %4 = stablehlo.power %arg2, %cst_0 : tensor<2xf32>
+// FORWARD-NEXT:    %3 = stablehlo.multiply %arg2, %arg2 : tensor<2xf32>
+// FORWARD-NEXT:    %4 = stablehlo.multiply %arg0, %arg0 : tensor<2xf32>
 // FORWARD-NEXT:    %5 = stablehlo.add %3, %4 : tensor<2xf32>
 // FORWARD-NEXT:    %6 = stablehlo.divide %2, %5 : tensor<2xf32>
 // FORWARD-NEXT:    %7 = stablehlo.atan2 %arg0, %arg2 : tensor<2xf32>


src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -971,8 +971,8 @@ def : HLODerivative<"AddOp", (Op $x, $y),
 
 def : HLODerivative<"Atan2Op", (Op $x, $y),
                     [
-                      (CheckedMul (DiffeRet), (Div (Neg $y), (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y))))),
-                      (CheckedMul (DiffeRet), (Div $x, (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y)))))
+                      (CheckedMul (DiffeRet), (Div $y, (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y))))),
```
wsmoses: actually @ptiede while you're at it, can you replace pow(x, 2) with just Mul(x, x) and such

it'll be faster and better here

ptiede: Ok I think I did the correct thing

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -971,10 +971,10 @@ def : HLODerivative<"AddOp", (Op $x, $y),
 
 def : HLODerivative<"Atan2Op", (Op $x, $y),
                     [
-                      (CheckedMul (DiffeRet), (Div (Neg $y), (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y))))),
-                      (CheckedMul (DiffeRet), (Div $x, (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y)))))
+                      (CheckedMul (DiffeRet), (Div $y, (Add (Mul $x, $x), (Mul $y, $y)))),
+                      (CheckedMul (DiffeRet), (Div (Neg $x), (Add (Mul $x, $x), (Mul $y, $y))))
                     ],
-                    (CheckedDiv (Sub (Mul $x, (Shadow $y)), (Mul $y, (Shadow $x))), (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y))))
+                    (CheckedDiv (Sub (Mul $x, (Shadow $y)), (Mul $y, (Shadow $x))), (Add (Mul $x, $x), (Mul $y, $y)))
```
wsmoses, line 977: @ptiede if you negated the forward mode version, i think the reverse one here at the bottom is also wrong and needs negating

wsmoses: or rather the operands of the sub should be inverted

ptiede: Ok I was wondering what that was xD

ptiede: Oh there wasn't even a negation here. That seems wrong.

ptiede: @wsmoses ok I'm pretty sure the forward version was doubly broken and x and y where swapped.

src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -971,10 +971,10 @@ def : HLODerivative<"AddOp", (Op $x, $y),
 
 def : HLODerivative<"Atan2Op", (Op $x, $y),
                     [
-                      (CheckedMul (DiffeRet), (Div (Neg $y), (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y))))),
-                      (CheckedMul (DiffeRet), (Div $x, (Add (Pow $x, (HLOConstantFP<"2">)), (Pow $y, (HLOConstantFP<"2"> $y)))))
+                      (CheckedMul (DiffeRet), (Div $y, (Add (Mul $x, $x), (Mul $y, $y)))),
```
wsmoses: @ptiede so this currently says

forward
d/dx = y / (x^2+y^2)
d/dy = -x / (x^2+y^2)

reverse is

( y dx - (-x) * dy ) / (x^2 + y^2) 

which is most definitely wrong


the old one said


d/dx = -y / (x^2+y^2)
d/dy = x / (x^2+y^2)

reverse is

( x dy - y * dx ) / (x^2 + y^2)

ptiede: Why is it
```
( y dx - (-x) * dy ) / (x^2 + y^2)
```

ptiede: like negative negative? These aren't complex numbers so why would dy get conjugated?

ptiede: the differential of atan2(x, y) is

```
(ydx - xdy)/(x^2 + y^2)
```
which is what I implemented for both right?

ptiede:
Got it, I missed `Sub` in the IR. I thought the second line was a tuple of something.

PR #2075, opened 2026-02-12
feat: capi for transform dialect passes

avik-pal:
LLM generated needs to be reviewed


wsmoses (commented):
this will be a solid quality of life improvement

avik-pal:
> this will be a solid quality of life improvement

The only downside being it will be harder to disable passes from julia end. I am thinking of allowing a filter list that prevents certain passes from being added

src/enzyme_ad/jax/Integrations/c/EnzymeXLAOptPasses.cpp
```diff
@@ -0,0 +1,763 @@
...
+  list.push_back("dynamic_broadcast_in_dim_op_not_actually_dynamic<16>");
+  list.push_back("chained_dynamic_broadcast_in_dim_canonicalization<16>");
+  list.push_back("dynamic_broadcast_in_dim_all_dims_non_expanding<16>");
+  list.push_back("noop_reduce_op_canon<16>");
+  list.push_back("empty_reduce_op_canon<16>");
+  list.push_back("dynamic_reshape_op_canon<16>");
+  list.push_back("get_tuple_element_op_canon<16>");
+  list.push_back("real_op_canon<16>");
+  list.push_back("imag_op_canon<16>");
+  list.push_back("conj_complex_negate<16>");
```
mofeing: I just added a new transform pass, so leaving it here so that is not overwritten when this gets merged

```suggestion
  list.push_back("conj_complex_negate<16>");
  list.push_back("negate_imag_conj<16>");
```

avik-pal:
Tests should now pass

src/enzyme_ad/jax/Implementations/WhileLoopInfo.cpp
```diff
@@ -400,9 +400,10 @@ bool WhileLoopInfo::isConstantAcrossIterations(
     return true;
   }
 
-  if (auto blockArg = dyn_cast<BlockArgument>(v)) {
+  auto &body = op.getBody().front();
```
wsmoses, line 403: related?

src/enzyme_ad/jax/Passes/ConvertAllConstantsToSplattedConstantPass.cpp
```diff
@@ -30,7 +30,12 @@ struct ConvertSHLOConstantsToSplat
     Attribute firstVal = *elementIt;
 
     auto splatAttr = SplatElementsAttr::get(op.getType(), firstVal);
-    rewriter.replaceOpWithNewOp<stablehlo::ConstantOp>(op, splatAttr);
+    auto newConstant =
+        stablehlo::ConstantOp::create(rewriter, op.getLoc(), splatAttr);
+    // Insert a optimization barrier to prevent the constant from being folded
```
wsmoses: related?

wsmoses (approved):
can we seaparate out the pure api from the other changes, but otherwise lgtm

src/enzyme_ad/jax/Implementations/WhileLoopInfo.cpp
avik-pal: yes, needed for the jaxmd fix (essentially for nested while loop handling, without this we hit a dominance issue)

src/enzyme_ad/jax/Passes/ConvertAllConstantsToSplattedConstantPass.cpp
avik-pal: will split

PR #2078, opened 2026-02-13
add lit tests for pass ConvertParalleltoGPU2

xys-syx:
The lit tests is generated using 
`enzymexlamlir-opt convert-parallel-to-gpu1.mlir   --pass-pipeline="builtin.module(convert-parallel-to-gpu1,gpu-kernel-outlining,canonicalize)" &> convert-parallel-to-gpu2.mlir`

Could any one tell me if there is a better way to generate the lit tests for pass ConvertParalleltoGPU2

### DIFF
--- a/src/enzyme_ad/jax/Passes/ParallelLower.cpp
+++ b/src/enzyme_ad/jax/Passes/ParallelLower.cpp
@@ -239,47 +239,56 @@ prepareForGPUInline(LLVM::CallOp callOp, Operation *hostInsertionPoint,
   SmallVector<Value> newArgs;
   SmallVector<DictionaryAttr> newArgAttrs;
   std::optional<mlir::ArrayAttr> argAttrs = lfn.getArgAttrs();
-  assert(argAttrs);
   SmallVector<std::function<Value(OpBuilder &, Value)>> prepArg;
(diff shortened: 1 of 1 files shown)


test/lit_tests/lowering/convert-parallel-to-gpu2.mlir
```diff
@@ -0,0 +1,353 @@
+// RUN: enzymexlamlir-opt %s --pass-pipeline="builtin.module(convert-parallel-to-gpu2{backend=rocm})" | FileCheck %s
+
+#di_basic_type = #llvm.di_basic_type<tag = DW_TAG_base_type, name = "char", sizeInBits = 8, encoding = DW_ATE_signed_char>
+#di_file = #llvm.di_file<"../cu2.cu" in "/home/wmoses/git/Reactant/enzyme/build">
```
vimarsh6739: do we need to hardcode the path here?

xys-syx: We could use a fake path, as it is not checked, I just use the same way of lit tests for pass ConvertParallelToGPU1 which created by @wsmoses

ivanradanov (approved):

PR #2081, opened 2026-02-13
feat: simplify complex add/sub

### LINKED ISSUE #2080: Complex optimization opportunity
> Also, there is a optimization opportunity in
> 
> ```mlir
>     %6 = stablehlo.complex %cst_1, %4 {enzymexla.symmetric_matrix = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<2x2xcomplex<f64>>
>     %7 = stablehlo.complex %5, %cst {enzymexla.symmetric_matrix = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<2x2xcomplex<f64>>
>     %8 = stablehlo.add %6, %7 {enzymexla.symmetric_matrix = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<2x2xcomplex<f64>>
> ```
> 
> as the can be fused into
> 
> ```mlir
>     %8 = stablehlo.complex %5, %4 {enzymexla.symmetric_matrix = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<2x2xcomplex<f64>>
> ``` 

 _Originally posted by @mofeing in [#2071](https://github.com/EnzymeAD/Enzyme-JAX/issues/2071#issuecomment-3887620457)_

avik-pal:
fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/2080

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -33291,9 +33291,20 @@ struct BinaryOpComplexSimplifyBase
     auto lhsComplexOp = lhs.template getDefiningOp<stablehlo::ComplexOp>();
     auto rhsComplexOp = rhs.template getDefiningOp<stablehlo::ComplexOp>();
 
-    if (!lhsComplexOp && !rhsComplexOp &&
-        !isOnlyUsedInOperation(lhsComplexOp, op) &&
+    if (!lhsComplexOp || !rhsComplexOp ||
+        !isOnlyUsedInOperation(lhsComplexOp, op) ||
         !isOnlyUsedInOperation(rhsComplexOp, op)) {
+      return failure();
+    }
+
+    // If any of the components are zero, we can potentially do more aggressive
+    // constant folding
+    auto isZeroRealOrImag = [&](mlir::Value val) {
+      return guaranteedPurelyImagResult(val, rewriter) ||
+             guaranteedPurelyRealResult(val, rewriter);
+    };
+
+    if (!isZeroRealOrImag(lhsComplexOp) && !isZeroRealOrImag(rhsComplexOp)) {
       return failure();
     }
 
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -33277,6 +33277,35 @@ struct SplitComplexGather final
   }
 };
 
+// binop(complex, complex) -> complex(binop, binop)
+template <typename OpTy>
+struct BinaryOpComplexSimplifyBase
+    : public CheckedOpRewritePattern<OpTy, BinaryOpComplexSimplifyBase<OpTy>> {
+  using CheckedOpRewritePattern<
+      OpTy, BinaryOpComplexSimplifyBase<OpTy>>::CheckedOpRewritePattern;
+
+  LogicalResult matchAndRewriteImpl(OpTy op, PatternRewriter &rewriter) const {
+    auto lhs = op.getLhs();
+    auto rhs = op.getRhs();
+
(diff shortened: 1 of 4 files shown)


test/lit_tests/binopcomplexconstsimplify.mlir
```diff
@@ -0,0 +1,33 @@
...
+    %1 = stablehlo.complex %arg2, %arg3 : tensor<2x2xcomplex<f64>>
+    %2 = stablehlo.add %0, %1 : tensor<2x2xcomplex<f64>>
+    return %2 : tensor<2x2xcomplex<f64>>
+  }
+}
+
+// CHECK: func.func @main(%arg0: tensor<2x2xf64>, %arg1: tensor<2x2xf64>, %arg2: tensor<2x2xf64>, %arg3: tensor<2x2xf64>) -> tensor<2x2xcomplex<f64>> {
+// CHECK-NEXT:   %0 = stablehlo.add %arg0, %arg2 : tensor<2x2xf64>
+// CHECK-NEXT:   %1 = stablehlo.add %arg1, %arg3 : tensor<2x2xf64>
+// CHECK-NEXT:   %2 = stablehlo.complex %0, %1 : tensor<2x2xcomplex<f64>>
```
wsmoses: Is this actually beneficial here?

In principle doing the complex would enable the two adds to be vectorized together?

mofeing: this optimisation doesn't improve performance per se, but enables other optimisations. in particular, the context in which I suggested this optimisation was that I found sth like the following code:

```mlir
%cst = stablehlo.constant dense<[0]> : tensor<4xf64>
%cst_0 = stablehlo.constant dense<[0]> : tensor<4xf64>
%0 = stablehlo.complex %arg0, %cst : tensor<4xcomplex<f64>>
%1 = stablehlo.complex %cst_0, %arg1 : tensor<4xcomplex<f64>>
%2 = stablehlo.add %0, %1 : tensor<4xcomplex<f64>>
```

clearly, this code should be simplified to this code using constant folding:

```mlir
%2 = stablehlo.complex %arg0, %arg1 : tensor<4xcomplex<f64>>
```

but this wasn't happening, and this optimization enables.

@avik-pal maybe we should also put a test with the full example I put above to check that the optimisation works for the original target case?

wsmoses: In this case j think it's worth limiting this to the case which is more directly similar to that case, with the zeros

avik-pal: marking as draft for now. We can do a real/imag analysis (https://github.com/EnzymeAD/Enzyme-JAX/issues/1985) and propagate up if it is either purely imag or purely real

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -33277,6 +33277,46 @@ struct SplitComplexGather final
...
+    }
+
+    // If any of the components are zero, we can potentially do more aggressive
+    // constant folding
+    auto isZeroRealOrImag = [&](mlir::Value val) {
+      return guaranteedPurelyImagResult(val, rewriter) ||
+             guaranteedPurelyRealResult(val, rewriter);
+    };
+
+    if (!isZeroRealOrImag(lhsComplexOp) && !isZeroRealOrImag(rhsComplexOp)) {
```
wsmoses: to avoid an infinite loop [worst case where the elementwise complex binds more closely than the add simplify], can we actually special case the add here to fold it with the zero immediately?

avik-pal: added in https://github.com/EnzymeAD/Enzyme-JAX/pull/2081/changes/38fa35c8b29d0caacfc21380ac973ca917ab5665

wsmoses (approved):

PR #2082, opened 2026-02-13
Add MPI Waitall support

romanlee:
* Adds support for `MPI_Waitall`:`EnzymeXLA_Op` and `LowerEnzymeXLAMPIPass`
* Adds fix for `Isend`/`Irecv`/`Wait` lowering passes: use i32 instead of i64 for MPI Requests, and change the lit tests accordingly.

### DIFF
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
@@ -55,7 +55,8 @@ struct MPICommRankOpLowering
 
       // Generate the enzymexla_wrapper_MPI_Comm_rank LLVM function body
       std::string wrapperFunctionName = "enzymexla_wrapper_" + mpiFunctionName;
-      {
+
+      if (!moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(wrapperFunctionName)) {
         OpBuilder::InsertionGuard guard(rewriter);
         rewriter.setInsertionPointToStart(moduleOp.getBody());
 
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1274,7 +1274,19 @@ def MPIIrecvOp : EnzymeXLA_Op<"mpi.irecv", []> {
...
+  let arguments = (ins TensorOf<[I32]> : $request);
+  let assemblyFormat = "`(` operands `)` attr-dict `:` type(operands)";
+}
+
+def MPIWaitallOp : EnzymeXLA_Op<"mpi.waitall", []> {
+  let summary = "Equivalent to "
+                "`MPI_Waitall(&request, &status)`";
+
+  let arguments = (
+    ins TensorOf<[I32]> : $count,
```
mofeing, line 1286: mmm are you sure the number of requests is gonna be determined at runtime? I think we can now statically in general

src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1274,7 +1274,19 @@ def MPIIrecvOp : EnzymeXLA_Op<"mpi.irecv", []> {
 def MPIWaitOp : EnzymeXLA_Op<"mpi.wait", []> {
   let summary = "Equivalent to "
                 "`MPI_Wait(&request, &status)`";
-  let arguments = (ins TensorOf<[I64]> : $request);
+  let arguments = (ins TensorOf<[I32]> : $request);
```
mofeing, line 1277: IMO we should return the tensor that we wait for.

wsmoses: that may not be as easily known at this time, though I agree it would be nice if we could have it as a design.

wsmoses (approved):
I'm okay with this as is (and fine to land to start to experiment), but like @mofeing mentioned we should also start to consider some bigger design questions

PR #2096, opened 2026-02-16
Add OS condition to nightly release update step

wsmoses:


.github/workflows/build.yml
```diff
@@ -224,7 +224,7 @@ jobs:
           
       - name: Update Nightly Release
         uses: ncipollo/release-action@v1
-        if: ${{ !cancelled() && github.ref == 'refs/heads/main' }} 
+        if: ${{ !cancelled() && github.ref == 'refs/heads/main' && os == 'linux-x86-n2-32' }}
```
giordano, line 227: This is wrong again 🫠 Should have been
```suggestion
        if: ${{ !cancelled() && github.ref == 'refs/heads/main' && matrix.os == 'linux-x86-n2-32' }} 
```

wsmoses: oh bleh

PR #2106, opened 2026-02-18
fix rev diff rule of `stablehlo.complex`

mofeing:
we were missing a negation on the imaginary part

cc @ptiede

### DIFF
--- a/test/lit_tests/diffrules/stablehlo/complex.mlir
+++ b/test/lit_tests/diffrules/stablehlo/complex.mlir
@@ -19,8 +19,8 @@ func.func @main(%a : tensor<2xf32>, %b : tensor<2xf32>) -> tensor<2xcomplex<f32>
 // REVERSE-NEXT:    %1 = chlo.conj %0 : tensor<2xcomplex<f32>> -> tensor<2xcomplex<f32>>
 // REVERSE-NEXT:    %2 = stablehlo.real %1 : (tensor<2xcomplex<f32>>) -> tensor<2xf32>
 // REVERSE-NEXT:    %3 = stablehlo.add %cst_0, %2 : tensor<2xf32>
-// REVERSE-NEXT:    %4 = stablehlo.negate %1 : tensor<2xcomplex<f32>>
-// REVERSE-NEXT:    %5 = stablehlo.imag %4 : (tensor<2xcomplex<f32>>) -> tensor<2xf32>
+// REVERSE-NEXT:    %4 = stablehlo.imag %1 : (tensor<2xcomplex<f32>>) -> tensor<2xf32>
+// REVERSE-NEXT:    %5 = stablehlo.negate %4 : tensor<2xf32>
 // REVERSE-NEXT:    %6 = stablehlo.add %cst_0, %5 : tensor<2xf32>
 // REVERSE-NEXT:    return %3, %6 : tensor<2xf32>, tensor<2xf32>
 // REVERSE-NEXT:  }


src/enzyme_ad/jax/Implementations/HLODerivatives.td
```diff
@@ -1086,7 +1086,7 @@ def : HLOInactiveOp<"CompareOp">;
 
 def : HLODerivative<"ComplexOp", (Op $re, $im), [
   (Real (DiffeRet)),
-  (Imag (DiffeRet)),
+  (Imag (Neg (DiffeRet))),
```
wsmoses: Can you negate the imag instead of imag the negate ?

test/lit_tests/diffrules/stablehlo/complex.mlir
```diff
@@ -19,7 +19,8 @@ func.func @main(%a : tensor<2xf32>, %b : tensor<2xf32>) -> tensor<2xcomplex<f32>
 // REVERSE-NEXT:    %1 = chlo.conj %0 : tensor<2xcomplex<f32>> -> tensor<2xcomplex<f32>>
 // REVERSE-NEXT:    %2 = stablehlo.real %1 : (tensor<2xcomplex<f32>>) -> tensor<2xf32>
 // REVERSE-NEXT:    %3 = stablehlo.add %cst_0, %2 : tensor<2xf32>
-// REVERSE-NEXT:    %4 = stablehlo.imag %1 : (tensor<2xcomplex<f32>>) -> tensor<2xf32>
-// REVERSE-NEXT:    %5 = stablehlo.add %cst_0, %4 : tensor<2xf32>
-// REVERSE-NEXT:    return %3, %5 : tensor<2xf32>, tensor<2xf32>
+// REVERSE-NEXT:    %4 = stablehlo.negate %1 : tensor<2xcomplex<f32>>
+// REVERSE-NEXT:    %5 = stablehlo.imag %4 : (tensor<2xcomplex<f32>>) -> tensor<2xf32>
+// REVERSE-NEXT:    %6 = stablehlo.add %cst_0, %5 : tensor<2xf32>
```
wsmoses, line 24: Do we have an optimization such that negate of imag of conj of x just becomes imag of x?

src/enzyme_ad/jax/Implementations/HLODerivatives.td
mofeing: I tried that before but this way enabled more optimizations. take this code:

```mlir
module {
  func.func private @imag_exp_complex(%arg0: tensor<f64>) -> (tensor<f64>, tensor<f64>) {
    %c = stablehlo.constant dense<0> : tensor<i64>
    %0 = stablehlo.convert %c : (tensor<i64>) -> tensor<f64>
    %1 = stablehlo.complex %0, %arg0 : tensor<complex<f64>>
    %2 = stablehlo.exponential %1 : tensor<complex<f64>>
    %3 = stablehlo.imag %2 : (tensor<complex<f64>>) -> tensor<f64>
    return %3, %arg0 : tensor<f64>, tensor<f64>
  }

  func.func @main(%arg0: tensor<f64>) -> (tensor<f64>, tensor<f64>) {
    %x = stablehlo.constant dense<1.0471975511965976> : tensor<f64>

    %c = stablehlo.constant dense<0> : tensor<i64>
    %0 = stablehlo.convert %c : (tensor<i64>) -> tensor<f64>
    %cst = stablehlo.constant dense<1.000000e+00> : tensor<f64>

    %2:2 = enzyme.autodiff @imag_exp_complex(%arg0, %cst, %0) {activity = [#enzyme<activity enzyme_active>], ret_activity = [#enzyme<activity enzyme_activenoneed>, #enzyme<activity enzyme_active>]} : (tensor<f64>, tensor<f64>, tensor<f64>) -> (tensor<f64>, tensor<f64>)
    check.expect_almost_eq_const %2#1, dense<0.5> : tensor<f64>

    return %2#0, %2#1 : tensor<f64>, tensor<f64>
  }
}
```

if we leave it the way it's now, the `Neg` is actually optimized away.

## with `(Neg (Imag (DiffeRet))`

```mlir
module {
  func.func @main(%arg0: tensor<f64>) -> (tensor<f64>, tensor<f64>) {
    %cst = stablehlo.constant {enzymexla.complex_is_purely_real = [#enzymexla<guaranteed NOTGUARANTEED>]} dense<(0.000000e+00,-1.000000e+00)> : tensor<complex<f64>>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
    %0 = stablehlo.complex %cst_0, %arg0 {enzymexla.complex_is_purely_real = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<complex<f64>>
    %1 = stablehlo.exponential %0 : tensor<complex<f64>>
    %2 = stablehlo.multiply %cst, %1 {enzymexla.complex_is_purely_real = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<complex<f64>>
    %3 = stablehlo.imag %2 : (tensor<complex<f64>>) -> tensor<f64>
    %4 = stablehlo.negate %3 : tensor<f64>
    check.expect_almost_eq_const %4, dense<5.000000e-01> : tensor<f64>
    return %arg0, %4 : tensor<f64>, tensor<f64>
  }
}
```

## with `(Imag (Neg (DiffeRet))`

```mlir
module {
  func.func @main(%arg0: tensor<f64>) -> (tensor<f64>, tensor<f64>) {
    %cst = stablehlo.constant {enzymexla.complex_is_purely_real = [#enzymexla<guaranteed NOTGUARANTEED>]} dense<(-0.000000e+00,1.000000e+00)> : tensor<complex<f64>>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f64>
    %0 = stablehlo.complex %cst_0, %arg0 {enzymexla.complex_is_purely_real = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<complex<f64>>
    %1 = stablehlo.exponential %0 : tensor<complex<f64>>
    %2 = stablehlo.multiply %cst, %1 {enzymexla.complex_is_purely_real = [#enzymexla<guaranteed NOTGUARANTEED>]} : tensor<complex<f64>>
    %3 = stablehlo.imag %2 : (tensor<complex<f64>>) -> tensor<f64>
    check.expect_almost_eq_const %3, dense<5.000000e-01> : tensor<f64>
    return %arg0, %3 : tensor<f64>, tensor<f64>
  }
}
```

wsmoses: hm that definitely implies we're missing an optimization, since ideally both are equivalent. My preference for the other form is because as generated by AD without optimization we're only negating a real number vs we're negating the full complex [and the only user of the full complex is us].

In this case the reason the other is simpler is because the negate is first folded into the constant mul, but for bigger codes that won't be the case.

Here at least one optimization which is missing is that we're doing imag of (mul (a pure imag constant)) which can itself be simplified.

test/lit_tests/diffrules/stablehlo/complex.mlir
mofeing: I don't think so, but I can add it in another pr.

wsmoses: yeah thats fine

src/enzyme_ad/jax/Implementations/HLODerivatives.td
mofeing: > Here at least one optimization which is missing is that we're doing imag of (mul (a pure imag constant)) which can itself be simplified.

maybe is because `enzymexla.complex_is_purely_imaginary` analysis is missing?

test/lit_tests/diffrules/stablehlo/complex.mlir
ptiede: Great catch. I think we need some documentation about how to write reverse rules for complex operations and when to conjugate and when to not. Right now I'm pretty spooked and confused about exactly the conventions that Enzyme uses. It looks like for a C->C function Enzyme tries to handle the conjugations for you, while for R->C and C->R it looks like we are expected to deal with it ourselves.

PR #2116, opened 2026-02-19
[CI] Adapt GB-25 tests for running also on TPU, but disable it

giordano:


giordano:
I'm going to merge this PR with the job on TPU disabled, only keeping the other adaption.  Will enable the TPU in a follow up PR, so that we can track the issue with compilation there.

PR #2121, opened 2026-02-20
Multi-X custom call lowering

wsmoses:


.github/workflows/test-gb-25.yml
```diff
@@ -54,6 +54,7 @@ jobs:
           # - '0123456789abcdef0123456789abcdef01234567'
         reactant_commit:
           - 'main'
+          - 'multirotate_custom_call'
```
giordano, line 57: Was this supposed to be kept?

PR #2136, opened 2026-02-21
fix: build deps

avik-pal:


giordano (approved):
This is also being fixed in #2127, but I guess it'd be better to take out this atomic change to resolve the CI failure rather than waiting for #2127 to be merged.

PR #2157, opened 2026-02-25
RedirectSliceThroughDUS pattern to enable in-place DUS bufferization

arpitj1:
When B = dus(A, update, idx) and a later slice(A) reads from a region that does not overlap with B's update region, this pattern redirects the slice to read from B instead. Since B is identical to A outside its update region, this is semantics-preserving.

The pattern handles both stablehlo.slice (static indices) and stablehlo.dynamic_slice (constant indices), with same-block dominance checking.

### DIFF
--- a/.github/workflows/test-gb-25.yml
+++ b/.github/workflows/test-gb-25.yml
@@ -48,7 +48,7 @@ jobs:
           # - '0123456789abcdef0123456789abcdef01234567'
 
         reactant_commit:
-          - '9344280654cc37e587d27ab63ff77d09f1804d83' # main before breaking pr
+          - 'main'
           - 'ss/apply-redirect-dus'
 
     steps:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -12769,6 +12769,98 @@ struct DUSSliceSimplify final
...
+        continue;
+
+      // Check to see if B dominates slice and and is in same basic block
+      if (user->getBlock() != dusOp->getBlock() ||
+          !dusOp->isBeforeInBlock(user))
+        continue;
+
+      // Check range for sliceOp and DynamicSliceOp
+      // Two intervals [a, b) and [c, d) do not overlap if b <= c
+      // (slice ends before update starts) or a >= d (slice starts
```
wsmoses, line 12816: @mofeing can you amend this to re-use https://github.com/EnzymeAD/Enzyme-JAX/blob/3b0b19ffd1b49142564ed322c4b533beeb4f4375/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L17042

also in a follow up PR we should consider also handling updatewithoutcorners [not just DUS]

PR #2158, opened 2026-02-25
Add reduce const prop

### LINKED ISSUE #2141: missing reduce constant prop
```mlir
module @reactant_asum attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<32xf32> {enzymexla.memory_effects = []}) -> tensor<f32> attributes {enzymexla.memory_effects = []} {
    %cst = stablehlo.constant {enzymexla.non_negative = [#enzymexla<guaranteed GUARANTEED>]} dense<0.000000e+00> : tensor<3xf32>
    %cst_0 = stablehlo.constant dense<0.000000e+00> : tensor<f32>
    %0 = stablehlo.slice %arg0 [0:11:5] {enzymexla.non_negative = [#enzymexla<guaranteed NOTGUARANTEED>]} : (tensor<32xf32>) -> tensor<3xf32>
    %1 = stablehlo.abs %0 : tensor<3xf32>
    %2 = stablehlo.reduce(%1 init: %cst_0) applies stablehlo.add across dimensions = [0] : (tensor<3xf32>, tensor<f32>) -> tensor<f32>
    %3 = stablehlo.reduce(%cst init: %cst_0) applies stablehlo.add across dimensions = [0] : (tensor<3xf32>, tensor<f32>) -> tensor<f32>
    %4 = stablehlo.add %2, %3 : tensor<f32>
    return %4 : tensor<f32>
  }
}
```

Pangoraw:
Fixes #2141.

Missing the case where input is a known constant but initValue is not and it's an add/mul. But I think that's rare.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -6236,6 +6236,70 @@ struct GammaConstProp final
   }
 };
 
+struct ReduceConstProp final
+    : CheckedOpRewritePattern<stablehlo::ReduceOp, ReduceConstProp> {
+  using CheckedOpRewritePattern::CheckedOpRewritePattern;
+  size_t max_constant_expansion;
+
+  ReduceConstProp(size_t max_constant_expansion, MLIRContext *context,
+                  PatternBenefit benefit = 1,
+                  ArrayRef<StringRef> generatedNames = {})
+      : CheckedOpRewritePattern(context, benefit, generatedNames),
+        max_constant_expansion(max_constant_expansion) {}
+
+  LogicalResult matchAndRewriteImpl(stablehlo::ReduceOp op,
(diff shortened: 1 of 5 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -6236,6 +6236,70 @@ struct GammaConstProp final
...
+        size_t numEl = 1;
+        for (auto s : st.getShape())
+          numEl *= s;
+
+        if (numEl >= max_constant_expansion)
+          return failure();
+      }
+    }
+
+    SmallVector<stablehlo::Tensor> outputs = stablehlo::reduceOp(
```
wsmoses, line 6289: calling this is rather expensive [especially for large sized tensors], could we handle this directly?

PR #2167, opened 2026-03-01
move blas attrs, ops to new BLAS dialect

mofeing:
this PR moves ops and attributes defined in the EnzymeXLA dialect and related to BLAS, to a new BLAS dialect.

~to avoid breaking stuff, it doesn't remove already existing ops or attrs but marks them as deprecated.~

@avik-pal @snonk since you have some PRs on these ops I don't want to break, i can either import them once merged or rebase the PRs on top of this if it gets merged fast. whatever is best for you.

### DIFF
--- a/src/enzyme_ad/jax/BUILD
+++ b/src/enzyme_ad/jax/BUILD
@@ -1215,6 +1215,7 @@ cc_library(
         ":stablehlo-derivatives",
         ":triton-derivatives",
         "//src/enzyme_ad/jax/Dialect/BLAS:BlasDialectUtils",
+        "//src/enzyme_ad/jax/Dialect/BLAS:BlasDialect",
         "//src/external/isl:Isl",
         "//src/external/polly:Polly",
         "@com_google_absl//absl/status",


avik-pal:
I think we can move the files / ignore compatibility. Reactant won't break, we just need to update the pass pipeline with the next jll

src/enzyme_ad/jax/Passes/LowerEnzymeXLABLAS.cpp
avik-pal, line 1: ~Is there a reason to split up the lowering? At one point we can have a decomposition pass into shlo ops (we can call that `legalize-blas-to-shlo` instead of lowering, inline with CHLO), but currently both the lowering files effectively do the same thing for different backends.~

NVM I see that only the fallbacks were moved.

Can we change the naming to `lower-blas` and `legalize-blas-to-shlo` instead, since the lowering also emits custom_calls and not exclusively jitcalls

mofeing: Right, those names fit better. Changing them now.

About the split of the lowering pass, what I would like is to be able to force the pure-stablehlo fallback (e.g. for testing, for sharding, ...).

IMHO, it makes sense to have  `legalize-to-shlo` and `lower-blas` separate but I'm ok with reverting it back and just adding an option for it.

PR #2171, opened 2026-03-02
enable SliceBroadcast on multiple use of BroadcastInDimOp

mofeing:
`SliceBroadcast` is not taking effect on cases where `BroadcastInDimOp` has multiple users

in pre_xla dump we see:

```mlir
    %398 = stablehlo.broadcast_in_dim %389, dims = [1] : (tensor<1520xf32>) -> tensor<4x1520x3056xf32>
    ...
    %455 = stablehlo.slice %398 [0:4, 1:1520, 0:3056] : (tensor<4x1520x3056xf32>) -> tensor<4x1519x3056xf32>
    ...
    %623 = stablehlo.slice %398 [0:4, 2:1519, 0:3056]  : (tensor<4x1520x3056xf32>) -> tensor<4x1517x3056xf32>
```

when it should be:
```
%foo = stablehlo.slice %arg0 [1:1520] : (tensor<1520xf32>) -> tensor<1519xf32>
%455 = stablehlo.broadcast_in_dim %foo, dims = [1] : (tensor<1519xf32>) -> tensor<4x1519x3056xf32>
%bar = stablehlo.slice %arg0 [2:1519] : (tensor<1520xf32>) -> tensor<1517xf32>
%623 = stablehlo.broadcast_in_dim %bar, dims = [1] : (tensor<1517xf32>) -> tensor<4x1517x3056xf32>
```

this should reduce message size on collective slices.  also, `slice`s can then be grouped into a simpler `multi_slice`.


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -3099,8 +3099,11 @@ struct SliceBroadcast final
       outidx++;
     }
 
-    if (innerSlice && !llvm::hasSingleElement(bcast->getUsers()))
-      return failure();
+    // if (innerSlice && !llvm::hasSingleElement(bcast->getUsers())) {
```
wsmoses: I would make this case here more specialized, e.g. `if (innerSlice && !hasSingleElement && !allUsersAreBroadcasts) return failure()`

mofeing: I guess you mean sth like `allUsersAreSlices`, but this disables the pattern since some of the cases are like this (used by slice and by multiply):

```mlir
%398 = stablehlo.broadcast_in_dim %389, dims = [1] : (tensor<1520xf32>) -> tensor<4x1520x3056xf32>
...
%455 = stablehlo.slice %398 [0:4, 1:1520, 0:3056] : (tensor<4x1520x3056xf32>) -> tensor<4x1519x3056xf32>
...
%623 = stablehlo.slice %398 [0:4, 2:1519, 0:3056]  : (tensor<4x1520x3056xf32>) -> tensor<4x1517x3056xf32>
...
%2289 = stablehlo.multiply %398, %2277 : tensor<4x1520x3056xf32>
...
%3338 = stablehlo.multiply %398, %3337 : tensor<4x1520x3056xf32>
```

mofeing: the result of this PR on the example above is

```mlir
%398 = stablehlo.broadcast_in_dim %389, dims = [1] : (tensor<1520xf32>) -> tensor<4x1520x3056xf32>
...
%foo = stablehlo.slice %389 [1:1520] : (tensor<1520xf32>) -> tensor<1519xf32>
%455 = stablehlo.broadcast_in_dim %foo, dims = [1] : (tensor<1519xf32>) -> tensor<4x1519x3056xf32>
...
%bar = stablehlo.slice %389 [2:1519] : (tensor<1520xf32>) -> tensor<1517xf32>
%623 = stablehlo.broadcast_in_dim %bar, dims = [1] : (tensor<1517xf32>) -> tensor<4x1517x3056xf32>
...
%2289 = stablehlo.multiply %398, %2277 : tensor<4x1520x3056xf32>
%3338 = stablehlo.multiply %398, %3337 : tensor<4x1520x3056xf32>
```

the `multiply`s still use the original `broadcast_in_dim` but the `slice`s work on a way smaller tensor, so we should see smaller communication for them without a degradation in other parts?

wsmoses: so it's not always beneficial to do this since it will increase the number of operations. In the case you have here, we can't get rid of the original slice so this will still end up doing the old communication adn the new slice ones, so it will add to comm cost

mofeing: I see. This is also breaking autobatch for some reason I don't understand.
Let's see if adding the `allUsersAreSlices` constraint manages to do sth.

PR #2183, opened 2026-03-05
skip lowering of wrap to custom_call on assertion error

mofeing:
it should avoid the segfault we run into on 12 GPUs

```
E0000 00:00:1772658857.820566  248565 status_macros.cc:58] INTERNAL: RET_CHECK failure (external/xla/xla/service/spmd/custom_call_handler.cc:662) right_amount < full_pre_wrap_size % participating_shards Right amount must be entirely within final shard
```

this is not a complete fix, as the code won't end up as optimized, but it should work

### DIFF
--- a/src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
+++ b/src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
@@ -2370,10 +2370,14 @@ struct WrapCustomCallOptimize : public OpRewritePattern<enzymexla::WrapOp> {
         full_pre_wrap_size % participating_shards == 0;
 
     if (divisible_by_participating_shards) {
-      if (rightAmount >= shard_size)
+      // NOTE should be `>=` or else XLA will assert error, but currently it's
+      // probably generating too many comms
+      if (rightAmount > shard_size)
         return failure();
     } else {
-      if (rightAmount >= full_pre_wrap_size % shard_size)
+      // NOTE should be `>=` or else XLA will assert error, but currently it's
+      // probably generating too many comms
+      if (rightAmount > full_pre_wrap_size % shard_size)
         return failure();
     }
 


src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
```diff
@@ -2355,6 +2355,16 @@ struct WrapCustomCallOptimize : public OpRewritePattern<enzymexla::WrapOp> {
     auto leftAmount = wrap.getLhs();
     auto rightAmount = wrap.getRhs();
 
+    // avoids assertion error in
+    // `external/xla/xla/service/spmd/custom_call_handler.cc:662` and leaves
+    // this op to be dealt in a later pass
+    // TODO remove this assert once we deal with the problem
+    auto pre_wrap_shape = wrap.getOperand().getType().getShape();
+    auto full_pre_wrap_size = pre_wrap_shape[rotateDimension];
+    auto shard_size = rotateSharding.getTileSizes()[rotateDimension];
```
wsmoses: this isn't quite right, you need to have the two different cases, the first case being if it's evenly divisible

mofeing: right, missed that. fixing it now

src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
```diff
@@ -2355,6 +2355,28 @@ struct WrapCustomCallOptimize : public OpRewritePattern<enzymexla::WrapOp> {
...
+    auto shard_size =
+        rotateSharding.getDimSharding(rotateDimension).getShardedSize(mesh);
+    int64_t participating_shards =
+        (full_pre_wrap_size + shard_size - 1) / shard_size; // CeilOfRatio
+    bool divisible_by_participating_shards =
+        full_pre_wrap_size % participating_shards == 0;
+
+    if (divisible_by_participating_shards) {
+      if (rightAmount < shard_size)
+        return failure();
```
wsmoses, line 2376: this condition is wrong, it should be if right amount > shard_size. and the same below

mofeing: right, because the xla code is a check

wsmoses: if it's equal it should honestly be fine

mofeing: I think so, but we would run into the assert error anyway because the XLA check states https://github.com/openxla/xla/blob/079790f386850f613a47f78b590ed2c2df20fdc3/xla/service/spmd/custom_call_handler.cc#L656-L663

```c++
  if (divisible_by_participating_shards) {
    TF_RET_CHECK(right_amount < shard_size)
        << "Right amount must be entirely within a shard";
  } else {
    TF_RET_CHECK(right_amount < full_pre_wrap_size % participating_shards)
        << "Right amount must be entirely within final shard";
  }
```

so it will assert error if it's equal

wsmoses: Fair enough, but we can concurrently fix that upstream (not a high priority though)

PR #2201, opened 2026-03-10
add helper script for converting HLO to GraphViz's DOT format

mofeing:
xprof's Graph Viewer isn't always available, and when it is, it doesn't work many times (at least for me). this script automates the graph plot using `hlo-translate` and `enzymexlamlir-opt`.

```sh
$ ./scripts/hlo-graphviz.sh -f 'private @fused_computation.720.clone.clone' ../profiles/tpu/2026-03-08T22-20-42.707.57974/simulation-xla-dump-linux-x86-ct6e-180-4tpu-1/xla_dump_run/module_0965.reactant_loop_.cl_878201828.after_optimizations_before_buffer_assignment.txt
digraph G {
  compound = true;
  subgraph cluster_1 {
    v2 [label = " ", shape = plain];
    label = "builtin.module : ()\l";
    subgraph cluster_3 {
      v4 [label = " ", shape = plain];
      label = "";
      subgraph cluster_5 {
        v6 [label = " ", shape = plain];
...
```

there are ways to improve this script, but it works well enough for now.

### DIFF
--- a/scripts/hlo-graphviz.sh
+++ b/scripts/hlo-graphviz.sh
@@ -30,11 +30,12 @@ fi
 
 bazel_run() {
     bazel run --action_env=CC=${CC-clang}\
-    --define using_clang=true\
+    --define=using_clang=true\
     --run_under="cd $PWD &&"\
     --ui_event_filters=ERROR\
     --noshow_progress\
     --noshow_loading_progress\
+    --color=no\
     $@
 }
 
@@ -54,7 +55,7 @@ fi
 
 # convert MLIR to graphviz dot format
 TMP_MLIR_OPT_CMD=$(mktemp)
-mlir_opt --view-op-graph $TMP_AWK 2>&1 >/dev/null | awk '{gsub(/\x1b\[[0-9;]*m/, ""); print}' >$TMP_MLIR_OPT_CMD
+mlir_opt --view-op-graph $TMP_AWK 2>$TMP_MLIR_OPT_CMD >/dev/null
 
 if [ -n "$OUTPUT" ]; then
     cp --interactive $TMP_MLIR_OPT_CMD $OUTPUT


scripts/hlo-graphviz.sh
```diff
@@ -0,0 +1,77 @@
...
+    esac
+done
+
+shift $((OPTIND - 1))
+INPUT="$1"
+if [ -z "$INPUT" ]; then
+    usage
+fi
+
+HLO_TRANSLATE_CMD=./$(bazel cquery --output=starlark --starlark:expr=target.files_to_run.executable.path @xla//xla/hlo/tools:hlo-translate 2>/dev/null)
```
wsmoses: there's also `bazel run @xla//xla/hlo/tools:hlo-translate -- args...`, which may be easier here

mofeing: I was using that in the beginning but it was giving me problems in hydra because...
- `hlo-translate` requires to be compiled with clang
- `enzymexlamlir-opt` build fails with clang, but works with gcc

I was setting `export CC=clang` inside the script to force using clang inside `bazel` but `enzymexlamlir-opt` would fail. If I only set `CC=clang` for `hlo-translate`, then it would recompile the commands on each call. 😕

mofeing: maybe @giordano has a better idea here on how to properly set the compiler in bazel?

Pangoraw: enzymexlamlir-opt should work with clang (`--define using_clang=true`)

mofeing: that flag is not enough to force to compile it with clang, as it compiles it with gcc and fails with
```
ERROR: /mnt2/mofeing/Enzyme-JAX/src/enzyme_ad/jax/BUILD:213:11: Compiling src/enzyme_ad/jax/TransformOps/RaisingTransformOps.cpp failed: (Exit 1): gcc failed: error executing CppCompile command (from target //src/enzyme_ad/jax:RaisingTransformOps)
```

I guess I need to combine it with `--action_env=CC=clang`?

giordano: `--define using_clang=true` doesn't force anything, it's informing Bazel that Clang is being used (ideally this should be inferred automatically, I don't know if anyone knows how to do that in Bazel)

> I guess I need to combine it with `--action_env=CC=clang`?

Yes, exporting `CC` would work too, but should also accept `CC` set from outside, I don't want to use `clang` on `hydra` for example, I want to use my own compiler wrapper.

mofeing: `hlo-translate` needs clang but if I put nothing in `bazel run //:enzymexlamlir-opt -- ...`, it will recompile as it will default to gcc and `hlo-translate` would have invalidated some build deps. I will change it but it can invalidate builds.

scripts/hlo-graphviz.sh
```diff
@@ -0,0 +1,70 @@
...
+done
+
+shift $((OPTIND - 1))
+INPUT="$1"
+if [ -z "$INPUT" ]; then
+    usage
+fi
+
+hlo_translate() {
+    bazel run --action_env=CC=clang --define using_clang=true --run_under="cd $PWD &&" @xla//xla/hlo/tools:hlo-translate -- $@
```
giordano:
```suggestion
    bazel run --action_env=CC=${CC-clang} --define using_clang=true --run_under="cd $PWD &&" @xla//xla/hlo/tools:hlo-translate -- $@
```

Issue #2104, opened 2026-02-17
Add kernel cannot be raised with with certain size-workgroup combos (0.2.220)


dkytezab:
```julia
using Reactant, KernelAbstractions, CUDA

Reactant.set_default_backend("cpu")

const ReactantBackend = Base.get_extension(Reactant, :ReactantKernelAbstractionsExt).ReactantBackend

@kernel function _add_one!(out, inp)
    i, j = @index(Global, NTuple)
    @inbounds out[i, j] = inp[i, j] + 1.0
end

function go!(out, inp)
    _add_one!(ReactantBackend(), wg)(out, inp; ndrange=size)
    return nothing
end

size = (4, 4)
wg = (16, 16)

out = Reactant.to_rarray(zeros(size...))
inp = Reactant.to_rarray(ones(size...))

compiled! = Reactant.@compile raise=true sync=true go!(out, inp)
compiled!(out, inp)
```
This breaks with 
```julia
loc(fused<#llvm.di_subprogram<id = distinct[0]<>, compileUnit = <id = distinct[1]<>, sourceLanguage = DW_LANG_Julia, file = <"julia" in ".">, producer = "julia", isOptimized = true, emissionKind = None, nameTableKind = None>, name = "gpu__add_one!", linkageName = "julia_gpu__add_one!_36086", file = <"none" in ".">, subprogramFlags = "Definition|Optimized", type = <>>>["none":0:0]): error: cannot raise if yet (non-pure or yielded values): 
"affine.if"(%arg3, %arg4) <{condition = affine_set<(d0, d1) : (-d0 - d1 * 4 + 3 >= 0)>}> ({
  %1 = "affine.load"(%arg1, %arg2, %arg3) <{map = affine_map<(d0, d1) -> (d0, d1)>}> : (memref<4x4xf64, 1>, index, index) -> f64
  %2 = "arith.addf"(%1, %0) <{fastmath = #arith.fastmath<none>}> {fastmathFlags = #llvm.fastmath<none>} : (f64, f64) -> f64
  "affine.store"(%2, %arg0, %arg2, %arg3) <{map = affine_map<(d0, d1) -> (d0, d1)>}> : (f64, memref<4x4xf64, 1>, index, index) -> ()
  "affine.yield"() : () -> ()
}, {
}) : (index, index) -> ()
[...]

PR #2203, opened 2026-03-11
generalized if support

wsmoses:
attempting to start off the vibe code if support

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -2257,8 +2257,7 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
       for (auto dim : storeValueMap.getOperands()) {
         // This dim is present in the masked update and not in the stored
         // dimensions.
-        if (std::find(storeOp.getIndices().begin(), storeOp.getIndices().end(),
-                      dim) == storeOp.getIndices().end()) {
+        if (!llvm::is_contained(storeOp.getIndices(), dim)) {
           auto err = op->emitError(
                          "masked affine.store is dependent on less dimensions "
                          "than masked stored value:\n")


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1727,7 +1809,7 @@ static LogicalResult tryRaisingLockStepForOpToStableHLO(
   if (pc.options.dump_failed_lockstep) {
     llvm::errs() << " failed lockstep of for raise: " << *forOp << "\n";
   }
-  return forOp.emitError("Not lockstep executable") << *forOp;
+  return failure();
```
wsmoses: probably was copilot's doing, but we should keep the better error message

wsmoses:
also might as well check/add https://github.com/EnzymeAD/Enzyme-JAX/issues/2104 as a test

cc @dkytezab

Pangoraw:
So mwe from https://github.com/EnzymeAD/Enzyme-JAX/issues/2104 would need additional simplification from AffineCFG:

```mlir
  %cst = arith.constant 1.000000e+00 : f64
  affine.parallel (%arg2, %arg3, %arg4) = (0, 0, 0) to (4, 4, 4) {
    affine.if affine_set<(d0, d1) : (-d0 - d1 * 4 + 3 >= 0)>(%arg3, %arg4) {
      %0 = affine.load %arg1[%arg2, %arg3] : memref<4x4xf64, 1>
      %1 = arith.addf %0, %cst {fastmathFlags = #llvm.fastmath<none>} : f64
      affine.store %1, %arg0[%arg2, %arg3] : memref<4x4xf64, 1>
    }
  }
  return
}
```

i.e. it make a reduction with respect to `%arg4` but the `affine.if` makes it so that it only runs the code for `%arg4 == 0`. So it should be able to simplify:

```mlir
  %cst = arith.constant 1.000000e+00 : f64
  affine.parallel (%arg2, %arg3) = (0, 0) to (4, 4) {
    affine.if affine_set<(d0) : (-d0 + 3 >= 0)>(%arg3) {
      %0 = affine.load %arg1[%arg2, %arg3] : memref<4x4xf64, 1>
      %1 = arith.addf %0, %cst {fastmathFlags = #llvm.fastmath<none>} : f64
      affine.store %1, %arg0[%arg2, %arg3] : memref<4x4xf64, 1>
    }
  }
  return
}

wsmoses:
we should definitely ideally fix the IR first [as the better/faster route], but shouldn't we be able to get affine.if generally raised here too

PR #2211, opened 2026-03-11
comment out isl warning

Pangoraw:
to preserve our log files


wsmoses:
cc @ivanradanov

ivanradanov:
What is the problem with aborting?

PR #2215, opened 2026-03-12
refactor: move the xla_ffi registrations from reactant to enzymejax

avik-pal:


wsmoses (commented):
+1 in princple for a number of reasons

wsmoses (approved):

PR #2263, opened 2026-03-17
Handle dynamic dim in ReshapeOpCreate helper

Pangoraw:
Arguably, the tested optimization should be disabled in this case, but at least it does not crash.

### DIFF
--- a/test/lit_tests/reshape_dyn.mlir
+++ b/test/lit_tests/reshape_dyn.mlir
@@ -1,4 +1,4 @@
-// RUN: enzymexlamlir-opt %s --enzyme-hlo-opt=passses=131072 | FileCheck %s
+// RUN: enzymexlamlir-opt %s  '--enzyme-hlo-generate-td=patterns=reshape_dynamic_slice(1)' --transform-interpreter --enzyme-hlo-remove-transform --canonicalize | FileCheck %s
 
 module {
   func.func @main(%arg0: tensor<?x1x1x128xf64>) -> tensor<1x1x128xf64> {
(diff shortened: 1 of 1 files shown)


test/lit_tests/reshape_dyn.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: enzymexlamlir-opt %s --enzyme-hlo-opt=passses=131072 | FileCheck %s
```
wsmoses: can you add the pattern by name for ease?

test/lit_tests/reshape_dyn.mlir
```diff
@@ -0,0 +1,21 @@
...
+  }
+}
+
+// CHECK:  func.func @main(%arg0: tensor<?x1x1x128xf64>) -> tensor<1x1x128xf64> {
+// CHECK-NEXT:    %c = stablehlo.constant dense<[1, 128]> : tensor<2xi32>
+// CHECK-NEXT:    %c_0 = stablehlo.constant dense<0> : tensor<i64>
+// CHECK-NEXT:    %0 = stablehlo.get_dimension_size %arg0, dim = 0 : (tensor<?x1x1x128xf64>) -> tensor<i32>
+// CHECK-NEXT:    %1 = stablehlo.reshape %0 : (tensor<i32>) -> tensor<1xi32>
+// CHECK-NEXT:    %2 = stablehlo.concatenate %1, %c, dim = 0 : (tensor<1xi32>, tensor<2xi32>) -> tensor<3xi32>
+// CHECK-NEXT:    %3 = stablehlo.dynamic_reshape %arg0, %2 : (tensor<?x1x1x128xf64>, tensor<3xi32>) -> tensor<?x1x128xf64>
```
wsmoses, line 19: concurrently in this case we should be able to, presumably change this to a static reshape [since its just the outermost dim which is ?]. Need to understand the spec better to decide if that is actually the case

Pangoraw: ReshapeOp currently expects a static shape output: https://github.com/openxla/stablehlo/blob/08ed068c4082c1309a51d9570bd2d96ddb142fa2/stablehlo/dialect/StablehloOps.td#L2883

which is what prompted this PR

PR #2276, opened 2026-03-19
FFT gradient fix

wsmoses:
@mofeing also trying to vibe code a numeric test, if it fails [or succeeds] you can take over the branch


wsmoses:
interpreter doesn't support fft so this is the way it is for now, go ahead and approve/merge if you think its correct

mofeing:
it does now :) https://github.com/openxla/stablehlo/pull/2899

I'm approving because this should be correct, and I can add the numerical tests in a new PR tomorrow

mofeing (approved):

PR #2294, opened 2026-03-21
[CI] Disable more HLO passes in GB-25 test

giordano:
From https://github.com/PRONTOLab/GB-25/blob/249114714d1bba59dc38aa5d1423960244ba5ec4/sharding/common_submission_generator.jl#L115.


giordano:
Doesn't seem to change much in terms of performance, but we might as well disable passes which caused problems in the past?

In any case this PR also resolves instantiation of the julia environment, which would be good right now that the PkgServer is sad.

PR #2295, opened 2026-03-21
support llvm.alloca for ParallelizeBlockOps

xys-syx:
This PR trying to support LLVM allocaOp in the ParallizeBlockOps
```
scf.parallel {
   A()
   scf.parallel {
    B()
   }
   llvm.alloca ...
   C()
}
```

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -1474,6 +1474,7 @@ struct MoveWhileToFor : public OpRewritePattern<WhileOp> {
         legal = true;
         assert(helper.lb);
         assert(helper.ub);
+        break;
       }
       if (!legal) {
         return rewriter.notifyMatchFailure(loop,
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -1446,6 +1446,16 @@ struct MoveWhileToFor : public OpRewritePattern<WhileOp> {
 
   LogicalResult matchAndRewrite(WhileOp loop,
                                 PatternRewriter &rewriter) const override {
+    if (getenv("DEBUG_WHILE_TO_FOR")) {
+      llvm::errs() << "MoveWhileToFor: trying to match WhileOp at ";
+      loop.getLoc().print(llvm::errs());
+      llvm::errs() << "\n  inits=" << loop.getInits().size()
+                    << " results=" << loop.getResults().size()
+                    << " before_args=" << loop.getBefore().front().getNumArguments()
+                    << " after_args=" << loop.getAfter().front().getNumArguments()
+                    << " before_ops=" << loop.getBefore().front().getOperations().size()
+                    << "\n";
+    }
     auto condOp = loop.getConditionOp();
     SmallVector<Value, 2> results = {condOp.getArgs()};
     WhileToForHelper helper(loop,
@@ -1501,10 +1511,12 @@ struct MoveWhileToFor : public OpRewritePattern<WhileOp> {
 
         if (auto trueBlockArg =
                 dyn_cast<BlockArgument>(selectOp.getTrueValue())) {
-          newArg = loop.getOperand(trueBlockArg.getArgNumber());
+          if (trueBlockArg.getOwner() == &condOp->getParentRegion()->front())
+            newArg = loop.getOperand(trueBlockArg.getArgNumber());
         } else if (auto falseBlockArg =
                        dyn_cast<BlockArgument>(selectOp.getFalseValue())) {
-          newArg = loop.getOperand(falseBlockArg.getArgNumber());
+          if (falseBlockArg.getOwner() == &condOp->getParentRegion()->front())
+            newArg = loop.getOperand(falseBlockArg.getArgNumber());
         }
       }
       if (!newArg)
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -1446,16 +1446,6 @@ struct MoveWhileToFor : public OpRewritePattern<WhileOp> {
 
   LogicalResult matchAndRewrite(WhileOp loop,
                                 PatternRewriter &rewriter) const override {
-    if (getenv("DEBUG_WHILE_TO_FOR")) {
-      llvm::errs() << "MoveWhileToFor: trying to match WhileOp at ";
-      loop.getLoc().print(llvm::errs());
-      llvm::errs() << "\n  inits=" << loop.getInits().size()
-                    << " results=" << loop.getResults().size()
-                    << " before_args=" << loop.getBefore().front().getNumArguments()
-                    << " after_args=" << loop.getAfter().front().getNumArguments()
-                    << " before_ops=" << loop.getBefore().front().getOperations().size()
-                    << "\n";
-    }
     auto condOp = loop.getConditionOp();
     SmallVector<Value, 2> results = {condOp.getArgs()};
     WhileToForHelper helper(loop,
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
```diff
@@ -1941,9 +1967,11 @@ struct ConvertParallelToGPU1Pass
       GreedyRewriteConfig config;
       config.enableFolding();
       if (failed(applyPatternsGreedily(m, std::move(patterns), config))) {
+        llvm::errs() << "=== InnerParallelSerialization: FAILED ===\n";
```
wsmoses: clean up the debug prints

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
```diff
@@ -1019,6 +1033,8 @@ struct ParallelizeBlockOps : public OpRewritePattern<scf::ParallelOp> {
     {
       auto zeroindex = arith::ConstantIndexOp::create(rewriter, loc, 0);
       rewriter.setInsertionPoint(innerBlock->getTerminator());
+      mlir::enzymexla::BarrierOp::create(rewriter, loc,
```
wsmoses, line 1061: is this needed?

test/lit_tests/lowering/convert-parallel-to-gpu1-llvm-alloca.mlir
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+// CHECK:  llvm.func local_unnamed_addr @test_parallelize_block_ops_llvm_alloca(%arg0: !llvm.ptr {llvm.noundef}) -> (i32 {llvm.noundef}) attributes {dso_local, passthrough = ["mustprogress"]} {
+// CHECK:    gpu.launch blocks(%arg1, %arg2, %arg3) in (%arg7 = %c4, %arg8 = %c1, %arg9 = %c1) threads(%arg4, %arg5, %arg6) in (%arg10 = %c8, %arg11 = %c1, %arg12 = %c1) {
+// CHECK:      %[[ALLOCA:.*]] = llvm.alloca %{{.*}} x !llvm.struct<"struct.Result", (f64)>
+// CHECK:      %[[LOAD_A:.*]] = memref.load %{{.*}}[%block_id_x]
+// CHECK:      %[[LOAD_B:.*]] = memref.load %{{.*}}[%thread_id_x]
+// CHECK:      arith.addf
+// CHECK:      memref.store
```
wsmoses: use the full IR output here for the checks

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
xys-syx: Ok, I will do that.

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
xys-syx: I think it may be better that we add a barrier for C' , according to lit tests, without it:
```
memref.store %5, %0[%thread_id_x]
%6 = arith.cmpi eq, %thread_id_x
scf.if %6 {
```
with it:
```
memref.store %5, %0[%thread_id_x]
gpu.barrier
%6 = arith.cmpi eq, %thread_id_x 
scf.if %6 {
```

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1458,6 +1458,7 @@ struct MoveWhileToFor : public OpRewritePattern<WhileOp> {
         legal = true;
         assert(helper.lb);
         assert(helper.ub);
+        break;
```
wsmoses, line 1477: can we write a separate pr and test for this?

this seems to have more general implications than the gpu stuff on own

xys-syx: Ok

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
xys-syx: Hello, I added barrier because of this line. https://github.com/EnzymeAD/Enzyme-JAX/blob/96e404b056bad1eb46b8551f3a54846886a2d529/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp#L926

xys-syx: Could you help check again?

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
```diff
@@ -1041,8 +1044,13 @@ struct ParallelizeBlockOps : public OpRewritePattern<scf::ParallelOp> {
           llvm_unreachable("Unhandled case");
           break;
         } else if (auto alloca = dyn_cast<LLVM::AllocaOp>(&op)) {
-          llvm_unreachable("Unhandled case");
-          break;
+          mlir::OpBuilder::InsertionGuard guard(rewriter);
+          rewriter.setInsertionPointToStart(innerBlock);
```
vimarsh6739: I think the insertion point is wrong. This will currently blow up stack allocation. 

If you have code like this:

```mlir
scf.parallel (%i) {
  scf.parallel(%j) { B(); }
  %tmp = llvm.alloca 4096 x i8
  func.call @foo(%tmp)
}
```

the current rewrite makes this

```mlir
scf.parallel (%i) {
  scf.parallel (%j) {
    %tmp = llvm.alloca 4096 x i8
    enzymexla.barrier
    scf.if (%j == 0) {
        func.call @foo(%tmp)
    }
  }
}
```

The correct code transform has to be(ensures threadlocal alloc)

```mlir
scf.parallel (%i) {
  scf.parallel (%j) {
    enzymexla.barrier
    scf.if (%j == 0) {
        %tmp = llvm.alloca 4096 x i8
        func.call @foo(%tmp)
    }
  }
}
```

src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
vimarsh6739: Ok, when you think of this in terms of happens-before relations, it makes sense to put up a barrier. 

For 
```mlir
scf.parallel(%outer) {
A()
scf.parallel(%i) { B()}
C()
}
```

The happens-before relations here are 

$$\forall_{inner}  HB\left( A \left[ outer \right] , B\left[ outer,inner \right]  \right)  \land \forall_{inner} HB \left(B \left[ outer,inner \right] , C\left[ outer \right]  \right)$$

If we miss a barrier before C, the happens-before relations on C only hold for %inner == 0 :

$$\forall_{inner}  HB\left( A \left[ outer \right] , B\left[ outer,inner \right]  \right)  \land \forall_{inner} HB \left(B \left[ outer,\mathbf{0} \right] , C\left[ outer \right]  \right)$$

So, we do need the barrier here. As for now, i'd add a test specifically for this case. I think something like this should work(its not valid mlir, but a stub). 

```mlir
scf.parallel %i = 0 to 2 {
   memref.load %x, buffer[%i]
   scf.parallel %j = 0 to 4 {
       memref.store to buffer[4 * %i + %j]
   }
    %y = memref.load buffer[(4 * %i + 3)] // load last elements
     memref.store %out[%i], (%x + %y)
}
```

This may also be a nice opportunity to add a `MustHappenBefore` analysis for these passes here (@wsmoses)

vimarsh6739: @EganJ

xys-syx: while I discover here 
https://github.com/EnzymeAD/Enzyme-JAX/blob/6cc2950f8dce3de34555762bae8299d1d4e3f8b2/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp#L966
which creates barrier between `scf.if { A_writes }` and B, but getif() only uses when before section has writes operation, while B writes the memory which A is reading, it could causes issues, for example,
```
scf.parallel (%i) {
  %v = memref.load %buf[%i]       // A: read
  scf.parallel (%j) {
    memref.store %x, %buf[%j]     // B: write
  }
}
```
What I am thinking is also add barrier explicitly between A' and B, @vimarsh6739 could you please tell your suggestions?

vimarsh6739: My understanding is that the barrier is correct here(it'll ensure that the load will happen before the store. 

I'm not following what's wrong with the barrier, it is in the right place (we ensure that every outer loop iteration of A(memref.load) happens before a corresponding memref.write in B for the same outer loop iteration).

If we're worried about B writing to a location which is read by A, that is also possible in the initial code (B in some outer iteration i2 can write to A in outer iteration i1 in parallel)

xys-syx: @vimarsh6739 Could we merge this first?

vimarsh6739 (approved):

vimarsh6739:
@codex review

PR #2296, opened 2026-03-21
Add comm pipelining

wsmoses:


patches/xla_gpu_pipeline_comms.patch
giordano, line 1: Is this upstreamed?

wsmoses: https://github.com/openxla/xla/pull/39619

PR #2303, opened 2026-03-23
Cleanup rotate spmd lowering

wsmoses:


src/enzyme_ad/jax/Passes/OptimizeCommunication.cpp
```diff
@@ -2183,13 +2182,14 @@ struct RotateSpmdOptimize : public OpRewritePattern<enzymexla::RotateOp> {
           /*result_layouts=*/nullptr,
           /*output_operand_aliases=*/nullptr);
 
-      SmallVector<sdy::TensorShardingAttr> newShardings(rotate.getAmount() + 1,
+      int32_t rotateAmount = rotate.getAmount();
```
Pangoraw: i think `rotateAmount` is already defined above but as uint32_t
```suggestion
```

PR #2307, opened 2026-03-23
add pad for alignment pass

wsmoses:


mofeing:
Yes, the handler of pad op doesn't yet work for the pads founds in GB-25; i.e. pads with both low and high paddings.

Working on it.

PR #2330, opened 2026-03-26
Add kernel float casting pass

Pangoraw:


src/enzyme_ad/jax/Passes/KernelCastPass.cpp
```diff
@@ -0,0 +1,164 @@
...
+  using KernelCastPassBase::KernelCastPassBase;
+
+  void runOnOperation() override {
+    func::FuncOp func = getOperation();
+
+    auto attr = func->getAttrOfType<StringAttr>("enzymexla.float_type");
+    if (!attr)
+      return;
+
+    Type targetFloat = parseFloatTypeAttr(attr.getValue(), func.getContext());
```
wsmoses: I think there's an mlir type attribute? so we don't need to do this via string

src/enzyme_ad/jax/Passes/LLVMToMemrefAccess.cpp
```diff
@@ -89,6 +89,20 @@ struct LLVMToMemrefAccessPass
             if (auto rtt = dyn_cast<RankedTensorType>(callerArgTy)) {
               auto thisShape = rtt.getShape();
               auto thisElTy = rtt.getElementType();
+
+              if (isa<FloatType>(thisElTy)) {
```
wsmoses, line 93: why do we need to modify here?

Pangoraw: On one side we have the kernel call taking `tensor<...xbf16>` so thisElTy would be bf16. On the callee side, there is `!llvm.ptr` which has no eltype, this basically prevents the wrapper to take `memref<...xbf16>` but `memref<...xf32>` instead which the kernel was compiled to be.

wsmoses: got it

PR #2342, opened 2026-03-28
Add name(id) for ISL params and sets

xys-syx:
Fix `src/external/isl/isl_space.c:3350: unexpected unnamed parameters`

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -207,6 +207,14 @@ getDomain(isl_ctx *ctx, Operation *op, bool overApproximationAllowed = false) {
 
   isl_space *space =
       isl_space_set_alloc(ctx, cst.getNumSymbolVars(), cst.getNumDimVars());
+  for (unsigned i = 0; i < cst.getNumDimVars(); i++) {
+    isl_id *id = isl_id_alloc(ctx, "dim", (void *)(size_t)(i + 1));
+    space = isl_space_set_dim_id(space, isl_dim_set, i, id);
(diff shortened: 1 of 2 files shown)


wsmoses (commented):
@ivanradanov @ftynse is this the right way to do things?

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -207,6 +207,14 @@ getDomain(isl_ctx *ctx, Operation *op, bool overApproximationAllowed = false) {
 
   isl_space *space =
       isl_space_set_alloc(ctx, cst.getNumSymbolVars(), cst.getNumDimVars());
+  for (unsigned i = 0; i < cst.getNumDimVars(); i++) {
```
ftynse, line 210: Nit: https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -207,6 +207,14 @@ getDomain(isl_ctx *ctx, Operation *op, bool overApproximationAllowed = false) {
...
+    isl_id *id = isl_id_alloc(ctx, "dim", (void *)(size_t)(i + 1));
```
ftynse, line 211: isl IDs with identical names but different "user" fields are different ids, but they _print the same_. This is _extremely surprising_ when debugging and I cannot emphasize enough the recommendation to use differnt names instead.

Also, it's a bad idea to C-cast a small integer to a pointer and pass it as a legit pointer.

PR #2344, opened 2026-03-28
DotGeneral to trmm

snonk:
Patterns to replace `DotGeneral` with `trmm` op and fuse multiply operations

### DIFF
--- a/src/enzyme_ad/jax/TransformOps/TransformOps.td
+++ b/src/enzyme_ad/jax/TransformOps/TransformOps.td
@@ -3028,11 +3028,16 @@ def ApplyDotGeneralToSyrkPatterns : EnzymeHLOPatternOp<
   let patterns = ["DotGeneralToSyrk"];
 }
 
-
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/TransformOps/TransformOps.td
```diff
@@ -2944,11 +2944,16 @@ def ApplyDotGeneralToSyrkPatterns : EnzymeHLOPatternOp<
...
 }
 
-
 def ApplyDotGeneralToSymmPatterns : EnzymeHLOPatternOp<
     "dot_general_to_symm"> {
   let patterns = ["DotGeneralToSymm"];
 }
+
+def ApplyDotGeneralToTrmmPatterns : EnzymeHLOPatternOp<
+    "dot_general_to_trmm"> {
```
avik-pal, line 3037: remember to add in the list in EnzymeXLA.cpp (it will get automatically included in the list of passes for reactant)

PR #2347, opened 2026-03-28
Fix crash in DelinearizeIndexing

Pangoraw:
https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/23658698871/job/68922878671?pr=2269

### DIFF
--- a/test/lit_tests/crash_delinearize_indexing.mlir
+++ b/test/lit_tests/crash_delinearize_indexing.mlir
@@ -1,0 +1,17 @@
+// RUN: enzymexlamlir-opt %s --delinearize-indexing | FileCheck %s
+
+module {
+  func.func @main(%arg0: memref<i64>, %i: index) -> f32 {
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Passes/DelinearizeIndexing.cpp
```diff
@@ -109,7 +109,7 @@ reshapeMemref2(Value memref, ArrayRef<int64_t> shape,
       AffineMap map = ainfo.map;
       for (auto expr : map.getResults()) {
         auto cst = dyn_cast<AffineConstantExpr>(expr);
-        if (cst.getValue() != 0)
+        if (cst && cst.getValue() != 0)
```
wsmoses: I think the logic is wrong, don't you want !cst || cst.getValue() != 0 ?

wsmoses: @ivanradanov

PR #2350, opened 2026-03-29
Add TGammaOp

dkytezab:
cc @Pangoraw. Provided definition is incorrect for negative operand

### DIFF
--- a/src/enzyme_ad/jax/Implementations/EnzymeXLADerivatives.td
+++ b/src/enzyme_ad/jax/Implementations/EnzymeXLADerivatives.td
@@ -31,6 +31,7 @@ class HLOConstantFP<string m> : ConstantFP<m, "stablehlo", "ConstantOp", "mlir::
 // Required operations from the EnzymeXLA dialect
 def Gelu : EnzymeXLAInst<"GeluOp">;
 def Relu : EnzymeXLAInst<"ReluOp">;
+def TGamma : EnzymeXLAInst<"TGammaOp">;
 
 // Required operations from StableHLO/CHLO dialects
 def Add : Inst<"AddOp", "stablehlo">;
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/LowerEnzymeXLAML.cpp
```diff
@@ -43,6 +44,22 @@ struct LowerReluOpToStablehlo : public OpRewritePattern<enzymexla::ReluOp> {
...
+
+  LogicalResult matchAndRewrite(enzymexla::TGammaOp op,
+                                PatternRewriter &rewriter) const override {
+    auto loc = op.getLoc();
+    auto operand = op.getOperand();
+
+    auto LGamma = stablehlo::materializeLgamma(rewriter, loc, operand);
+    auto TGamma = stablehlo::ExpOp::create(rewriter, loc, LGamma);
+
+    rewriter.replaceOp(op, TGamma);
```
wsmoses: make this a select of x < 0, then you're corrent wrt nv tgamma:

https://docs.nvidia.com/cuda/libdevice-users-guide/__nv_tgamma.html

src/enzyme_ad/jax/TransformOps/TransformOps.td
```diff
@@ -253,6 +253,10 @@ def ApplyGeluConstProp : EnzymeHLOPatternOp<
     "gelu_const_prop">{
   let patterns = ["GeluConstProp"];
 }
+def ApplyTGammaConstProp : EnzymeHLOPatternOp<
+    "tgamma_const_prop">{
+  let patterns = ["TGammaConstProp"];
```
Pangoraw, line 258: The TGammaConstProp pattern needs to be written, you can take example on the gelu :

https://github.com/EnzymeAD/Enzyme-JAX/blob/68266757735dd73bd3dbafe8be42673af73a7c12/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp#L6468-L6485

PR #2367, opened 2026-03-31
fix: add a optimization barrier to prevent folding

avik-pal:


src/enzyme_ad/jax/Passes/ConvertAllConstantsToSplattedConstantPass.cpp
```diff
@@ -30,7 +30,12 @@ struct ConvertSHLOConstantsToSplat
     Attribute firstVal = *elementIt;
 
     auto splatAttr = SplatElementsAttr::get(op.getType(), firstVal);
-    rewriter.replaceOpWithNewOp<stablehlo::ConstantOp>(op, splatAttr);
+    auto newConstant =
+        stablehlo::ConstantOp::create(rewriter, op.getLoc(), splatAttr);
+    // Insert a optimization barrier to prevent the constant from being folded
+    auto barrier = stablehlo::OptimizationBarrierOp::create(
```
wsmoses, line 36: Can you explain why we want this?

avik-pal: This is effectively a "this large constant makes the file too big to load, so lets do an illegal transform to make it readable". Once we remove the large non-splatted constant there is now a chance that the new constant gets folded.

PR #2374, opened 2026-04-01
Add MPI Bcast operation, lowering and mlir test

yanzin00:
This PR introduces support for the **MPI_Bcast** collective communication operation into the EnzymeXLA MLIR dialect. It follows the same pattern established by the existing MPIAllreduceOp and adds the full pipeline: dialect op definition, CPU lowering pass, and a lit MLIR test.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -1428,7 +1428,7 @@ def MPIBcastOp : EnzymeXLA_Op<"mpi.bcast", []> {
                 "`MPI_Bcast(&buf, count, datatype, root, MPI_COMM_WORLD)`";
 
   let arguments = (
-    ins AnyTensor : $buf,
+    ins AnyTensor : $inbuf,
     TensorOf<[I32]> : $count,
     TensorOf<[I32]> : $root,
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
```diff
@@ -1423,4 +1423,22 @@ def MPIAllreduceOp : EnzymeXLA_Op<"mpi.allreduce", []> {
   let assemblyFormat = "`(` operands `)` attr-dict `:` functional-type(operands, results)";
 }
 
+def MPIBcastOp : EnzymeXLA_Op<"mpi.bcast", []> {
+  let summary = "Equivalent to "
+                "`MPI_Bcast(&buf, count, datatype, root, MPI_COMM_WORLD)`";
+
+  let arguments = (
+    ins AnyTensor : $buf,
```
romanlee:
```suggestion
    ins AnyTensor : $inbuf,
```
Minor, but to make consistent with the other ops

yanzin00: Sure! But just to clarify, I've added it as $buf to be consistent with the arguments of send/isend that are using $buf instead of $inbuf

PR #2424, opened 2026-04-14
Implement reverse-mode adjoint for Pointer2MemrefOp and Memref2PointerOp

xys-syx:
Register reverse-mode ReverseAutoDiffOpInterface and forward-mode ReadOnlyIdentityOp for enzymexla.pointer2memref and enzymexla.memref2pointer. Without this, the pipeline would fail with could not compute the adjoint for this operation.

The issues are:
Pointer2MemrefOp and Memref2PointerOp are pure view casts that carry no computation and only reinterpret the same underlying storage under a different type. Reverse-mode AD needs to know about them so that later memref.load and memref.store adjoints can locate the corresponding shadow view.

### DIFF
--- a/src/enzyme_ad/jax/Implementations/EnzymeXLAAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/EnzymeXLAAutoDiffOpInterfaceImpl.cpp
@@ -129,6 +129,35 @@ struct GPUWrapperOpEnzymeOpsRemover
   }
 };
 
+template <typename OpTy>
+struct ViewCastOpInterfaceReverse
+    : public ReverseAutoDiffOpInterface::ExternalModel<
+          ViewCastOpInterfaceReverse<OpTy>, OpTy> {
+  LogicalResult createReverseModeAdjoint(Operation *op, OpBuilder &builder,
+                                         MGradientUtilsReverse *gutils,
+                                         SmallVector<Value> caches) const {
+    return success();
+  }
+
+  SmallVector<Value> cacheValues(Operation *op,
+                                 MGradientUtilsReverse *gutils) const {
+    return {};
+  }
+
+  void createShadowValues(Operation *op, OpBuilder &builder,
+                          MGradientUtilsReverse *gutils) const {
+    auto castOp = cast<OpTy>(op);
+    auto newCastOp = cast<OpTy>(gutils->getNewFromOriginal(op));
+    Value source = castOp.getSource();
+    if (!gutils->isConstantValue(source)) {
+      Value sourceShadow = gutils->invertPointerM(source, builder);
+      auto shadowCast = cast<OpTy>(builder.clone(*newCastOp));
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/Implementations/EnzymeXLAAutoDiffOpInterfaceImpl.cpp
```diff
@@ -129,6 +129,35 @@ struct GPUWrapperOpEnzymeOpsRemover
...
+    return {};
+  }
+
+  void createShadowValues(Operation *op, OpBuilder &builder,
+                          MGradientUtilsReverse *gutils) const {
+    auto castOp = cast<OpTy>(op);
+    auto newCastOp = cast<OpTy>(gutils->getNewFromOriginal(op));
+    Value source = castOp.getSource();
+    if (!gutils->isConstantValue(source)) {
+      Value sourceShadow = gutils->invertPointerM(source, builder);
```
Pangoraw, line 153: Here the adjoint should be cached instead of using the source directly.

See https://github.com/EnzymeAD/Enzyme/pull/2508

xys-syx: Please correct me if I am wrong: my implementation here is structurally identical to `SubViewOpInterfaceReverse` https://github.com/EnzymeAD/Enzyme/blob/7f6c355995bd53ce9c8958329b0688069eeabab7/enzyme/Enzyme/MLIR/Implementations/MemRefAutoDiffOpInterfaceImpl.cpp#L198
which also calls `invertPointerM(source, builder)` directly inside `createShadowValues` and registers the result via `setInvertedPointer`. https://github.com/EnzymeAD/Enzyme/pull/2508 updated load/store to cache the shadow memref through push/pop, but did not touch the view ops like `SubViewOp`. My understanding is that the new caching pattern on the consumer side already handles the cross-block dominance issue, as long as producers like `SubViewOp` and `ViewCastOp` keep registering their shadow via `setInvertedPointer`. 

Could you please tell what the intended fix looks like here?
1. Keep the current structure (matching `SubViewOp`) and the [#2508](https://github.com/EnzymeAD/Enzyme/pull/2508) caching on the consumer side handles the cross-block issue which already works in my lit test covering the common loop case.
2. Migrate both `SubViewOp` and the new `ViewCastOp` to an explicit cache pattern in cacheValues (pushing the shadow to a cache).

Pangoraw (approved):

xys-syx:
Cache (enzyme.push/enzyme.pop) is used for Ops which has calculation in the reverse block. While enzyme.pointer2memref and enzyme.memref2pointer ops only reinterpret the same storage between !llvm.ptr and memref<?xT>, so the reverse adjoint is a no-op and no push/pop cache is needed. We only need to create shadow view in augmented forward pass to tell downstream memref.load/store where to get shadow.

PR #2426, opened 2026-04-14
add diff rule for `chlo.square`

mofeing:
closes #2404

cc @joelberkeley


Pangoraw:
Can you add a test with complex numbers?

https://openxla.org/stablehlo/generated/chlo#chlosquare_chlosquareop

mofeing:
while testing the complex version of this, I found out that diff rules that use `HloConstantOp` are broken for complex numbers.

waiting for EnzymeAD/Enzyme#2790 to be merged to continue on this.

PR #2436, opened 2026-04-16
add diff rules for `stablehlo.triangular_solve`

mofeing:


mofeing:
UPDATE: the diff rules require an update. they are wrong depending on the combination of `lower`, `leftSide`, `unitDiagonal` and `transpose_a`.

But openxla/stablehlo#2943 should help definitely to test all combinations correctly.

joelberkeley:
Has this been put aside?

PR #2437, opened 2026-04-17
Tessera Pattern Matching with Struct Arguments

jessicacotturone21:
The goal of this pull request is to allow for more pattern matching to occur by eliminating intermediate stores when we have struct arguments and returns. We also allow users to specify whether arguments are byref in the source code (as well as the size of the arguments), and any arguments specified as byref will be loaded and converted to byval in our tessera passes.

The user can now specify that an argument is byref like this:
`tessera_op = eigen.inv(x: byref, sizeof(Mat))`
where Mat is a struct and x is the matrix we want to take the inverse of.

Previously, after running the `llvm-to-tessera` pass and pulling out the sret argument, we would end up with something that looks like this:
```
%3 = tessera.call @eigen.inv(%2)
llvm.store %3, %1
%4 = tessera.call @eigen.inv(%1)
```

By loading the struct value from the pointer if it is byref and checking that the `tessera.call` op does not capture the pointer, we can perform store-to-load forwarding with the `polygeist-mem2reg` pass and end up with:
```
%3 = tessera.call @eigen.inv(%2)
%4 = tessera.call @eigen.inv(%3)
```
which our PDL pattern can match. This pull request updates the `polygeist-mem2reg` pass with additional type conversion logic and adds a canonicalization on tessera calls to allow this to happen.

Any operands that were changed from pointers to loaded values are kept track of and new `llvm.alloca` instances are created and pointer operands are reinserted when lowering back to LLVM.

The `MemoryEffectOpInterface` has also been implemented and `pure_tessera_ops` added, so that a user can declare a function as side effect free. This will allow for the elimination of lingering `tessera.call` ops that remain after the optimization rewrite has been performed.


src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
```diff
@@ -149,13 +151,20 @@ LogicalResult CallOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
-    if (getOperand(i - startIdx).getType() != fnType.getInput(i))
-      return emitOpError("operand type mismatch: expected operand type ")
-             << fnType.getInput(i) << ", but provided "
-             << getOperand(i - startIdx).getType() << " for operand number "
-             << i - startIdx;
+  for (unsigned i = startIdx, e = fnType.getNumInputs(); i != e; ++i) {
+    if (getOperand(i - startIdx).getType() == fnType.getInput(i))
+      continue;
+    if (isa<LLVM::LLVMPointerType>(fnType.getInput(i)) &&
+        fn.getArgAttr(i, LLVM::LLVMDialect::getReadonlyAttrName()))
```
wsmoses, line 161: you also need to check nocapture, and it can be readonly or readnone

src/enzyme_ad/jax/Passes/Tessera/LLVMToTessera.cpp
```diff
@@ -127,37 +139,72 @@ class CallOpRewrite final : public OpRewritePattern<LLVM::CallOp> {
...
-              firstArgAttrs.get(LLVM::LLVMDialect::getStructRetAttrName())) {
+      if (auto sretAttr = defineOp.getArgAttr(
+              0, LLVM::LLVMDialect::getStructRetAttrName())) {
         sretPtr = callOp.getOperand(0);
         sretType = cast<TypeAttr>(sretAttr).getValue();
-        // Build operands and arg attributes without first element
-        for (int i = 1; i < operands.size(); i++)
-          newOperands.push_back(callOp.getOperand(i));
+        // Build operands without first element. If a pointer operand is
+        // readonly and there is a preceding store to that pointer,
```
wsmoses, line 165: consider the
```
double sum(eigen::Matrix)
```

case you still may need to lower the parameter, even if the return isn't an sret

wsmoses (commented):
consider a fn

```
double* pointeradd(double* x, int y) {
   return &x[y];
}
```

we could define an opt `pointeradd(pointeradd(x, y), z) -> pointeradd(x, y+z)`

double x[10] -> double* [ptr]

PR #2440, opened 2026-04-19
fix: lower cuda symm to gemm if tensor is pre-populated

avik-pal:
@snonk can you check if this fixes the benchmark we have?


snonk (approved):
Seems like it fixes it! In newton schulz the `symm` becomes a `gemm` now and the runtime goes down from 2.46s to 2.16s for 4096x4096 matrices

PR #2444, opened 2026-04-22
Cublas Support

dsding2:
Adds support for cublasSgemm, including caching and runtime shape propagation.


mofeing:
~Doesn't XLA already lower `stablehlo.dot_general` to `cuBlas<t>gemm` when running on GPU?~

EDIT: nvm, I see that this has nothing to do with it

PR #2448, opened 2026-04-25
define 'hydra' bazel config profile

mofeing:
this should flag setting in hydra for development, by just appending `--config=hydra` option.

```
bazel build --config=hydra :enzymexlamlir-opt
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -62,3 +62,8 @@ common --define=allow_oversize_protos=true
 build --@rules_python//python/config_settings:precompile=force_disabled
 
 build -c opt
+
+# hydra config
+common:hydra -c dbg
+common:hydra --repo_env=CC=clang-18
+common:hydra --copt=-fbracket-depth=1024 --host_copt=-fbracket-depth=1024
--- a/DEVDOCS.md
+++ b/DEVDOCS.md
@@ -7,8 +7,9 @@ The `.devcontainer` configuration is helpful but not recommended. While it provi
 ## Building the Project
 
 ### Quick Build
+
 ```bash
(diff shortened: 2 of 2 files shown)


DEVDOCS.md
```diff
@@ -7,8 +7,9 @@ The `.devcontainer` configuration is helpful but not recommended. While it provi
 ## Building the Project
 
 ### Quick Build
+
 ```bash
-bazel build --repo_env=CC=clang-18 --color=yes --copt=-fbracket-depth=1024 --host_copt=-fbracket-depth=1024 -c dbg :enzymexlamlir-opt
+bazel build --config=hydra :enzymexlamlir-opt
```
giordano, line 12: Why is it special? Also, hardcoding clang-18 as default compiler isn't a good idea, not everyone has that.

.bazelrc
```diff
@@ -62,3 +62,8 @@ common --define=allow_oversize_protos=true
 build --@rules_python//python/config_settings:precompile=force_disabled
 
 build -c opt
+
+# hydra config
+common:hydra -c dbg
+common:hydra --repo_env=CC=clang-18
+common:hydra --copt=-fbracket-depth=1024 --host_copt=-fbracket-depth=1024
```
wsmoses, line 69: the bracket depth stuff shouldn't be needed anymore, it should be autodetected depending on the host compiler?

mofeing: Okay, we should then update the DEVDOCS

wsmoses: yeah for sure! I think we fixed in https://github.com/EnzymeAD/Enzyme-JAX/pull/2235

DEVDOCS.md
mofeing: I guess devdocs were written with just hydra in mind. That's why I named the config profile like it.

mofeing:
Given that the bracket flags should be auto detected by now, this PR doesn't do much. So I'm closing and updating the devdocs.

PR #2451, opened 2026-04-28
Refactor clang_compile, compile_with_xla and enzyme_call to reduce reliance on nanobind

Antipath1:
I tried to use `compile_mhlo_to_llvm_with_xla` and `createLLVMMod` from Julia to add new reflection capabilities to Reactant.jl; however, `nanobind` does not work well when called from Julia. 

This PR refactors the code to use `absl::Status` and `absl::StatusOr` for all error handling previously handled by `nanobind`. 
* Moved `createLLVMMod` into `clang_compile`.
* `enzyme_call` is now strictly used for the Python interop. It unwraps the `absl` returns, meaning the Python side has no actual changes if I did my job right.
* Changed `ArgumentList` to use `std::list` to prevent dangling `c_str()` pointers during Clang argument compilation, resolving a segmentation fault. (Unsure if the segfault was the fault of the refactor or existed beforehand and just was exposed through the refactor)

The Reactant.jl side can now cleanly use `compile_mhlo_to_llvm_with_xla` and `createLLVMMod`. My downstream Reactant PR is still a work in progress. 

**Test Status:**
997 tests are passing on my machine. The refactor initially resulted in a segfault in test_custom_cpp_kernel. After I changed `ArgumentList` to fix the segfault, it now fails with an LLVM assertion (cannot get terminator of non-well-formed block) in _enzyme_rev_lowering. I am unsure if the segfault was an existing problem or if I messed something up during the refactor, but I think it was problematic beforehand. I don't really know why the LLVM assertion error happens, and I hope one of you can help me with that.


src/enzyme_ad/jax/clang_compile.cc
```diff
@@ -96,38 +108,17 @@ using namespace llvm;
 
 class ArgumentList {
 private:
-  /// Helper storage.
-  llvm::SmallVector<llvm::SmallString<0>> Storage;
-  /// List of arguments
+  std::list<std::string> Storage;
```
wsmoses: this specifically was coded like this to avoid a memory error. Can you separate code motion / status returning PR's from those which change functionality?

Antipath1: Reverted that commit back to the original

Antipath1: Just to clarify: would you like me to separate the code motion (moving the functions into clang_compile) and the absl::Status refactor into two distinct PRs?

I initially bundled them because Reactant needs both the code motion (to avoid pulling in nanobind) and the status returns (to well aalso avoid pulling in nanobind) to actually use the functions. I can split them into two dependent PRs if that is your better. Just let me know.

PR #2454, opened 2026-04-29
moving creaateLLVMMod inside of clang_compile.cc

Antipath1:
Code-motion PR in setup of nanobind removal from clang_compile and compile_with_xla. I also changed the header files accordingly and I moved the enums from enzyme_call into the header of clang_compile since they are now needed in both enzyme_call and clang_compile.


wsmoses (approved):
approved, assuming is purely code motion

Antipath1:
Hi! I was a bit unsure whether the CI failures here were due to my changes or a pre-existing problem. I checked the main branch and it looks like it's failing there with the exact same error, so it seems like a pre-existing issue with the tests.

wsmoses:
clang format for sure is, other ones can you rebase?

Antipath1:
Formatting is done and the branch is rebased. All the required tests are now passing, except for macOS, which is failing due to a Google Cloud Auth error (it looks like it's trying to push to the cache without PR secrets).

PR #2459, opened 2026-05-02
Add PDLL implementation of SliceExtend

jumerckx:
I'm interested in porting all patterns in EnzymeHLOOpt.cpp to PDLL. Hopefully, this will lead to compile time improvements as multiple patterns can be matched together.
My ulterior motive here is to have all patterns available in pdl,  such that we can run equality saturation on them using https://github.com/jumerckx/Tamagoyaki

For now, I've just implemented the PDLL pattern for SliceExtend. The PDLL file contains both native constraints and rewrites. When more patterns are ported, we could perhaps look into code sharing with the original implementations.
But maybe, in the long run, PDLL/pdl could serve as a single source of truth? @wsmoses I remember you also made a case for having everything in tablegen. To me, PDLL looks like the lesser of two evils?

Interested to hear people's thoughts!


wsmoses:
Can we instead move the patterns to tablegen (we have a few at the moment), and then have tablgen lower to pdl?

Either way though solid goal

jumerckx:
Could you elaborate on why you prefer DRR? Would it only be preferred because of the ability to generate C++?

I'd argue PDLL is a nicer fit for rewrite patterns compared to DRR ([the docs have this to say](https://mlir.llvm.org/docs/PDLL/#why-build-a-new-language-instead-of-improving-tablegen-drr), better constraints and multiple replacements especially seem useful).
I also expect LLMs to fare better with PDLL, I plan on using those to do large chunks of the porting.
Could we instead not have `pdl_interp` to C++ compilation? In measurements I've done, pdl_interp interpretation overhead is significant (but doesn't outweigh benefit of matcher sharing), so compiling pdl_interp would be useful anyhow.

wsmoses:
Let me look more into this [and/or we can discuss in a meeting?] My initial thought is that we'd ideally want to make sure rewrites are compatible with the existing infrastructure and also pdl so we don't duplicated patterns during a transition

jumerckx:
A meeting at some point would be great. I marked this pr as draft for now because there's actually a small bug upstream in how named operation arguments in constraints are handled. I've opened a pr here https://github.com/llvm/llvm-project/pull/195554

jumerckx:
@wsmoses 
I've made some changes to this. The pdll file now only contains the high level pattern, actual rewrites and constraints are moved to a C++ file. This allows us to share constraint functions between pdll and regular C++ patterns. 

Furthermore, constraints and rewrites are parametrized: the user can choose local-pure, local-checked, and global-checked. With the first being what I'm interested in w.r.t. equality saturation, and the last being what the current c++ patterns do. I'm not entirely sure if the middle option is useful.

If you like this direction, I can split this pr into two: first one introducing pdll infra and the next deduplicating code with the existing c++ code.
Otherwise, of course open for your thoughts.

Note CI still fails because Enzyme-JAX does not yet include https://github.com/llvm/llvm-project/pull/195554

PR #2464, opened 2026-05-09
comm.mpi subdialect

mofeing:
WIP


mofeing:
merged to the `comm-dialect` branch, not main

PR #2465, opened 2026-05-09
Add rcp raising

wsmoses:


giordano:
Does this catch `llvm.nvvm.rcp.approx.ftz.f` and `llvm.nvvm.rcp.approx.ftz.d` (which were introduced by https://github.com/JuliaGPU/CUDA.jl/pull/3077)? I only see `__nv_drcp_rn` and `__nv_frcp_rn`

wsmoses:
No it does not, we still want the fast div to go to a div instruction so we don't add a bunch of newton steps unnecessarily to an already precise div

PR #2475, opened 2026-05-11
multifloat pow

wsmoses:
a quick vibe code, cc @dkytezab 

doesn't seem like it did a good job tbh


dkytezab:
@wsmoses could this get speedy approve

wsmoses (commented):
so I'm super confused/question why anything other than the new op needs to be changed

dkytezab:
the change to multifloat pass also refactored log op conversion, other two tests were broken by changes to `IsResultOrOperandTypeLegal` i believe

wsmoses:
yeah can we extract this to only do pow [the original code here was just an initial vibe code which went a bit wild]

dkytezab:
q, why not just write `pow(x, y)` as `exp(y * log(x))`, with case handling, seeing as we have `exponential, log, multiply` already. also easier down the line as we don't have to write diff versions of `pow` for more limbs

wsmoses:
well the reason being that doing so might introduce more errors

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -2546,6 +2546,29 @@ struct CompareOpConversion : public OpConversionPattern<stablehlo::CompareOp> {
       if (res.getType() != resultType) {
         res = rewriter.create<stablehlo::ReshapeOp>(loc, resultType, res);
       }
+
+      // Infinity correction: two multifloat values that both represent ±∞
+      // compare as equal regardless of their lo limbs (which may be NaN due
+      // to inf_f64 - inf_f32 = NaN during splitting). We check only the hi
+      // limb: (hi_lhs == hi_rhs) AND !is_finite(hi_lhs).
+      // NaN is excluded automatically since NaN == NaN is false at f32 level.
+      Value hi_lhs = extractLimb(lhs, 0, rewriter, loc, concatDimension);
```
wsmoses, line 2670: can you add numeric tests in the checks for this (so we don't regress)?

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -3147,6 +3194,23 @@ struct LogOpConversion : public OpConversionPattern<stablehlo::LogOp> {
     Value is_inf = rewriter.create<stablehlo::CompareOp>(
         loc, x_hi, inf, stablehlo::ComparisonDirection::EQ);
 
+    if (concatDimension == "tuple") {
+      // In TUPLE mode the packed result is a TupleType, so we cannot
+      // broadcast a scalar condition over it or use SelectOp on tuples.
+      // Apply the special-case selects per limb, then pack.
```
wsmoses, line 3323: also add checks for this?

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4170,6 +4234,101 @@ struct MultiFloatConversionPass
     LLVM_DEBUG(llvm::dbgs() << "concatDimension: " << concatDimension << "\n");
     LLVM_DEBUG(llvm::dbgs() << "expansionSize: " << expansionSize << "\n");
 
+    // Pre-lower pow(x,y) -> exp(y * log(x)) before dialect conversion so the
+    // existing log/mul/exp patterns handle it without TUPLE-mode ordering
+    // issues.
+    {
+      IRRewriter rewriter(context);
```
wsmoses: we really shouldn't do this, it's going to likely be more numerically unstable. At minimum, put it within a pattern rather than here

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4170,6 +4414,28 @@ struct MultiFloatConversionPass
...
+    // (integer fast path / sqrt / general exp-log fallback). See its docstring.
+    {
+      SmallVector<func::FuncOp> funcsToLower;
+      op->walk([&](func::FuncOp func) {
+        if (!func->hasAttr("enzyme.no_multifloat"))
+          funcsToLower.push_back(func);
+      });
+      RewritePatternSet powPatterns(context);
+      powPatterns.add<PowOpLowerPattern>(context, srcTy);
+      FrozenRewritePatternSet frozenPow(std::move(powPatterns));
```
wsmoses: why is this separate here and not in the general pattern?

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4147,6 +4211,186 @@ template <typename OpTy> struct IsResultOrOperandTypeLegal {
...
+          DenseElementsAttr::get(tensorTy, rewriter.getFloatAttr(elemTy, val));
+      return rewriter.create<stablehlo::ConstantOp>(loc, tensorTy, attr);
+    };
+    auto makeNaN = [&]() -> Value {
+      APFloat nan = APFloat::getNaN(elemTy.getFloatSemantics());
+      auto attr = DenseElementsAttr::get(tensorTy, FloatAttr::get(elemTy, nan));
+      return rewriter.create<stablehlo::ConstantOp>(loc, tensorTy, attr);
+    };
+
+    Value zero = makeConst(0.0);
```
wsmoses, line 4608: cc @vimarsh6739 @sbrantq 

how does this algorithmically compare to the multifloat.jl case

sbrantq: `constexpr int64_t kMaxUnrollExponent = 32;` understandable that we would need a cap for IR size whereas multifloat.jl's runtime loop uncapped is fine. Not sure if it would be a bit too small for certain kinds of applications?

nvm i think it's fine 👍

wsmoses:
@oscarsmith

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4147,6 +4211,186 @@ template <typename OpTy> struct IsResultOrOperandTypeLegal {
...
+//
+// The general fallback is exp(y * log(|x|)) with IEEE corrections, but that
+// path compounds the per-call error of log and exp (each itself approximated
+// in multifloat) and is therefore numerically unstable for non-trivial y.
+// We avoid it whenever the exponent reveals a structurally exact form:
+//
+//   * Constant integer n with |n| small  →  exponentiation by squaring
+//     (only multiplications; exact for n>=0, reciprocal for n<0).
+//   * Constant 0.5                       →  stablehlo.sqrt
+//     (multifloat sqrt is much more accurate than exp(0.5*log(x))).
```
oscardssmith: you likely want to handle `y*2` is an int. i.e. `x^3.5 => x^3*sqrt(x)`

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4147,6 +4211,186 @@ template <typename OpTy> struct IsResultOrOperandTypeLegal {
...
+//
+// For exponents that don't match a fast path (non-constant y, or constants
+// like 2/7 that have no closed form in elementary ops) we fall back to the
```
oscardssmith, line 4371: For rational `y=a/b`, you might want to use `x^(a/b) => bth_root(x^a, b)` where you compute b_th root using newton iteration with an initial value provided by the bth root of the highest limb.

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4147,6 +4211,186 @@ template <typename OpTy> struct IsResultOrOperandTypeLegal {
...
+// exp(y * log(|x|)) lowering. Restricted to ops whose lhs element type is
```
oscardssmith, line 4372: will Enzyme/LLVM be smart enough to constant fold `log(x)` when `x` is constant? If not, you definitely want that as a fast path.

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4147,6 +4211,186 @@ template <typename OpTy> struct IsResultOrOperandTypeLegal {
...
+            DenseElementsAttr::get(tensorTy,
+                                   rewriter.getFloatAttr(elemTy, 1.0)));
+        if (*n == 0) {
+          rewriter.replaceOp(op, one.getResult());
+          return success();
+        }
+        uint64_t k = static_cast<uint64_t>(std::abs(*n));
+        Value base = x;
+        Value result;
+        while (k > 0) {
```
oscardssmith: I think you want to do all the math in this loop in higher precision (because you will exponentially accumulate error)

src/enzyme_ad/jax/Passes/MultiFloatConversion.cpp
```diff
@@ -4147,6 +4211,186 @@ template <typename OpTy> struct IsResultOrOperandTypeLegal {
...
+//   * Constant integer n with |n| small  →  exponentiation by squaring
+//     (only multiplications; exact for n>=0, reciprocal for n<0).
+//   * Constant 0.5                       →  stablehlo.sqrt
+//     (multifloat sqrt is much more accurate than exp(0.5*log(x))).
+//
+// For exponents that don't match a fast path (non-constant y, or constants
+// like 2/7 that have no closed form in elementary ops) we fall back to the
+// exp(y * log(|x|)) lowering. Restricted to ops whose lhs element type is
+// the multifloat source type, and skipped inside enzyme.no_multifloat funcs.
+struct PowOpLowerPattern : public OpRewritePattern<stablehlo::PowOp> {
```
oscardssmith, line 4374: I think you're missing a required check for 1^x (to preserve behavior like 1^Inf). Look at Julia's pow (in Base/math.jl). It has a pretty clear list of the required checks in the right order if you care about ieee semantics.

PR #2490, opened 2026-05-16
refactor: rename the `enzymexla.ml` subdialect to `enzymexla.math`

avik-pal:
Most of the ops in this subdialect have nothing to do with ML at this point


wsmoses (approved):

wsmoses:
also will make sense when some of the special functions and sinc @jumerckx / @ptiede were looking at earlier get a dialect impl

avik-pal:
> also will make sense when some of the special functions and sinc @jumerckx / @ptiede were looking at earlier get a dialect impl

Yes, I was about to get an agent to do that when I realised we call this "ml"

avik-pal:
Changing `lower-enzymexla-ml` to `lower-enzymexla-math`. We need to update that when updating the enzymejax commit

avik-pal:
> Changing `lower-enzymexla-ml` to `lower-enzymexla-math`. We need to update that when updating the enzymejax commit

I added a fallback path that forwards the ml lowering call to math to avoid the widespread breakage. We can delete once reactant is updated

PR #2498, opened 2026-05-17
feat: gather with iota indexing generalization

avik-pal:


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -15334,68 +15384,235 @@ struct GatherOpCanon final
...
+    }
+
+    // Check that the sliced region is within bounds along all dimensions.
+    for (int64_t k = 0; k < opRank; ++k) {
+      if (sliceStarts[k] + sliceSizes[k] > operandShape[k]) {
+        return failure();
+      }
+    }
+
+    // Reshape the 1D operand to the factored multi-dimensional shape.
```
wsmoses, line 15558: we may [later on] need to handle more than just one dim splatting

wsmoses (approved):

PR #2501, opened 2026-05-19
Optimize complex multiplication between a complex number and a purely real or purely imginary number

### LINKED ISSUE #1988: Decompose elementwise multiply for complex numbers (one has zero imag)
```
func.func @main(%arg0: tensor<4368xf64> {enzymexla.memory_effects = []}, %arg1: tensor<4368xcomplex<f64>> {enzymexla.memory_effects = []}) -> tensor<4368xcomplex<f64>> attributes {enzymexla.memory_effects = []} {
    %0 = stablehlo.convert %arg0 : (tensor<4368xf64>) -> tensor<4368xcomplex<f64>>
    %1 = stablehlo.multiply %0, %arg1 : tensor<4368xcomplex<f64>>
    return %1 : tensor<4368xcomplex<f64>>
  }

// ----
// rewrite to the decomposed version
// ----

module @reactant_fn2 attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
  func.func @main(%arg0: tensor<4368xf64> {enzymexla.memory_effects = []}, %arg1: tensor<4368xcomplex<f64>> {enzymexla.memory_effects = []}) -> tensor<4368xcomplex<f64>> attributes {enzymexla.memory_effects = []} {
    %0 = stablehlo.real %arg1 : (tensor<4368xcomplex<f64>>) -> tensor<4368xf64>
    %1 = stablehlo.multiply %arg0, %0 : tensor<4368xf64>
    %2 = stablehlo.imag %arg1 : (tensor<4368xcomplex<f64>>) -> tensor<4368xf64>
    %3 = stablehlo.complex %1, %2 : tensor<4368xcomplex<f64>>
    return %3 : tensor<4368xcomplex<f64>>
  }
}
```

Antipath1:
closes #1988
This PR adds a rewrite pattern for stablehlo.multiply to optimize complex multiplication when the imaginary component of the first operand is zero.Mathematical simplification:Given 
$(a + bi) \times (c + di)$ : If $b = 0$, the expression simplifies: 
$(a \cdot c - b \cdot d) + (c \cdot b + a \cdot d)i \implies (a \cdot c - 0 \cdot d) + (c \cdot 0 + a \cdot d)i = a \cdot c + a \cdot d \cdot i$

### DIFF
--- a/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
+++ b/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
@@ -552,6 +552,8 @@ static void addBaseTransformPasses(std::vector<std::string> &list,
   list.push_back("reduce_max_min_mul_positive_scalar");
   list.push_back("add_complex_simplify");
   list.push_back("sub_complex_simplify");
+  list.push_back("real_complex_mul_simplify");
+  list.push_back("imaginary_complex_mul_simplify");
   list.push_back("exponential_minus_one_fuse");
   list.push_back("scatter_of_scatter_simplify");
 }
--- a/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
+++ b/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
@@ -587,6 +587,8 @@ static void addBaseTransformPasses(std::vector<std::string> &list,
   list.push_back("reduce_max_min_mul_positive_scalar");
   list.push_back("add_complex_simplify");
   list.push_back("sub_complex_simplify");
+  list.push_back("real_complex_mul_simplify");
+  list.push_back("imaginary_complex_mul_simplify");
   list.push_back("exponential_minus_one_fuse");
   list.push_back("scatter_of_scatter_simplify");
 }
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -7942,8 +7942,10 @@ struct RealComplexMulSimplify
         loc, realTensor.getType(), complexTensor);
     auto newRealPart = rewriter.create<stablehlo::MulOp>(
         loc, realTensor.getType(), realTensor, realPart);
+    newRealPart->setAttrs(op->getAttrs());
     auto newImagPart = rewriter.create<stablehlo::MulOp>(
         loc, realTensor.getType(), realTensor, imagPart);
+    newImagPart->setAttrs(op->getAttrs());
     auto newComplex = rewriter.create<stablehlo::ComplexOp>(
         loc, op.getType(), newRealPart, newImagPart);
     rewriter.replaceOp(op, newComplex);
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -7932,6 +7932,10 @@ struct RealComplexMulSimplify
     } else {
       return failure();
     }
+    // Should be better handled by constant folding
+    if (matchPattern(purelyRealOperand, m_Constant())) {
+      return failure();
+    }
     auto complexTensorTy = cast<TensorType>(op.getType());
     auto floatElTy =
         cast<ComplexType>(complexTensorTy.getElementType()).getElementType();
@@ -8000,6 +8004,10 @@ struct ImaginaryComplexMulSimplify
       purelyImaginaryOperand = op.getRhs();
       complexOperand = op.getLhs();
     } else {
+      return failure();
+    }
+    // Should be better handled by constant folding
+    if (matchPattern(purelyImaginaryOperand, m_Constant())) {
       return failure();
     }
     auto complexTensorTy = cast<TensorType>(op.getType());
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -7905,47 +7905,49 @@ struct RealComplexMulSimplify
     if (!isa<ComplexType>(resultType)) {
       return failure();
     }
-    auto convertOp = op.getLhs().getDefiningOp<stablehlo::ConvertOp>();
 
-    if (!convertOp) {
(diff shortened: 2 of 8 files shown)


mofeing (commented):
maybe you should make use of the `PurelyRealResultAnalysis` and `PurelyImagResultAnalysis`, so that this pattern can propagate further https://github.com/EnzymeAD/Enzyme-JAX/blob/14b57c2952e1ba41d1891db2a233f915890ebf7e/src/enzyme_ad/jax/Utils.h#L867-L893

the way I would do is check if the the ops from the values you're checking (i.e. the ancestor ops to the `stablehlo.multiply`) actually contain such attributes, and run the pattern after the analysis.

Antipath1:
Okay, I have addressed this for the case of multiplication for a purely real number and a complex number. I have not gotten around to the case of a purely Imaginary number with a complex number, but I think other than that, I have addressed your feedback.

I had to edit two unrelated tests since using the pattern fired in two of the tests:

-    conj_convert.mlir: I removed the enzymexla.complex_is_purely_real = [#enzymexla] attribute from the second check since the new pattern firing does result in the exact same ops generated. The attribute is not carried with it, but since the result is a real number anyway, I do not think it really makes sense to have that attribute there, and at a cursory glance I do not think it will enable any other patterns that would not work anyway.
-    split_complex_scatter_setindex.mlir: The pattern also fires there, which is correct, but that, of course, leads to me also having to update the test's expected IR. I think the tested optimization there works exactly as before, there is just a previous part of the IR that gets changed, which I decided to change.

(cc @avik-pal: since you wrote these tests, could you double-check that my updates to the expected IR look right to you and tell me if I messed anything up there?)

Antipath1:
I added a pattern to simplify the case of multiplication between a purely imaginary number and a complex number. I also tried to make the changes to the existing tests as small as possible, and rebased against main.

Antipath1:
@mofeing, I could use your help here. There is a conflict with the last test defined in test/lit_tests/linalg/blas/gemm.mlir

```mlir

func.func @matmul_adjoint_adjoint(%A: tensor<4x2xcomplex<f32>>, %B: tensor<3x4xcomplex<f32>>) -> tensor<2x3xcomplex<f32>> {

    %alpha = stablehlo.constant dense<(1.0, 0.0)> : tensor<complex<f32>>

    %beta = stablehlo.constant dense<(0.0, 0.0)> : tensor<complex<f32>>

    %C = stablehlo.constant dense<(0.0, 0.0)> : tensor<2x3xcomplex<f32>>

    %0 = enzymexla.blas.gemm %alpha, %A, %B, %beta, %C {transa = #enzymexla.transpose<adjoint>, transb = #enzymexla.transpose<adjoint>} : (tensor<complex<f32>>, tensor<4x2xcomplex<f32>>, tensor<3x4xcomplex<f32>>, tensor<complex<f32>>, tensor<2x3xcomplex<f32>>) -> tensor<2x3xcomplex<f32>>

    return %0 : tensor<2x3xcomplex<f32>>

}

// CHECK-NEXT: func.func @matmul_adjoint_adjoint(%arg0: tensor<4x2xcomplex<f32>>, %arg1: tensor<3x4xcomplex<f32>>) -> tensor<2x3xcomplex<f32>> {

// CHECK-NEXT:   %cst = stablehlo.constant dense<(1.000000e+00,0.000000e+00)> : tensor<2x3xcomplex<f32>>

// CHECK-NEXT:   %0 = chlo.conj %arg0 : tensor<4x2xcomplex<f32>> -> tensor<4x2xcomplex<f32>>

// CHECK-NEXT:   %1 = chlo.conj %arg1 : tensor<3x4xcomplex<f32>> -> tensor<3x4xcomplex<f32>>

// CHECK-NEXT:   %2 = stablehlo.dot_general %0, %1, contracting_dims = [0] x [1] : (tensor<4x2xcomplex<f32>>, tensor<3x4xcomplex<f32>>) -> tensor<2x3xcomplex<f32>>

// CHECK-NEXT:   %3 = stablehlo.multiply %cst, %2 : tensor<2x3xcomplex<f32>>

// CHECK-NEXT:   return %3 : tensor<2x3xcomplex<f32>>

// CHECK-NEXT: }

```


The test expects to have the result above but due to the new realcomplexmulsimplify pattern seeing that %cst is a purely real number and then does what is supposed and rewrites the pattern in the end to this

```mlir

            func.func @matmul_adjoint_adjoint(%arg0: tensor<4x2xcomplex<f32>>, %arg1: tensor<3x4xcomplex<f32>>) -> tensor<2x3xcomplex<f32>> {

          %0 = chlo.conj %arg0 : tensor<4x2xcomplex<f32>> -> tensor<4x2xcomplex<f32>>

            %1 = chlo.conj %arg1 : tensor<3x4xcomplex<f32>> -> tensor<3x4xcomplex<f32>>

            %2 = stablehlo.dot_general %0, %1, contracting_dims = [0] x [1] : (tensor<4x2xcomplex<f32>>, tensor<3x4xcomplex<f32>>) -> tensor<2x3xcomplex<f32>>

            %3 = stablehlo.real %2 : (tensor<2x3xcomplex<f32>>) -> tensor<2x3xf32>

            %4 = stablehlo.imag %2 : (tensor<2x3xcomplex<f32>>) -> tensor<2x3xf32>

            %5 = stablehlo.complex %3, %4 : tensor<2x3xcomplex<f32>>

            return %5 : tensor<2x3xcomplex<f32>>

          }

          }

```
The mulop is gone I believe because my pattern splits the multiplication into real(%2) * 1.0 and imag(%2) * 1.0. They are then probably folded away because those 1.0 multiplications are identity operations, leaving only the complex reconstruction.

Which should be mathematically correct but causes a test failure. I personally can think of two options forward 
1.  either change the %alpha from 1.0 + 0.0i to 1.0 + 1.0i and rewrite the check to match it or 
2. In the first line of the test script, remove the --enzyme-hlo-op option.


Which one do you think would be better or do you have a better idea what to do here?

mofeing:
> The mulop is gone I believe because my pattern splits the multiplication into real(%2) * 1.0 and imag(%2) * 1.0. They are then probably folded away because those 1.0 multiplications are identity operations, leaving only the complex reconstruction.

this is alright. you can update the test. this is actually sth common that happens when we  add new passes. just change it to `1.0 + 1.0i` or sth more random so that it reflects the semantics of `gemm`.

sorry, I didn't se this notification at that time.

PR #2502, opened 2026-05-19
PolygeistMem2Reg: Type Casting in Store-to-Load Forwarding

jessicacotturone21:
Adds more cases for type casting in order to perform store-to-load forwarding, including:
- Struct to vector
- Struct to scalar (using intermediate vector)
- Array to vector
- Array to scalar (using intermediate vector)
- Vector to scalar

This works for nested structs as well.

### DIFF
--- a/test/lit_tests/mem2reg_type_casting.mlir
+++ b/test/lit_tests/mem2reg_type_casting.mlir
@@ -11,14 +11,10 @@ llvm.func @forward_same_type(%val: i32) -> i32 {
   llvm.return %0 : i32
 }
 
-// CHECK-LABEL: llvm.func @forward_same_type
-// CHECK-NOT:   llvm.inttoptr
-// CHECK-NOT:   llvm.ptrtoint
-// CHECK-NOT:   arith.bitcast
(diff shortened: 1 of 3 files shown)


test/lit_tests/mem2reg_type_casting.mlir
```diff
@@ -0,0 +1,215 @@
...
+  %al = llvm.alloca %c1 x f32 : (i32) -> !llvm.ptr
+  llvm.store %val, %al : f32, !llvm.ptr
+  %0 = llvm.load %al : !llvm.ptr -> i32
+  llvm.return %0 : i32
+}
+
+// CHECK-LABEL: llvm.func @forward_float_to_int
+// CHECK-NOT:   llvm.store
+// CHECK-NOT:   llvm.load
+// CHECK:       %[[BC:.+]] = arith.bitcast %[[VAL:.+]] : f32 to i32
```
wsmoses: You should check all instructions and operands here, this is not sufficient to ensure correctness as is (aka could have an indexing error)

wsmoses (approved):

PR #2509, opened 2026-05-25
Add rules for erfc and erfinv

ptiede:
Constructed with help from Claude


ptiede:
@wsmoses adds the other erf rules

wsmoses (approved):
I'm going to trust you checked the math, if so LGTM

PR #2512, opened 2026-05-26
Change index logic in affine raising

dkytezab:
Addresses https://github.com/EnzymeAD/Reactant.jl/issues/2923, somewhat vibed


wsmoses:
This should not be done here. Either the indexing should be simplified in affine-cfg or the post raising optimizations that @avik-pal is working on should simplify it back

PR #2513, opened 2026-05-27
Added check on function calls for no capture

jessicacotturone21:
Checks function calls that implement the `FunctionOpInterface` to see if they have the `llvm.nocapture` attribute and decide whether or not an alloca instance can be promoted based on whether or not the pointer is captured

### DIFF
--- a/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
+++ b/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
@@ -25,6 +25,7 @@
 #include "mlir/Support/LLVM.h"
 #include "mlir/Transforms/Passes.h"
 #include "src/enzyme_ad/jax/Dialect/Ops.h"
+#include "src/enzyme_ad/jax/Dialect/Tessera/Dialect.h"
 #include "src/enzyme_ad/jax/Passes/Passes.h"
 #include "src/enzyme_ad/jax/Utils.h"
 #include "llvm/ADT/SetVector.h"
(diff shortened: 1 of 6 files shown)


src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
```diff
@@ -1153,6 +1154,36 @@ Value castToType(Type elType, Value val, Operation *op) {
...
 }
 
+// Check if tessera call captures alloca instance by checking for llvm.nocapture
+// attribute
+bool isTesseraCallNonCapturing(tessera::CallOp callOp, Value val) {
+  auto calleeAttr = callOp.getCalleeAttr();
+  if (calleeAttr) {
+    auto callee = SymbolTable::lookupSymbolIn(
+        callOp->getParentOfType<ModuleOp>(), calleeAttr);
+    auto defineOp = dyn_cast_or_null<tessera::DefineOp>(callee);
```
wsmoses: functionopinterface [and then it applies to all fns]

src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
```diff
@@ -1153,6 +1154,36 @@ Value castToType(Type elType, Value val, Operation *op) {
...
+    if (!defineOp)
+      return false;
+
+    int offset = 0;
+    if (defineOp.getNumArguments() > 0 &&
+        defineOp.getArgAttr(0, LLVM::LLVMDialect::getStructRetAttrName()))
+      offset = 1;
```
wsmoses: this shouldn't be here if using function op interface [but the getArgAttr for defineOp should do the indexing change]

wsmoses (approved):

PR #2516, opened 2026-05-30
[CI] Use `pypa/gh-action-pypi-publish`

giordano:
@wsmoses I _think_ we need to tell it where to find the package installation (for example with the [`packages-dir`](https://github.com/pypa/gh-action-pypi-publish/blob/ba38be9e461d3875417946c167d0b5f3d385a247/README.md#customizing-target-package-dists-directory) input), but I don't really know where that would be here.  Is that `bazel-bin/`?

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -11,40 +11,98 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [openstack22-large]
-        python: ["3.11", "3.12", "3.13"]
+        os:
+          - linux-x86-n2-32
+          - linux-arm64-c4a-32
+        python:
+          - "3.11"
+          - "3.12"
+          - "3.13"
     timeout-minutes: 500
     steps:
-      - name: add llvm
+      - name: Set BAZEL_FLAGS
+        shell: bash
         run: |
-            if [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
-              sudo apt-get update
-              sudo apt-get install -y git gcc g++
-            fi
-      - uses: actions/setup-python@v6
-        with:
-          python-version: ${{ matrix.python }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tag.yml
```diff
@@ -11,42 +11,98 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [openstack22-large]
-        python: ["3.11", "3.12", "3.13"]
+        os:
+          - linux-x86-n2-32
+          - linux-arm64-c4a-32
```
giordano, line 16: These runners require a container

.github/workflows/tag.yml
```diff
@@ -11,42 +11,98 @@ jobs:
...
+        run: |
+          bazel run $BAZEL_FLAGS --color=yes //builddeps:requirements.update
+
+      - name: Build wheel
+        timeout-minutes: 240
+        run: |
+          bazel build $BAZEL_FLAGS --color=yes :wheel --test_output=errors
+
+      - name: Publish package distribution to PyPI
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
```
giordano, line 106: I don't really know how python publishing works: the fact that this is a matrix job means that this will be published multiple times? Or do we need a single job at the end where we put all the builds together?

.github/workflows/tag.yml
giordano: #2517

.github/workflows/tag.yml
wsmoses: At least for twine each job separately uploaded to the release containing the different artifact.

@giordano with your fix in the other PR do you want to delete the recent release, merge and retry release and we'll see what happens

PR #2547, opened 2026-06-10
AffineCFG: delinearize parallel IVs with implicit remainders

Pangoraw:
`SplitParallelInductions` failed to delinearize packed parallel induction variables, leaving complex floordiv access maps that hinder `raise-affine-to-stablehlo`.

Observed on an Oceananigans kernel where `affine.parallel (0,0,0) to (90,16,16)` packs three tile coordinates into one IV; with this fix the loop reaches the raising pass as `(0,0,0) to (10,40,40)` with trivial access maps (raised HLO for the kernel shrank from 768 to 550 stablehlo ops).

### 1. Fold implicit remainders in affine sums (`recreateExpr`)

The ISL round-trip in simplify-affine-exprs expands `(i mod k) * b` into the floordiv-difference form

```
i*b - (i floordiv k) * (b*k)
```

and `recreateExpr` failed to fold it back (the gcd-factoring in `commonAddWithMul` never reaches a shape that MLIR's built-in `e - q*(e floordiv q)` rule matches). Since `SplitParallelInductions` keys its profitability on seeing an explicit `mod`, the hidden remainders prevented the split.

`recreateExpr` now folds `c*e - c*k*(e floordiv k)` → `c * (e mod k)` — a pure integer identity, applied to both access maps and `affine.if` sets. Nested divisors need no dedicated rule: after the first split substitutes `iv = major*base + minor`, the bound-aware map optimization reduces e.g. `iv floordiv 9` to `major floordiv 3`, and the same fold applies on the next round, enabling the second split.

### 2. Use-list-order-dependent base detection

A `mod` divisor had to equal the running base exactly, while `floordiv` divisors allowed divisibility-and-take-the-min. With divisors 3 and 9 present, legality depended on the order `iv.getUsers()` was visited: if a floordiv-9-only user (e.g. an `affine.if` guard `iv floordiv 9 >= 1`) was seen first, a later `mod 3` marked the IV illegal. This made the pattern fail inside the full pipeline while succeeding on the reparsed IR dump of the very same function (print/parse reshuffles use lists).

`mod` now gets the same min-base divisibility handling. This is safe because affine uses are rewritten by exact substitution (`iv -> major * base + minor`); the `arith.remui` path, which substitutes `minor` directly and genuinely requires equality, keeps its strict check.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -4074,7 +4074,13 @@ struct SplitParallelInductions
                 base = newBase;
               } else if (!base.isValue && !newBase.isValue &&
                          (base.i_val == newBase.i_val ||
-                          (kind == AffineExprKind::FloorDiv &&
+                          // Affine uses are rewritten by exact substitution
+                          // (iv -> major * base + minor), so any base dividing
+                          // the seen divisors is correct; keep the smallest.
+                          // This must not depend on the order in which uses
+                          // are visited, so allow it for mod as well.
+                          ((kind == AffineExprKind::FloorDiv ||
+                            kind == AffineExprKind::Mod) &&
                            (base.i_val.urem(newBase.i_val) == 0 ||
                             newBase.i_val.urem(base.i_val) == 0)))) {
                 base.i_val =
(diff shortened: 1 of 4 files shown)


wsmoses (approved):
I'm not following the logic fully, but @Pangoraw if you're fine with it [and check the logic fully], I'm okay with it

Issue #2415, opened 2026-04-09
AffineCFG: MoveSIToFPToAffine causes infinite loop when sitofp operand is a valid symbol in scope


xys-syx:
The `MoveSIToFPToAffine` pattern in the `affine-cfg` pass enters an infinite loop when the `arith.sitofp` operand is an arithmetic op (`arith.muli`) defined outside the affine loop, making it already a valid symbol in the affine scope.

The minimal reproducible example is:
```
// RUN: enzymexlamlir-opt --affine-cfg %s

func.func @repro(%arg0: i32, %arg1: i32) -> f64 {
  %cst = arith.constant 0.0 : f64
  %res = affine.for %iv = 0 to 10 iter_args(%acc = %cst) -> (f64) {
    %mul = arith.muli %arg0, %arg1 : i32
    %fp = arith.sitofp %mul : i32 to f64
    %add = arith.addf %acc, %fp : f64
    affine.yield %add : f64
  }
  return %res : f64
}
```
I was thought pre-decompose the top-level arithmetic op in MoveSIToFPToAffine before calling fully2ComposeAffineMapAndOperands. For example, decompose `muli(%arg0, %arg1)` into `map = ()[s0, s1] -> (s0 * s1)`, `operands = {%arg0, %arg1}` instead of `map = ()[s0] -> (s0)`, `operands = {muli}`. This produces a non-identity `affine.apply` that won't be folded, breaking the cycle. However, this duplicates logic already present in `AffineApplyNormalizer`.

Could anyone suggest a better solution?

xys-syx:
**input is `%mul = arith.muli %arg0, %arg1 : i32`**
-> 
```
func.func @repro(%arg0: i32, %arg1: i32) -> f64 {
  %cst = arith.constant 0.0 : f64
  %res = affine.for %iv = 0 to 10 iter_args(%acc = %cst) -> (f64) {
    %mul = arith.muli %arg0, %arg1 : i32
    %fp = arith.sitofp %mul : i32 to f64 <- match here
    %add = arith.addf %acc, %fp : f64
    affine.yield %add : f64
  }
  return %res : f64
}
```
->

```
func.func @repro(%arg0: i32, %arg1: i32) -> f64 {
  %cst = arith.constant 0.0 : f64
  %mul = arith.muli %arg0, %arg1 : i32          // fix() remove to outside of loop
[...]

PR #2550, opened 2026-06-11
 Fix infinite loop in MoveSIToFPToAffine on non-decomposable operands

xys-syx:
Fix issue #2415 

The fix here:
After composition, if the operand did not decompose into any affine structure (the map is like `()[s0] -> (s0)`), return failure() instead of emitting the no-op `affine.apply`.

The reason I did not choose to fix in `isValidIndex`:
- https://github.com/EnzymeAD/Enzyme-JAX/blob/f5727c4a5043408379cf007c711e39a6df6ff3e9/src/enzyme_ad/jax/Passes/AffineCFG.cpp#L1220
https://github.com/EnzymeAD/Enzyme-JAX/blob/f5727c4a5043408379cf007c711e39a6df6ff3e9/src/enzyme_ad/jax/Passes/AffineCFG.cpp#L1246
`muli(%m,%n)` could be matched and return true in two if guard, for the first one, to fix it, need to change the condition here: https://github.com/EnzymeAD/Enzyme-JAX/blob/f5727c4a5043408379cf007c711e39a6df6ff3e9/src/enzyme_ad/jax/Passes/AffineCFG.cpp#L77
which I think would obey that: loop-invariant `muli(%m,%n)` can be valid symbol.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -2251,7 +2251,18 @@ struct MoveSIToFPToAffine : public OpRewritePattern<arith::SIToFPOp> {
     auto *parentScope = scope->getParentOp();
     DominanceInfo DI(parentScope);
 
-    fully2ComposeAffineMapAndOperands(rewriter, &map, &operands, DI, scope);
+    SmallVector<Operation *> insertedOps;
+    fully2ComposeAffineMapAndOperands(rewriter, &map, &operands, DI, scope,
+                                      &insertedOps);
+
+    if (map.getNumResults() == 1 &&
+        map.getResult(0).getKind() == AffineExprKind::SymbolId) {
+      for (Operation *op : llvm::reverse(insertedOps))
+        if (op->use_empty())
+          rewriter.eraseOp(op);
+      return failure();
+    }
+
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -2251,7 +2251,18 @@ struct MoveSIToFPToAffine : public OpRewritePattern<arith::SIToFPOp> {
     auto *parentScope = scope->getParentOp();
     DominanceInfo DI(parentScope);
 
-    fully2ComposeAffineMapAndOperands(rewriter, &map, &operands, DI, scope);
+    SmallVector<Operation *> insertedOps;
+    fully2ComposeAffineMapAndOperands(rewriter, &map, &operands, DI, scope,
+                                      &insertedOps);
+
+    if (map.getNumResults() == 1 &&
+        map.getResult(0).getKind() == AffineExprKind::SymbolId) {
```
wsmoses: Check that the symbol input is the same as the input operand

xys-syx: @wsmoses  now confirms if the symbol input is the same as the original sitofp operand before skipping the rewrite

wsmoses (approved):

PR #2557, opened 2026-06-14
[AffineCFG] Re-arrange sum expression after recreateexpr optimization

xys-syx:
The minimal reproducer example is:
```mlir
func.func @forbound_reassoc(%arg0: i32, %init: f64) -> f64 {
  %cst = arith.constant 1.000000e+00 : f64
  %s = arith.index_cast %arg0 : i32 to index
  %r = affine.for %i = 0 to %s iter_args(%acc = %init) -> (f64) {
    %ri = affine.for %j = 0 to affine_map<(d0)[s0] -> (-d0 + s0 - 1)>(%i)[%s]
            iter_args(%a = %acc) -> (f64) {
      %sum = arith.addf %a, %cst : f64
      affine.yield %sum : f64
    }
    affine.yield %ri : f64
  }
  return %r : f64
}
```
The bug here is:
when `CanonicalieForBounds` apply on `affine.for %j = 0 to affine_map<(d0)[s0] -> (-d0 + s0 - 1)>(%i)[%s]...`'s upper bound, the upper bound map and operands `affine_map<(d0)[s0] -> (-d0 + s0 - 1)>(%i)[%s]` will not change, but `recreateExpr` https://github.com/EnzymeAD/Enzyme-JAX/blob/6af35e740558bd5edbe5d7b1b694b4c39329aa07/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp#L453
change the tree associations of sum expression:
from `Add(Add(Mul(d0, C-1), s0), C-1)` to `Add(Mul(d0, C-1), Add(s0, C-1))`
causing `--affine-cfg` fails to converage.

The fix:
I add a `canonicalizeSum` to rearrange the sum expression, align with the rule in default mlir `simplifyAdd`, https://github.com/llvm/llvm-project/blob/5db7e5b483a32f37d1d77a6f18d2bc4c59fa460a/mlir/lib/IR/AffineExpr.cpp#L662

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -1914,19 +1914,19 @@ struct CanonicalieForBounds : public OpRewritePattern<affine::AffineForOp> {
     // ubMap.dump();
     // forOp.dump();
 
-    // Any canonicalization change in map or operands always leads to updated
-    // map(s).
-    if ((lbMap == prevLbMap && ubMap == prevUbMap) &&
-        (!areChanged(lbOperands, origLbOperands)) &&
-        (!areChanged(ubOperands, origUbOperands)))
+    bool lbChanged = canonicalizeSum(lbMap) != canonicalizeSum(prevLbMap) ||
+                     areChanged(lbOperands, origLbOperands);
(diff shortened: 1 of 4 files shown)


vimarsh6739:
@codex review

Issue #2514, opened 2026-05-28
Improve affine-cfg to remove if statement around affine.parallel


wsmoses:
https://tinyurl.com/25r7syxf

PR #2558, opened 2026-06-16
Add masked affine.parallel bounds refinement

Pangoraw:
handles the pattern where a affine.parallel has a single affine.if (with no else) in its body and we can therefore change the bounds of the parallel iteration space using the isl min/max analysis on the if integer set.

follow up is to do the reverse (for in if) #2514.


src/enzyme_ad/jax/Passes/AffineCFG.cpp
```diff
@@ -23,7 +23,9 @@
 #include "llvm/ADT/MapVector.h"
 
 #include <deque>
+#include <isl/aff.h>
```
wsmoses, line 26: we can't always assume the existance of isl and these need to be include gated [see other passes for an example]

Pangoraw: do you think about `DISABLE_POLYMER`?

https://github.com/EnzymeAD/Enzyme-JAX/blob/6af35e740558bd5edbe5d7b1b694b4c39329aa07/src/enzyme_ad/jax/Passes/RemoveAtomics.cpp#L12

I could not find something just for ISL

wsmoses: ah right, okay this is fine then without a guard

PR #2567, opened 2026-06-19
Add raw C++ no-AD call ABI

moomoohorse321:


moomoohorse321:
Sorry, agent acting toooooooo fast

PR #2575, opened 2026-06-22
Fix CbrtOp derivative NaN gradient on negative inputs (#2571)

### LINKED ISSUE #2571: stablehlo.cbrt derivative returns NaN gradient for negative inputs
## Human section

Spawning off a new issue, related to issue #2570. I'm somewhat certain that this is worth reporting becuase the generated tests show the fault clearly. I considered appending to that issue but this is located elsewhere. Is this valuable? Is it slop? I won't open another issue until I get a confirmation. Thanks for attention

## Machine section

**Component:** AD derivative rules (forward + reverse mode), not the HLO rewrite library.
**File:** `src/enzyme_ad/jax/Implementations/HLODerivatives.td`, `CbrtOp` rule (L1081).
**Class:** gradient-soundness. Autodiff returns NaN gradient where the true derivative is a finite real; the primal is correct.

**Rule:** `d/dx cbrt(x) = (1/3) * pow(x, -2/3)`, emitted as `DiffeRet * (pow(x, -2/3) / 3)` in both modes.

**Bug:** `stablehlo.cbrt` is IEEE `rootn(x, 3)`, so negative inputs are in-domain (`cbrt(-8) = -2`), smooth, with finite true derivative `1/(3 * cbrt(x)^2)`. The rule routes through `pow(x, -2/3)`, and a non-integer exponent on a negative base is NaN (StableHLO `power` spec example: `power([-36.0], [1.1]) = -nan`). So for finite `x < 0` the primal is correct but the gradient is NaN.

**Your own test already exhibits it.** `test/lit_tests/diffrules/stablehlo/cbrt.mlir` generates (forward and reverse) `stablehlo.power %arg0, dense<-0.666666686>`, and its `@main` numerically checks the gradient under `stablehlo-translate --interpret`. But it samples only positive inputs `[8.0, 27.0]`. Flip them to `[-8.0, -27.0]`: the true gradients are the same magnitudes `[0.083333, 0.037037]` (the derivative is sign-independent), so `%expected` is unchanged, but the computed gradient is NaN and the check fails. One-line mutation of the existing test:

```mlir
func.func @main() {
  %x = stablehlo.constant dense<[-8.0, -27.0]> : tensor<2xf32>          // was [8.0, 27.0]
  %out = stablehlo.constant dense<[-2.0, -3.0]> : tensor<2xf32>         // cbrt is real on negatives
  %expected = stablehlo.constant dense<[0.083333336, 0.037037037]> : tensor<2xf32>  // unchanged
  %dx = stablehlo.constant dense<1.0> : tensor<2xf32>
  %fwd:2 = enzyme.fwddiff @cbrt(%x, %dx) {
    activity=[#enzyme<activity enzyme_dup>], ret_activity=[#enzyme<activity enzyme_dup>]
  } : (tensor<2xf32>, tensor<2xf32>) -> (tensor<2xf32>, tensor<2xf32>)
  check.expect_almost_eq %fwd#0, %out : tensor<2xf32>        // primal cbrt(-8) = -2, passes
  check.expect_almost_eq %fwd#1, %expected : tensor<2xf32>   // gradient NaN, expected 0.0833, FAILS
  func.return
}
```

**Reference (non-circular):** `jax.lax.cbrt`'s JVP uses `g * (1/3) * integer_pow(ans, -2)`, i.e. the real primal answer, not `pow(x, -2/3)`. `jax.grad(jnp.cbrt)(-8.0) = 0.0833`, matching `1/(3 * cbrt(-8)^2)`.

**Fix:** compute from the already-real primal, as JAX does: `d/dx = DiffeRet / (3 * cbrt(x)^2)`. Because the rule broke by routing through `power` on a negative base, the safe form drops the `power` op entirely rather than swapping its exponent, so the fix can't repeat the failure. Minimal patch, mirroring the `LogOp` rule's `CheckedDiv` shape:

```tablegen
def : HLODerivative<"CbrtOp", (Op $x), [
  (CheckedDiv (DiffeRet), (Mul (HLOConstantFP<"3">), (Mul (Cbrt $x), (Cbrt $x)))),
]>;
```

Confirmed against `jax.grad(jnp.cbrt)` across the sign range (float64):

| x | `jax.grad` | current `pow(x,-2/3)/3` | proposed `1/(3·cbrt²)` |
|---|---|---|---|
| -8 | 0.083333 | NaN | 0.083333 |
| -2 | 0.209987 | NaN | 0.209987 |
| -0.5 | 0.529134 | NaN | 0.529134 |
| 2 | 0.209987 | 0.209987 | 0.209987 |

`integer_pow(cbrt(x), -2)` is equivalent if you'd rather keep a `pow`. Correct for finite `x != 0`; at `x = 0` the derivative is a genuine singularity (`+/-inf`), the same boundary the existing `SqrtOp` rule special-cases.

**Scope:** `stablehlo.cbrt` only. A definedness check over the unary derivative table flags this as the single rule that narrows the domain on negatives; the others (`log`, `log1p`, `sqrt`, `rsqrt`, `exp`, `expm1`, `sin`, `cos`, `tanh`, `logistic`) stay real, as do the recent erfinv/erfc/square rules.

kimjune01:
Fixes #2571 (maintainer-approved: "PR also welcome here!").

## Problem

The `CbrtOp` AD rule (`HLODerivatives.td`) computes the gradient as `(1/3) * pow(x, -2/3)`. A non-integer exponent on a negative base is NaN (StableHLO `power` spec: `power([-36.0], [1.1]) = -nan`), but `stablehlo.cbrt` is real and smooth for `x < 0` (`cbrt(-8) = -2`). So for finite `x < 0` the **primal is correct but the gradient is NaN**, where the true derivative `1/(3*cbrt(x)^2)` is a finite real.

## Fix

Compute the derivative from the already-real primal, dropping the `power` op entirely:

```
d/dx = DiffeRet / (3 * cbrt(x)^2)
```

This mirrors `jax.lax.cbrt`'s JVP (`g * (1/3) * integer_pow(ans, -2)`). Because the failure came from routing through `power` on a negative base, removing `power` means the fix cannot reproduce it. Verified against `jax.grad(jnp.cbrt)` across the sign range (float64):

| x | `jax.grad` | old `pow(x,-2/3)/3` | new `1/(3*cbrt^2)` |
|---|---|---|---|
| -8 | 0.083333 | NaN | 0.083333 |
| -2 | 0.209987 | NaN | 0.209987 |
| -0.5 | 0.529134 | NaN | 0.529134 |
| 2 | 0.209987 | 0.209987 | 0.209987 |

At `x = 0` the derivative is a genuine singularity (+/-inf), the same boundary `SqrtOp` already special-cases.

## Test

`test/lit_tests/diffrules/stablehlo/cbrt.mlir`'s `@main` numerically checks the gradient under `stablehlo-translate --interpret`, but only sampled positive inputs. Flipped to `[-8.0, -27.0]`: true gradients are sign-independent so `%expected` is unchanged, but the old rule produced NaN there. The FORWARD/REVERSE FileCheck blocks are updated to the new (power-free) IR.

> Note: I cannot build Enzyme-JAX locally, so the FORWARD/REVERSE `CHECK` lines are a best-effort prediction of the post-optimization IR. If CI's FileCheck flags an operand-order or constant-format mismatch, I will push a fixup to match. The source fix and the numeric `@main` check are the substantive parts.

Found mechanically with a value-and-gradient soundness gate over the rewrite/AD-rule library; happy to share the process if useful.

### DIFF
--- a/test/lit_tests/diffrules/stablehlo/cbrt.mlir
+++ b/test/lit_tests/diffrules/stablehlo/cbrt.mlir
@@ -25,31 +25,53 @@ func.func @cbrt(%x : tensor<2xf32>) -> tensor<2xf32> {
 // REVERSE-NEXT:    return %3 : tensor<2xf32>
 // REVERSE-NEXT:  }
 
-// Negative inputs: cbrt is real and smooth for x < 0 (cbrt(-8) = -2), so the
+// Exercise both signs. cbrt is real and smooth for x < 0 (cbrt(-8) = -2), so the
 // true gradient 1/(3*cbrt(x)^2) is a finite real with the same magnitude as on
-// the positives. The old pow(x,-2/3) rule returned NaN here. See issue #2571.
+// the positives. The old pow(x,-2/3) rule returned NaN on negatives. See issue #2571.
 func.func @main() {
-  %x = stablehlo.constant dense<[-8.0, -27.0]> : tensor<2xf32>
-  %out = stablehlo.constant dense<[-2.0, -3.0]> : tensor<2xf32>
-  %expected = stablehlo.constant dense<[0.083333336, 0.037037037]> : tensor<2xf32>
-
   %dx = stablehlo.constant dense<1.0> : tensor<2xf32>
 
-  %fwd:2 = enzyme.fwddiff @cbrt(%x, %dx) {
+  // Positive inputs: unchanged from the pre-fix behavior.
+  %xp = stablehlo.constant dense<[8.0, 27.0]> : tensor<2xf32>
+  %outp = stablehlo.constant dense<[2.0, 3.0]> : tensor<2xf32>
+  %expectedp = stablehlo.constant dense<[0.083333336, 0.037037037]> : tensor<2xf32>
+
+  %fwdp:2 = enzyme.fwddiff @cbrt(%xp, %dx) {
     activity=[#enzyme<activity enzyme_dup>],
     ret_activity=[#enzyme<activity enzyme_dup>]
   } : (tensor<2xf32>, tensor<2xf32>) -> (tensor<2xf32>, tensor<2xf32>)
 
-  check.expect_almost_eq %fwd#0, %out : tensor<2xf32>
-  check.expect_almost_eq %fwd#1, %expected : tensor<2xf32>
+  check.expect_almost_eq %fwdp#0, %outp : tensor<2xf32>
+  check.expect_almost_eq %fwdp#1, %expectedp : tensor<2xf32>
 
-  %rev:2 = enzyme.autodiff @cbrt(%x, %dx) {
+  %revp:2 = enzyme.autodiff @cbrt(%xp, %dx) {
     activity=[#enzyme<activity enzyme_active>],
     ret_activity=[#enzyme<activity enzyme_active>]
   } : (tensor<2xf32>, tensor<2xf32>) -> (tensor<2xf32>, tensor<2xf32>)
 
-  check.expect_almost_eq %rev#0, %out : tensor<2xf32>
-  check.expect_almost_eq %rev#1, %expected : tensor<2xf32>
+  check.expect_almost_eq %revp#0, %outp : tensor<2xf32>
+  check.expect_almost_eq %revp#1, %expectedp : tensor<2xf32>
+
+  // Negative inputs: the regression this PR fixes. Same gradient magnitude.
+  %xn = stablehlo.constant dense<[-8.0, -27.0]> : tensor<2xf32>
+  %outn = stablehlo.constant dense<[-2.0, -3.0]> : tensor<2xf32>
+  %expectedn = stablehlo.constant dense<[0.083333336, 0.037037037]> : tensor<2xf32>
+
+  %fwdn:2 = enzyme.fwddiff @cbrt(%xn, %dx) {
+    activity=[#enzyme<activity enzyme_dup>],
+    ret_activity=[#enzyme<activity enzyme_dup>]
+  } : (tensor<2xf32>, tensor<2xf32>) -> (tensor<2xf32>, tensor<2xf32>)
+
+  check.expect_almost_eq %fwdn#0, %outn : tensor<2xf32>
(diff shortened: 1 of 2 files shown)


kimjune01:
@wsmoses only red here is the A100 job, and it died with a Malt.TerminatedWorkerException in nn/convolution and nn/scatter_gather (looks like a cuDNN/OOM crash, unrelated to this cbrt gradient change). CPU and TPU are both green. mind kicking off a re-run when you get a sec?

test/lit_tests/diffrules/stablehlo/cbrt.mlir
```diff
@@ -8,27 +8,29 @@ func.func @cbrt(%x : tensor<2xf32>) -> tensor<2xf32> {
...
 // REVERSE-NEXT:  }
 
+// Negative inputs: cbrt is real and smooth for x < 0 (cbrt(-8) = -2), so the
+// true gradient 1/(3*cbrt(x)^2) is a finite real with the same magnitude as on
+// the positives. The old pow(x,-2/3) rule returned NaN here. See issue #2571.
 func.func @main() {
-  %x = stablehlo.constant dense<[8.0, 27.0]> : tensor<2xf32>
-  %out = stablehlo.constant dense<[2.0, 3.0]> : tensor<2xf32>
+  %x = stablehlo.constant dense<[-8.0, -27.0]> : tensor<2xf32>
+  %out = stablehlo.constant dense<[-2.0, -3.0]> : tensor<2xf32>
```
Antipath1: I do not think we should test only for negative numbers here. I think we should have both the previous case with positive numbers and the case with the negative numbers, so just add a second test to it.

Antipath1 (commented):
Except for the small thing in the test this looks good to me :) .

kimjune01:
Good call. `2f4998ac` keeps both: `@main` now runs the positive case (`[8, 27]`, unchanged from pre-fix) and the negative case (`[-8, -27]`, the #2571 regression) back to back. The symbolic FORWARD/REVERSE checks are input-independent so those stay as-is.

Antipath1 (approved):
LGTM thanks a lot :)

PR #2576, opened 2026-06-22
Guard LogSimplify rewrites against domain narrowing (#2570)

### LINKED ISSUE #2570: log(a*a) → 2*log(a) returns NaN for negative inputs
## Human section

I'm working with some dev tooling for generalizing tests, and I found these cases
involving LogSimplify that I believe will be a strict correctness improvement. If the
following analysis is valuable, I'd be happy to share the process I used to find the
bugs mechanically.

## Machine section

**Component:** `EnzymeHLOOptPass` (default `--enzyme-hlo-opt`)
**File:** `src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp`, `struct LogSimplify` (L27376)
**Class:** value-soundness. Optimized program returns NaN where unoptimized returns a finite real.

**Root cause:** four sub-cases apply a real-analysis identity valid only for positive bases,
with no domain guard. `CheckedOpRewritePattern` checks disable-attr + static-shape only
(`CheckedRewrite.h`). The original is real for the full input domain; the rewritten form is
real only for positive operands. Domain narrowed.

| # | line | rewrite | unsound when | example (f64) before to after |
|---|------|---------|--------------|------------------------------|
| 1 | 27406 | `log(a*a) -> 2*log(a)` | any `a<0` | a=-2: `log(4)=1.386` to `2*log(-2)=NaN` |
| 2 | 27391 | `log(pow(x,y)) -> y*log(x)` | `x<0` (even y), or `y=0` | y=0,x=-2: `log(pow(-2,0))=log(1)=0` to `0*log(-2)=NaN`; also y=2: `log(4)=1.386` to `NaN` |
| 3 | 27416 | `log(a*b) -> log(a)+log(b)` (one const) | const `<0` | b=-3,a=-2: `log(6)=1.792` to `NaN` |
| 4 | 27469 | `log(a/b) -> log(a)-log(b)` (one const) | const `<0` | b=-3,a=-2: `log(2/3)=-0.405` to `NaN` |

Note #1/#2 trigger on a runtime negative alone (no constant needed); `a*a` is always >= 0 so the
input `log(a*a)` is real for every `a != 0`, but `2*log(a)` is NaN for all `a<0`. #3/#4 need the
constant operand negative (positive constant is sound; both sides NaN for `x<0`).

**Why this is not a precision/fast-math question.** It is tempting to file these under the same
opt-in float policy as `log(sqrt(x)) -> 0.5*log(x)`, but they are a different class. That rewrite
is an exact identity in the reals and only perturbs float rounding/quantization, which is what the
opt-in flag is for. These four are *false in the reals themselves*: as partial functions the two
sides have different domains (`log(a*a)` is defined for `a != 0`, `2*log(a)` only for `a > 0`), so
the divergence appears before floating point is involved and no precision flag should sanction it.
That also gives a clean guard rule: a sub-case is sound exactly when the rewrite is a real-field
identity on the operand's sign, which for #3/#4 is precisely "the constant operand is non-negative"
(positive const: both sides agree, NaN-for-NaN on negatives; negative const: the rewrite
manufactures a NaN). This is the same finite-to-NaN domain-narrowing as the `CbrtOp` derivative in
#2571; both are the optimized form undefined where the original is real.

**Repro (case #1):**
```mlir
func.func @repro(%a: tensor<f64>) -> tensor<f64> {
  %sq = stablehlo.multiply %a, %a : tensor<f64>
  %r  = stablehlo.log %sq : tensor<f64>
  return %r : tensor<f64>
}
// enzymexlamlir-opt --enzyme-hlo-opt  ==>  2.0 * stablehlo.log(%a)
// %a = -2.0 :  unopt = log(4) = 1.3863   opt = 2*log(-2) = NaN
```
(The rewrite firing is already asserted by `test/lit_tests/logsimplify.mlir @main2`; that test
checks IR text only and never evaluates, which is why the value divergence is invisible to it.)

**Fix:** gate each sub-case on a provably-non-negative base (reuse the non-negativity analysis
already used by `AbsPositiveSimplify` / `guaranteedNonNegativeResult`). Per case:

- #1 `log(a*a) -> 2*log(abs(a))` is sound unconditionally (one extra `abs`, `a*a` and `abs(a)`
  share a domain), so it needs no guard.
- #2 fire only when `x` is provably non-negative (or `pow`'s result is).
- #3/#4 fire only when the constant operand is non-negative.

The `abs` form for #1 fixes the domain (finite-to-NaN) issue but is not bit-exact under the
multiply's overflow/underflow/rounding, so it still belongs under the float-rewrite policy as a
precision matter, distinct from the domain bug above.

**Scope:** These four are the only rewrites in this file that turn a finite real into NaN (a
domain-guard bug). Differences that are purely overflow/underflow/rounding are a separate
fast-math question and I set them aside, including the same log rewrites at extreme magnitudes and
`log(a+a) -> log(2)+log(a)`. At the domain level `log(sqrt)/log(cbrt)/log(rsqrt)`,
`SquareAbsSimplify`, `SignAbsSimplify`, `PowSimplify`, `DivideSqrtToMultiplyRsqrt`, multiply/power
(`ChainedMultiplyToPower`; `pow(-2,3)` returns `-8`, the unsoundness hypothesis did not reproduce),
and the `NoNan*` patterns are sound.

**Acceptance:** after the fix, for `a<0` the optimized and unoptimized programs agree (both compute
`log(a*a)` correctly, i.e. a finite real).

kimjune01:
Fixes #2570 (maintainer-approved: "makes sense, PR welcome!").

## Problem

Four `LogSimplify` sub-cases (`EnzymeHLOOpt.cpp`) apply a real-analysis identity valid only for positive bases, with no domain guard, so the optimized program returns NaN where the unoptimized one returns a finite real:

| rewrite | unsound when | example (f64) |
|---|---|---|
| `log(a*a) -> 2*log(a)` | any `a < 0` | a=-2: `log(4)=1.386` becomes `2*log(-2)=NaN` |
| `log(pow(x,y)) -> y*log(x)` | `x < 0` (even y), or `y = 0` | y=0,x=-2: `log(1)=0` becomes `0*log(-2)=NaN` |
| `log(a*b) -> log(a)+log(b)` (one const) | const `< 0` | b=-3,a=-2: `log(6)=1.79` becomes `NaN` |
| `log(a/b) -> log(a)-log(b)` (one const) | const `< 0` | b=-3,a=-2: `log(2/3)=-0.405` becomes `NaN` |

This is **not** the same as the opt-in float policy behind `log(sqrt(x)) -> 0.5*log(x)`. That rewrite is an exact identity in the reals and only perturbs float rounding. These four are false in the reals themselves: as partial functions the two sides have different domains (`log(a*a)` is real for `a != 0`, `2*log(a)` only for `a > 0`), so the divergence appears before floating point is involved. Same finite-to-NaN domain narrowing as the `CbrtOp` derivative in #2571 / #2575.

## Fix

Reuse the existing `guaranteedNonNegativeResult` analysis to gate each case on a provably-non-negative base:

- `log(a*a) -> 2*log(abs(a))`: unconditionally domain-sound, since `a*a` and `abs(a)` share a domain. No guard needed; just take the log of `|a|`.
- `log(pow(x,y))`: fire only when `x` is provably non-negative.
- `log(a*b)` / `log(a/b)`: fire only when the constant operand is non-negative. A positive constant agrees (NaN-for-NaN) on negative inputs; a negative constant manufactures the NaN.

## Test

`test/lit_tests/logsimplify.mlir`'s `@main2`/`@main6` (the `log(a*a)` cases) are updated to the new `abs` form, and three cases are added asserting the guarded rewrites no longer fire (negative-constant mul, negative-constant div, unknown-sign `pow` base). The pipeline enables only `log_simplify;log_const_prop`, so the positive-constant cases (`@main3`/`@main4`/`@main7`/`@main8`) still fire unchanged.

> Notes:
> - I cannot build Enzyme-JAX locally, so the updated `CHECK` blocks are a best-effort prediction of the post-pass IR (operand order, SSA numbering, the inserted `stablehlo.abs`). If CI's FileCheck flags a mismatch I will push a fixup; the source guards are the substantive change.
> - The `abs` form for `log(a*a)` fixes the domain (finite-to-NaN) bug but is not bit-exact under the multiply's overflow/underflow/rounding, so it remains a precision matter under the float-rewrite policy, distinct from the domain bug fixed here. Happy to gate it the same way as the others instead if you'd prefer.

Found mechanically with a value-and-gradient soundness gate over the rewrite library; happy to share the process if useful.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -27390,7 +27390,11 @@ struct LogSimplify final
 
     { // log(pow(x, y)) -> y * log(x)
       auto defOp = op.getOperand().getDefiningOp<stablehlo::PowOp>();
-      if (defOp) {
+      // Only sound when the base x is provably non-negative. For x < 0,
+      // log(pow(x, y)) can be a finite real (e.g. even integer y, where
+      // pow(x, y) > 0) while y * log(x) is NaN; likewise y = 0 gives
+      // log(pow(x, 0)) = 0 but 0 * log(x) = NaN. See issue #2570.
+      if (defOp && guaranteedNonNegativeResult(defOp.getLhs(), rewriter)) {
         rewriter.replaceOpWithNewOp<stablehlo::MulOp>(
             op, defOp.getRhs(),
             stablehlo::LogOp::create(rewriter, op.getLoc(), defOp.getLhs()));
@@ -27403,23 +27407,35 @@ struct LogSimplify final
       if (defOp) {
         auto lhs = defOp.getLhs();
         auto rhs = defOp.getRhs();
-        if (lhs == rhs) { // log(mul(a, a)) -> 2 * log(a)
+        if (lhs == rhs) { // log(mul(a, a)) -> 2 * log(abs(a))
+          // a*a is non-negative for every real a, but 2 * log(a) is NaN for
+          // a < 0. Take the log of |a| so the rewrite keeps log(a*a)'s domain
+          // (log(a*a) = 2 * log|a|). See issue #2570.
           rewriter.replaceOpWithNewOp<stablehlo::MulOp>(
               op,
               stablehlo::ConstantOp::create(
                   rewriter, op.getLoc(), lhs.getType(),
                   cast<ElementsAttr>(makeAttr(lhs.getType(), 2))),
-              stablehlo::LogOp::create(rewriter, op.getLoc(), lhs));
+              stablehlo::LogOp::create(
+                  rewriter, op.getLoc(),
+                  stablehlo::AbsOp::create(rewriter, op.getLoc(),
+                                           lhs.getType(), lhs)));
           return success();
         }
 
         if (anyOperandIsConstant(defOp) &&
             !allOperandsAreConstant(defOp)) { // log(mul(a, b)) -> log(a) +
                                               // log(b) if a or b is constant
-          rewriter.replaceOpWithNewOp<stablehlo::AddOp>(
-              op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
-              stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
-          return success();
+          // Only sound when the constant operand is non-negative: a negative
+          // constant makes log(const) NaN where log(a*b) was a finite real
+          // (a < 0 makes both sides NaN, so they still agree). Issue #2570.
+          Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
+          if (guaranteedNonNegativeResult(cst, rewriter)) {
+            rewriter.replaceOpWithNewOp<stablehlo::AddOp>(
+                op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
+                stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
+            return success();
+          }
         }
       }
     }
@@ -27469,10 +27485,16 @@ struct LogSimplify final
         if (anyOperandIsConstant(defOp) &&
             !allOperandsAreConstant(defOp)) { // log(div(a, b)) -> log(a) -
                                               // log(b) if a or b is constant
-          rewriter.replaceOpWithNewOp<stablehlo::SubtractOp>(
-              op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
-              stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
-          return success();
+          // Only sound when the constant operand is non-negative: a negative
+          // constant makes log(const) NaN where log(a/b) was a finite real.
+          // Issue #2570.
+          Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
+          if (guaranteedNonNegativeResult(cst, rewriter)) {
+            rewriter.replaceOpWithNewOp<stablehlo::SubtractOp>(
+                op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
+                stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
+            return success();
+          }
         }
       }
     }
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -27743,6 +27743,25 @@ struct CommonAssociativeCommutativeOpReorder final
   }
 };
 
+static bool isNonNegativeConstant(Value value) {
+  DenseElementsAttr attr;
+  if (!matchPattern(value, m_Constant(&attr)) ||
+      !isa<FloatType>(attr.getElementType()))
+    return false;
+
+  return llvm::all_of(attr.getValues<APFloat>(), [](const APFloat &element) {
+    return !element.isNaN() && !element.isNegative();
+  });
+}
+
+static bool isStrictlyPositiveConstant(Value value) {
+  DenseElementsAttr attr;
+  return isNonNegativeConstant(value) &&
+         matchPattern(value, m_Constant(&attr)) &&
+         llvm::none_of(attr.getValues<APFloat>(),
+                       [](const APFloat &element) { return element.isZero(); });
+}
+
 struct LogSimplify final
     : public CheckedOpRewritePattern<stablehlo::LogOp, LogSimplify> {
   using CheckedOpRewritePattern<stablehlo::LogOp,
@@ -27754,20 +27773,6 @@ struct LogSimplify final
       auto defOp = op.getOperand().getDefiningOp<stablehlo::ExpOp>();
       if (defOp) {
         rewriter.replaceAllUsesWith(op.getResult(), defOp.getOperand());
-        return success();
-      }
-    }
-
-    { // log(pow(x, y)) -> y * log(x)
-      auto defOp = op.getOperand().getDefiningOp<stablehlo::PowOp>();
-      // Only sound when the base x is provably non-negative. For x < 0,
-      // log(pow(x, y)) can be a finite real (e.g. even integer y, where
-      // pow(x, y) > 0) while y * log(x) is NaN; likewise y = 0 gives
-      // log(pow(x, 0)) = 0 but 0 * log(x) = NaN. See issue #2570.
-      if (defOp && guaranteedNonNegativeResult(defOp.getLhs(), rewriter)) {
-        rewriter.replaceOpWithNewOp<stablehlo::MulOp>(
-            op, defOp.getRhs(),
-            stablehlo::LogOp::create(rewriter, op.getLoc(), defOp.getLhs()));
         return success();
       }
     }
@@ -27796,11 +27801,11 @@ struct LogSimplify final
         if (anyOperandIsConstant(defOp) &&
             !allOperandsAreConstant(defOp)) { // log(mul(a, b)) -> log(a) +
                                               // log(b) if a or b is constant
-          // Only sound when the constant operand is non-negative: a negative
-          // constant makes log(const) NaN where log(a*b) was a finite real
-          // (a < 0 makes both sides NaN, so they still agree). Issue #2570.
+          // Require a strictly positive constant to avoid narrowing the
+          // domain for negative or zero constants. This does not make the
+          // rewrite bit-exact under floating-point rounding. See issue #2570.
           Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
-          if (guaranteedNonNegativeResult(cst, rewriter)) {
+          if (isStrictlyPositiveConstant(cst)) {
             rewriter.replaceOpWithNewOp<stablehlo::AddOp>(
                 op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
                 stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
@@ -27855,11 +27860,15 @@ struct LogSimplify final
         if (anyOperandIsConstant(defOp) &&
             !allOperandsAreConstant(defOp)) { // log(div(a, b)) -> log(a) -
                                               // log(b) if a or b is constant
-          // Only sound when the constant operand is non-negative: a negative
-          // constant makes log(const) NaN where log(a/b) was a finite real.
-          // Issue #2570.
-          Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
-          if (guaranteedNonNegativeResult(cst, rewriter)) {
+          // A constant numerator must be strictly positive; a zero numerator
+          // can turn a finite log(-0) into NaN after splitting. A zero
+          // denominator is safe here, so it only needs to be non-negative.
+          // This does not make the rewrite bit-exact under floating-point
+          // rounding. See issue #2570.
+          bool constantIsLhs = matchPattern(lhs, m_Constant());
+          Value cst = constantIsLhs ? lhs : rhs;
+          if ((constantIsLhs && isStrictlyPositiveConstant(cst)) ||
+              (!constantIsLhs && isNonNegativeConstant(cst))) {
             rewriter.replaceOpWithNewOp<stablehlo::SubtractOp>(
                 op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
                 stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -27780,11 +27780,9 @@ struct LogSimplify final
       if (defOp) {
         auto lhs = defOp.getLhs();
         auto rhs = defOp.getRhs();
-        // log(mul(a, a)) -> 2 * log(a) when a is provably non-negative.
-        // For a < 0, 2 * log(a) is NaN even though log(a*a) is finite, so we
-        // only fire this when a is provably non-negative. Gating on the
-        // non-negative analysis keeps the rewrite correct without inserting an
-        // abs (which may be slower than the mul). See issue #2570.
+        // log(mul(a, a)) -> 2 * log(a). Only when a >= 0: for a < 0, 2*log(a)
+        // is NaN while log(a*a) is finite. Gating on the analysis avoids an abs
+        // (possibly slower than the mul). See #2570.
         if (lhs == rhs && guaranteedNonNegativeResult(lhs, rewriter)) {
           rewriter.replaceOpWithNewOp<stablehlo::MulOp>(
               op,
@@ -27798,9 +27796,9 @@ struct LogSimplify final
         if (anyOperandIsConstant(defOp) &&
             !allOperandsAreConstant(defOp)) { // log(mul(a, b)) -> log(a) +
                                               // log(b) if a or b is constant
-          // Require a strictly positive constant to avoid narrowing the
-          // domain for negative or zero constants. This does not make the
-          // rewrite bit-exact under floating-point rounding. See issue #2570.
+          // Split only for a strictly positive constant; negative/zero narrows
+          // the domain, and the split isn't bit-exact under FP rounding. See
+          // #2570.
           Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
           if (isPositiveNonNanConstant(cst, /*allowZero=*/false)) {
             rewriter.replaceOpWithNewOp<stablehlo::AddOp>(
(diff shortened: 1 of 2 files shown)


Antipath1:
//test/lit_tests:logsimplify.mlir.test is failing in CI

kimjune01:
Lit test fixed in `ba6fd786`. The remaining red lanes (ct6e TPU, a100 GPU, macOS) are failing on `main` too, so they're not from this change.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -27760,7 +27760,11 @@ struct LogSimplify final
 
     { // log(pow(x, y)) -> y * log(x)
       auto defOp = op.getOperand().getDefiningOp<stablehlo::PowOp>();
-      if (defOp) {
+      // Only sound when the base x is provably non-negative. For x < 0,
+      // log(pow(x, y)) can be a finite real (e.g. even integer y, where
+      // pow(x, y) > 0) while y * log(x) is NaN; likewise y = 0 gives
+      // log(pow(x, 0)) = 0 but 0 * log(x) = NaN. See issue #2570.
+      if (defOp && guaranteedNonNegativeResult(defOp.getLhs(), rewriter)) {
```
Antipath1: The comment here claims that this would fix the log(pow(x, 0)) case, but it actually does not. In the specific case where $x = 0$ and $y = 0$:If we start with 
$\log(\text{pow}(0, 0)) = \log(1) = 0$,
our pattern rewrites it into $0 \cdot \log(0)$, 
which evaluates to NaN (because $\log(0)$ is $-\infty$). That is mathematically incorrect. Soooooooooooooo we still have a correctness bug here, actually. Proving that $x$ or $y$ is never $0$ is kind of hard, so we might genuinely need to put that behind a flag, or just outright remove this rewrite, or try to prove that $x$ is strictly greater than zero (which, I might be wrong, is only provable in very specific scenarios, like the definingOp of $x$ or $y$ being an exponential op which can never be $0$). Nevertheless, I don't think you need to fix this in this PR, since having guaranteedNonNegativeResult is already a massive improvement. So, I think I just need you to update the comment to include this info.

(cc @wsmoses I think this needs your authority on how to deal with this)

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -27773,23 +27777,35 @@ struct LogSimplify final
...
+          // Only sound when the constant operand is non-negative: a negative
+          // constant makes log(const) NaN where log(a*b) was a finite real
+          // (a < 0 makes both sides NaN, so they still agree). Issue #2570.
+          Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
+          if (guaranteedNonNegativeResult(cst, rewriter)) {
+            rewriter.replaceOpWithNewOp<stablehlo::AddOp>(
+                op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
+                stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
+            return success();
+          }
```
Antipath1: Fails when the constant is exactly 0 and the variable is negative. Because guaranteedNonNegativeResult allows the constant to be exactly 0
if cst = 0 and the variable is -2:
Original: $\log(0 \cdot -2) \rightarrow \log(-0.0) \rightarrow \mathbf{-\infty}$ 
Optimized: $\log(0) + \log(-2) \rightarrow -\infty + \text{NaN} \rightarrow \mathbf{\text{NaN}}$

Similar problem as before, and I still think this should not necessarily block the PR since the status quo is actually a good bit worse.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -27839,10 +27855,16 @@ struct LogSimplify final
...
+          // Only sound when the constant operand is non-negative: a negative
+          // constant makes log(const) NaN where log(a/b) was a finite real.
+          // Issue #2570.
+          Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
+          if (guaranteedNonNegativeResult(cst, rewriter)) {
+            rewriter.replaceOpWithNewOp<stablehlo::SubtractOp>(
+                op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
+                stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
+            return success();
+          }
```
Antipath1: Same problem and still do not think this should be blocking

Fails when the numerator ($a$) is the constant 0, and the denominator ($b$) is negative.The code allows the constant to be either the LHS ($a$) or the RHS ($b$). If the constant happens to be the LHS, and it equals 0, we hit the same trap. 
 $a = 0$ and $b = -2$:
Original: $\log(0 / -2) \rightarrow \log(-0.0) \rightarrow \mathbf{-\infty}$
Optimized: $\log(0) - \log(-2) \rightarrow -\infty - \text{NaN} \rightarrow \mathbf{\text{NaN}}$(Note: If the denominator $b$ is the constant 0, and $a$ is negative, both sides correctly evaluate to NaN. The bug only exists when the numerator is 0).

Antipath1 (changes requested):
I think this improves the status quo quite a lot thank you so I think that this should actually get merged even in this state but I think the comments should be updated to reflect the remaining problems.

kimjune01:
Good catches, thanks. I tightened the patch:

- removed the `log(pow(x, y))` rewrite, since non-negativity isn’t enough;
- require strictly positive constants for the multiply rewrite and for a constant division numerator, while preserving the safe zero-denominator case;
- added regressions for the zero and `pow` cases.

But even with positive constants, the multiply/divide identities aren’t fully IEEE-equivalent because rounding, overflow, and underflow can differ. It may be out of scope. If required, those rewrites may need separate gating or removal.

wsmoses:
@Antipath1 and I spoke about this during the meeting today.

@kimjune01 rather than folding to an abs (which may be slower than the mul), what if you gated the transformation on the result of the is value positive analysis

kimjune01:
Done -- dropped the `abs` and gated on `guaranteedNonNegativeResult` as suggested. It now folds `log(a*a) -> 2*log(a)` only when `a` is provably non-negative (the `a*a`, `abs`, `exp`/`sqrt` producers), and leaves `log(a*a)` untouched otherwise. Added positive + negative lit coverage.

While wiring this up I hit a latent soundness bug in that analysis's square rule -- it compares defining ops, so `mul(%argN, %argM)` of two distinct block args gets `null == null` and is wrongly proven non-negative. Filed separately as #2648 to keep this PR focused; the gate here is correct against the analysis contract and doesn't regress current behavior.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -27786,10 +27794,16 @@ struct LogSimplify final
...
                                               // log(b) if a or b is constant
-          rewriter.replaceOpWithNewOp<stablehlo::AddOp>(
-              op, stablehlo::LogOp::create(rewriter, op.getLoc(), lhs),
-              stablehlo::LogOp::create(rewriter, op.getLoc(), rhs));
-          return success();
+          // Require a strictly positive constant to avoid narrowing the
+          // domain for negative or zero constants. This does not make the
+          // rewrite bit-exact under floating-point rounding. See issue #2570.
+          Value cst = matchPattern(lhs, m_Constant()) ? lhs : rhs;
+          if (isValidLogSplitConstant(cst, /*allowZero=*/false)) {
```
wsmoses: can you rename this from isValidLogSplitConstant to isPositiveFiniteConstant for clarity

wsmoses (approved):
lgtm modulo the fn name for simplicity

kimjune01:
Renamed to `isPositiveFiniteConstant` in `aa065977`.

One nuance on the name so it doesn't mislead later: the predicate currently accepts `+inf` (it rejects NaN and negatives but not infinity), and in the division-denominator case it's called with `allowZero=true`, so there it's really "non-negative finite-or-inf" rather than strictly positive. Happy to tighten the body to match the name (reject `+inf`, and/or split the zero-allowing path) if you'd like -- just flagging that it's a behavior change rather than folding it into the rename.

wsmoses:
maybe the isPositiveNonNanConstant then?

kimjune01:
Good call -- `isPositiveNonNanConstant` it is (`b76081e1`). That name is honest about what the predicate checks (rejects NaN and negatives, doesn't assert finiteness).

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -27750,6 +27750,19 @@ struct LogSimplify final
 
   LogicalResult matchAndRewriteImpl(stablehlo::LogOp op,
                                     PatternRewriter &rewriter) const {
+    auto isPositiveNonNanConstant = [](Value value, bool allowZero) {
```
wsmoses: sorry one final minor comment, can you make this a static member function rather than a lambda?

it'll be easier to debug in the future [and faster to compile/run]

kimjune01:
Done in `51ec6b63` -- `isPositiveNonNanConstant` is now a `static` member of `LogSimplify` instead of a local lambda.

Antipath1 (approved):
lgtm thank you :)

wsmoses:
ci fails:
```

[17,048 / 18,101] 16 / 1069 tests; Testing //test/lit_tests:mpi_comm_rank.mlir.test; 17s remote-cache, processwrapper-sandbox ... (32 actions running)
[17,050 / 18,101] 18 / 1069 tests; Testing //test/lit_tests:dus_to_pad.mlir.test; 18s remote-cache, processwrapper-sandbox ... (32 actions running)
[17,050 / 18,101] 18 / 1069 tests; Testing //test/lit_tests:dus_to_pad.mlir.test; 19s remote-cache, processwrapper-sandbox ... (32 actions running)
[17,053 / 18,101] 21 / 1069 tests; Testing //test/lit_tests:dus_to_pad.mlir.test; 20s remote-cache, processwrapper-sandbox ... (32 actions running)
FAIL: //test/lit_tests:logsimplify.mlir.test (see /root/.bazel/execroot/__main__/bazel-out/aarch64-opt/testlogs/test/lit_tests/logsimplify.mlir.test/test.log)
INFO: From Testing //test/lit_tests:logsimplify.mlir.test:
==================== Test output for //test/lit_tests:logsimplify.mlir.test:
-- Testing: 1 tests, 1 workers --
FAIL: Enzyme-JaX :: lit_tests/logsimplify.mlir (1 of 1)
******************** TEST 'Enzyme-JaX :: lit_tests/logsimplify.mlir' FAILED ********************
Exit Code: 1

Command Output (stderr):
--
enzymexlamlir-opt /root/.bazel/sandbox/processwrapper-sandbox/6746/execroot/__main__/bazel-out/aarch64-opt/bin/test/lit_tests/logsimplify.mlir.test.runfiles/__main__/test/lit_tests/logsimplify.mlir --pass-pipeline="builtin.module(enzyme-hlo-generate-td{patterns=log_simplify;log_const_prop},transform-interpreter,enzyme-hlo-remove-transform)" | FileCheck /root/.bazel/sandbox/processwrapper-sandbox/6746/execroot/__main__/bazel-out/aarch64-opt/bin/test/lit_tests/logsimplify.mlir.test.runfiles/__main__/test/lit_tests/logsimplify.mlir # RUN: at line 1
+ enzymexlamlir-opt /root/.bazel/sandbox/processwrapper-sandbox/6746/execroot/__main__/bazel-out/aarch64-opt/bin/test/lit_tests/logsimplify.mlir.test.runfiles/__main__/test/lit_tests/logsimplify.mlir '--pass-pipeline=builtin.module(enzyme-hlo-generate-td{patterns=log_simplify;log_const_prop},transform-interpreter,enzyme-hlo-remove-transform)'
+ FileCheck /root/.bazel/sandbox/processwrapper-sandbox/6746/execroot/__main__/bazel-out/aarch64-opt/bin/test/lit_tests/logsimplify.mlir.test.runfiles/__main__/test/lit_tests/logsimplify.mlir
/root/.bazel/sandbox/processwrapper-sandbox/6746/execroot/__main__/bazel-out/aarch64-opt/bin/test/lit_tests/logsimplify.mlir.test.runfiles/__main__/test/lit_tests/logsimplify.mlir:85:16: error: CHECK-NEXT: is not on the line after the previous match
// CHECK-NEXT: %0 = stablehlo.multiply %arg0, %arg0 : tensor<f64>
               ^
<stdin>:55:2: note: 'next' match was here
 %0 = stablehlo.multiply %arg0, %arg0 : tensor<f64>
 ^
<stdin>:30:61: note: previous match ended here
 %cst = stablehlo.constant dense<2.000000e+00> : tensor<f64>
                                                            ^
<stdin>:31:1: note: non-matching line after previous match is here
 %0 = stablehlo.multiply %arg0, %arg0 {enzymexla.non_negative = [#enzymexla<guaranteed GUARANTEED>]} : tensor<f64>
^

Input file: <stdin>
Check file: /root/.bazel/sandbox/processwrapper-sandbox/6746/execroot/__main__/bazel-out/aarch64-opt/bin/test/lit_tests/logsimplify.mlir.test.runfiles/__main__/test/lit_tests/logsimplify.mlir

-dump-input=help explains the following input dump.

Input was:
<<<<<<
           .
           .
           .
          25:  %0 = stablehlo.power %cst, %arg0 : tensor<f64> 
          26:  %1 = stablehlo.log %0 : tensor<f64> 
          27:  return %1 : tensor<f64> 
          28:  } 
          29:  func.func @main2(%arg0: tensor<f64>) -> tensor<f64> { 
          30:  %cst = stablehlo.constant dense<2.000000e+00> : tensor<f64> 
next:85'0                                                                {   search range start (exclusive)
          31:  %0 = stablehlo.multiply %arg0, %arg0 {enzymexla.non_negative = [#enzymexla<guaranteed GUARANTEED>]} : tensor<f64> 
          32:  %1 = stablehlo.log %0 : tensor<f64> 
          33:  %2 = stablehlo.multiply %cst, %1 : tensor<f64> 
          34:  return %2 : tensor<f64> 
          35:  } 
           .
           .
           .
          50:  %0 = stablehlo.log %arg0 : tensor<f64> 
          51:  %1 = stablehlo.add %cst, %0 : tensor<f64> 
          52:  return %1 : tensor<f64> 
          53:  } 
          54:  func.func @main6(%arg0: tensor<f64>) -> tensor<f64> { 
          55:  %0 = stablehlo.multiply %arg0, %arg0 : tensor<f64> 
next:85'1      !~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~   error: match on wrong line
          56:  %1 = stablehlo.log %0 : tensor<f64> 
          57:  return %1 : tensor<f64> 
          58:  } 
          59:  func.func @main7(%arg0: tensor<f64>) -> tensor<f64> { 
          60:  %cst = stablehlo.constant dense<1.3862943611198906> : tensor<f64> 
           .
           .
           .
         109:  %1 = stablehlo.log %0 : tensor<f64> 
         110:  %2 = stablehlo.multiply %cst, %1 : tensor<f64> 
//test/lit_tests:while_licm.mlir.test                                    PASSED in 0.2s
//test/lit_tests:while_noop.mlir.test                                    PASSED in 0.2s
//test/lit_tests:while_pad_induction_reduction.mlir.test                 PASSED in 0.2s
//test/lit_tests:while_repeated_induction_reduction.mlir.test            PASSED in 0.2s
//test/lit_tests:while_simplify.mlir.test                                PASSED in 0.2s
//test/lit_tests:whileconcat.mlir.test                                   PASSED in 0.2s
//test/lit_tests:whiledeadarg.mlir.test                                  PASSED in 0.3s
//test/lit_tests:whiledus.mlir.test                                      PASSED in 0.2s
//test/lit_tests:whiledus2.mlir.test                                     PASSED in 0.2s
//test/lit_tests:whileupdatewithoutcorners.mlir.test                     PASSED in 0.2s
//test/lit_tests:whilewrap.mlir.test                                     PASSED in 0.2s
//test/lit_tests:widen_extend.mlir.test                                  PASSED in 0.2s
//test/lit_tests:widen_wrap.mlir.test                                    PASSED in 0.2s
//test/lit_tests:widen_wrap_reshape.mlir.test                            PASSED in 0.2s
//test/lit_tests:wrap_const.mlir.test                                    PASSED in 0.2s
//test/lit_tests:wrap_unary_elementwise.mlir.test                        PASSED in 0.2s
//test/lit_tests:wrapelementwise.mlir.test                               PASSED in 0.2s
//test/lit_tests:xlawrapper_alloca.mlir.test                             PASSED in 0.2s
//test/lit_tests:xor.mlir.test                                           PASSED in 0.2s
//test/lit_tests:zerobasepowsimplify.mlir.test                           PASSED in 0.2s
//test/lit_tests:zeroproductreshapepad.mlir.test                         PASSED in 0.2s
//test/lit_tests:logsimplify.mlir.test                                   FAILED in 0.9s
  /root/.bazel/execroot/__main__/bazel-out/aarch64-opt/testlogs/test/lit_tests/logsimplify.mlir.test/test.log
  ```

kimjune01:
Fixed in `48aaf2b1`. The failure was the `main2` CHECK block: the non-negativity gate stamps `enzymexla.non_negative` on the inner `arg0*arg0` when the `log(mul(sq,sq))` rewrite queries the analysis, so FileCheck skipped past it to `main6`'s unannotated multiply and flagged a wrong-line match. Updated the CHECK-NEXT to match the annotated op. `main6` is untouched (there `arg0` itself is queried, not the multiply, so no stamp). Only the `n2`/build lit lane was real; the TPU/GPU/macOS lanes are red on `main` too.

kimjune01:
Also excluded infinite constants from the splits.

PR #2583, opened 2026-06-24
More type casting logic for aggregates

jessicacotturone21:
Adds ability to type cast from more complex aggregates, including those with multiple levels of nested structs and arrays, by recursively unwrapping the aggregates. This should prevent the code from crashing in these cases when running `polygeist-mem2reg`.

### DIFF
--- a/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
+++ b/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
@@ -1080,22 +1080,30 @@ void removeRedundantBlockArgs(
   }
 }
 
-// Recursively extract all scalar leaves from a nested struct/array aggregate
-static void collectLeaves(OpBuilder &b, Value val,
-                          SmallVectorImpl<Value> &leaves) {
+// Recursively extract sub-values, stopping as soon as a node matches the target
+// type. Falls all the way to scalar leaves when no intermediate match is found.
+static void collectSubvalues(OpBuilder &b, Value val,
+                             SmallVectorImpl<Value> &subvalues,
+                             Type targetType) {
   Type t = val.getType();
+
+  if (t == targetType) {
+    subvalues.push_back(val);
+    return;
+  }
+
   if (auto ST = dyn_cast<LLVM::LLVMStructType>(t)) {
     for (unsigned i = 0; i < ST.getBody().size(); i++) {
       Value extracted = LLVM::ExtractValueOp::create(b, val.getLoc(), val, i);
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
```diff
@@ -1122,29 +1122,34 @@ Value castToType(Type elType, Value val, Operation *op) {
...
-      }
-    }
-  } else if (auto AT = dyn_cast<LLVM::LLVMArrayType>(val.getType())) {
-    Type elemType = AT.getElementType();
-    if (isa<VectorElementTypeInterface>(elemType)) {
-      Value vec = buildVector(b, val, AT.getNumElements(), elemType);
-      return castToType(elType, vec, op);
+  } else if (isa<LLVM::LLVMStructType, LLVM::LLVMArrayType>(val.getType())) {
+    SmallVector<Value> leaves;
+    collectLeaves(b, val, leaves);
```
wsmoses, line 1132: you may not want/need to go all the way to the leaves.

For example you could have 

struct A {
   B
}

struct B {
   [100000x i8]
}

you could leave wise reconstruct them.

but also casting an A -> B should just be a single extractvalue

jessicacotturone21: This is a great point -- I have adjusted to code to address this case

test/lit_tests/mem2reg_type_casting.mlir
```diff
@@ -208,4 +209,74 @@ llvm.func @forward_vector_to_int_bitcast(%val: vector<4xi8>) -> i32 {
...
+// CHECK-NEXT: %[[C1:.*]] = llvm.mlir.constant(1 : i32) : i32
+// CHECK-NEXT: %[[EV:.*]] = llvm.extractvalue %[[VAL]][0] : !llvm.struct<(struct<(array<100 x i8>)>)>
+// CHECK-NEXT: llvm.return %[[EV]] : !llvm.struct<(array<100 x i8>)>
+// CHECK-NEXT: }
+
+// -----
+
+// Nested array stored, inner array loaded
+
+llvm.func @forward_outer_to_inner_array(%val: !llvm.array<1 x array<100 x i8>>) -> !llvm.array<100 x i8> {
```
wsmoses: just for ease lets change 100 to something absurd like 100000000 [to make sure it doesnt take forever and do the bad compile time thing]

wsmoses (approved):
lgtm minus a minor comment

PR #2589, opened 2026-06-26
feat: fold concat of reverse slices into stablehlo.reverse

### LINKED ISSUE #1409: `concat(a[n-1:n], a[n-2:n-1], ...., a[0:1])` to `reverse`
```
%19 = stablehlo.slice %18 [0:1, 0:3, 0:10] : (tensor<5x3x10xf32>) -> tensor<1x3x10xf32>
    %20 = stablehlo.slice %18 [4:5, 0:3, 0:10] : (tensor<5x3x10xf32>) -> tensor<1x3x10xf32>
    %21 = stablehlo.slice %18 [3:4, 0:3, 0:10] : (tensor<5x3x10xf32>) -> tensor<1x3x10xf32>
    %22 = stablehlo.slice %18 [2:3, 0:3, 0:10] : (tensor<5x3x10xf32>) -> tensor<1x3x10xf32>
    %23 = stablehlo.slice %18 [1:2, 0:3, 0:10] : (tensor<5x3x10xf32>) -> tensor<1x3x10xf32>
    %24 = stablehlo.concatenate %20, %21, %22, %23, %19, dim = 0 : (tensor<1x3x10xf32>, tensor<1x3x10xf32>, tensor<1x3x10xf32>, tensor<1x3x10xf32>, tensor<1x3x10xf32>) -> tensor<5x3x10xf32>
```

amirjon-1:
Adds ConcatSlicesToReverse canonicalization pattern in EnzymeHLOOpt.cpp 
that detects N unit-stride slices of the same tensor concatenated in 
reverse order and replaces the sequence with a single stablehlo.reverse op.

Changes:
- EnzymeHLOOpt.cpp: added ConcatSlicesToReverse pattern, registered 
  alongside ConcatFuse and ConcatToBroadcast
- test/lit_tests/concat_slices_to_reverse.mlir: 3 test cases (3D positive, 
  1D positive, negative/no-fold case)

Fixes #1409

### DIFF
--- a/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
+++ b/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
@@ -388,6 +388,7 @@ static void addBaseTransformPasses(std::vector<std::string> &list,
   list.push_back("slice_elementwise<1>");
   list.push_back("dot_reshape_dot<1>");
   list.push_back("concat_fuse<1>");
+  list.push_back("concat_slices_to_reverse<1>");
   list.push_back("concat_push_binop_add<1>");
   list.push_back("concat_push_binop_mul<1>");
   list.push_back("reduce_concat<1>");
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -6219,7 +6219,8 @@ struct ConcatFuse final
   }
 };
 
-// concat(slice(src,[N-1:N,...]), slice(src,[N-2:N-1,...]), ..., slice(src,[0:1,...]), dim=D)
+// concat(slice(src,[N-1:N,...]), slice(src,[N-2:N-1,...]), ...,
+// slice(src,[0:1,...]), dim=D)
 // -> reverse(src, dims=[D])
 struct ConcatSlicesToReverse final
     : CheckedOpRewritePattern<stablehlo::ConcatenateOp, ConcatSlicesToReverse> {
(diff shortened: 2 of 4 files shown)


jumerckx (commented):
Very cool, this looks good to me! Can you run clang-format?

As a small nit, I think you could do all the validity checks within the first loop over the operands? That way there is no need for a `slices` SmallVector, the pattern can fail from the moment one of the operands does not fit? Not entirely sure.

Would also be good to add tests for other pattern failure case instead of only wrong order.

amirjon-1:
@jumerckx 
Hey, thank you for the feedback! Made the following changes:
- Applied clang format 
- Refactored to a single pass loop (no more SmallVector)
- Added 3 more negative test cases. 

Let me know what you think!

jumerckx:
Great! One last thing I'd forgotten about, new RewritePatterns should also be added to `TransformOps.td` and in the list in `src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp` (in this case it should be added in `addBaseTransformPasses`, I believe).
A bit hidden but at some point we tried documenting these steps in `DEVDOCS.md`.

I think there might be a version mismatch or something else going on with the formatter because CI doesn't seem to agree: https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/28272486169/job/83800957370?pr=2589

amirjon-1:
Thanks @jumerckx! I added ConcatSlicesToReverse to TransformOps.td and addBaseTransformPasses in EnzymeXLA.cpp.

Also it looked like I had a different version of clang format locally vs CIs pinned v18, so I reformatted with the correct version. Should be all set now.

jumerckx (approved):
Great!

PR #2590, opened 2026-06-30
Tessera: loading byref argument types using global variable approach

jessicacotturone21:
Creates a `byRefTypes` `ArrayAttr` stored on the `tessera.define` op, which contains the types of each of the arguments that are marked byRef, or unit if an argument is not byRef. The order of the types in the array aligns with the order of the arguments. The sret argument is not included. `byRefArgs` (the boolean array which indicated true or false based on if an argument was marked as byRef or not) has been eliminated, since `byRefTypes` contains this information by either storing a `TypeAttr` or a `UnitAttr`.

If an argument is marked as byRef, the type is found by looking up the global variable emitted in the Clang plugin.

The types are used in the `tessera.call` op rewrite to load the arguments that are marked as byref.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
@@ -24,7 +24,7 @@ namespace mlir::enzyme::tessera {} // namespace mlir::enzyme::tessera
 
 void DefineOp::build(OpBuilder &builder, OperationState &state, StringRef name,
                      FunctionType type, DenseBoolArrayAttr byRefArgs,
-                     DenseI64ArrayAttr argSizes, bool pure,
+                     DenseI64ArrayAttr globalTypeIndices, bool pure,
                      StringAttr sym_visibility, ArrayRef<NamedAttribute> attrs,
                      ArrayRef<DictionaryAttr> argAttrs) {
   state.addAttribute(SymbolTable::getSymbolAttrName(),
@@ -32,7 +32,7 @@ void DefineOp::build(OpBuilder &builder, OperationState &state, StringRef name,
   state.addAttribute(getFunctionTypeAttrName(state.name), TypeAttr::get(type));
   state.addAttribute("pure", builder.getBoolAttr(pure));
   state.addAttribute("byRefArgs", byRefArgs);
-  state.addAttribute("argSizes", argSizes);
+  state.addAttribute("globalTypeIndices", globalTypeIndices);
 
   if (sym_visibility)
     state.addAttribute(getSymVisibilityAttrName(state.name), sym_visibility);
--- a/src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
@@ -24,15 +24,15 @@ namespace mlir::enzyme::tessera {} // namespace mlir::enzyme::tessera
 
 void DefineOp::build(OpBuilder &builder, OperationState &state, StringRef name,
                      FunctionType type, DenseBoolArrayAttr byRefArgs,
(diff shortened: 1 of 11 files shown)


src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
```diff
@@ -24,15 +24,15 @@ namespace mlir::enzyme::tessera {} // namespace mlir::enzyme::tessera
...
+                     DenseI64ArrayAttr globalTypeIndices, bool pure,
                      StringAttr sym_visibility, ArrayRef<NamedAttribute> attrs,
                      ArrayRef<DictionaryAttr> argAttrs) {
   state.addAttribute(SymbolTable::getSymbolAttrName(),
                      builder.getStringAttr(name));
   state.addAttribute(getFunctionTypeAttrName(state.name), TypeAttr::get(type));
   state.addAttribute("pure", builder.getBoolAttr(pure));
   state.addAttribute("byRefArgs", byRefArgs);
-  state.addAttribute("argSizes", argSizes);
+  state.addAttribute("globalTypeIndices", globalTypeIndices);
```
wsmoses: Why is this an array of i64s, wouldn't this make more sense as an array of typeattrs?

src/enzyme_ad/jax/Dialect/Tessera/Ops.td
```diff
@@ -94,6 +94,7 @@ def DefineOp : TesseraOp<"define", [
 
   }];
   let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
```
wsmoses, line 105: I might recommend adding a helper function that returns an ArrayRef<Type>

src/enzyme_ad/jax/Passes/Tessera/LLVMToTessera.cpp
```diff
@@ -91,24 +94,37 @@ class FuncOpRewrite final : public OpRewritePattern<LLVM::LLVMFuncOp> {
...
-      for (auto s : sizeParts) {
-        int64_t size;
-        s.trim().getAsInteger(10, size);
-        sizes.push_back(size);
+    // Find the types of the byref arguments by looking for global variables
+    // with names that match the pattern "__tessera_byref_arg_type_<index>"
+    // where <index> is a number parsed in the tessera_op attribute after
+    // "globals=".
+    SmallVector<Attribute> byRefTypes;
+    StringRef indicesStr = raw.substr(raw.find(')') + 1);
```
wsmoses: can you add a test where only some values are byref. What's stored in the array if something is not byref?

I would expect this loop to have pushed a null into the list in that case [such that the length of the args == the length of the ArrayAttr, with the indices matching]

src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
```diff
@@ -147,21 +145,51 @@ Attribute DefineOp::getSretAttr() {
...
   return nullptr;
 }
 
+ArrayRef<Attribute> DefineOp::getByRefTypesArray() {
+  return getByRefTypes().getValue();
+}
+
+Type DefineOp::getByRefType(unsigned argIdx) {
+  auto types = getByRefTypesArray();
+  if (argIdx >= types.size())
```
wsmoses: we shouldn't do this we should throw an out of bounds error

src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
```diff
@@ -147,21 +145,51 @@ Attribute DefineOp::getSretAttr() {
...
+    return Type();
+  auto attr = types[argIdx];
+  if (!isa<TypeAttr>(attr))
+    return Type();
```
wsmoses: return nullptr here

wsmoses (approved):

PR #2596, opened 2026-07-03
Add `excluded_passes` argument to C API

jumerckx:


wsmoses:
x/ref: https://github.com/EnzymeAD/Reactant.jl/pull/3022

wsmoses (approved):

jumerckx:
C API changes with this PR so Reactant CI is expected to fail (?). I verified locally that things work with https://github.com/EnzymeAD/Reactant.jl/pull/3025

PR #2598, opened 2026-07-07
Support kernel lowering when kernel has no args

Pangoraw:
As reported in https://github.com/EnzymeAD/Reactant.jl/issues/3026.


wsmoses (approved):
sure, add a test?

wsmoses:
going to merge for now but @Pangoraw can you investigate what is making ci red here?

PR #2602, opened 2026-07-08
Raise more parallel reduction types

### LINKED ISSUE #2607: `raise=true`: `affine.parallel ... reduce ("mulf")` product reduction fails to raise
## Summary

Only `addf` reductions appear to lower. A KA product over a runtime loop bound fails raise. A literal `for k in 1:3` raises and returns `[0.952]`; the runtime `@Const(n_layers)` bound is what emits `reduce ("mulf")`. Reproduced on Reactant 0.2.271 (2026-07-09).

## Environment

Julia 1.12.5 · Reactant 0.2.271 · Enzyme 0.13.173 · CUDA 6.2.0 · KernelAbstractions 0.9.42 · CPU backend (CUDA loaded for ReactantCUDAExt)

## Reproducer

```julia
using Reactant
using CUDA
using KernelAbstractions
import KernelAbstractions as KA

@assert Reactant.is_extension_loaded(Val(:CUDA))

@kernel function product_reduce_kernel!(cover, cloud_fraction, @Const(n_layers))
    col = @index(Global)
    FT = eltype(cover)
    prod_clear = one(FT)
    @inbounds for k in 1:n_layers
        cf = clamp(cloud_fraction[col, k], zero(FT), one(FT))
        prod_clear *= ifelse(cf > convert(FT, 1.0e-12), one(FT) - cf, one(FT))
    end
    @inbounds cover[col] = one(FT) - prod_clear
end

function product_reduce!(cover, cloud_fraction, n_layers)
    product_reduce_kernel!(KA.get_backend(cover))(cover, cloud_fraction, n_layers; ndrange = length(cover))
    return nothing
end

cover = Reactant.to_rarray(fill(-999.0, 1))
cloud_fraction = Reactant.to_rarray(reshape([0.7, 0.2, 0.8], 1, 3))
compiled = Reactant.@compile raise=true raise_first=true sync=true product_reduce!(cover, cloud_fraction, 3)
compiled(cover, cloud_fraction, 3)
```

## Expected behavior

The product reduction should lower like the supported additive reduction; the compiled kernel should return approximately `[0.952]`.

## Actual behavior

```text
ERROR: LoadError: CompilationError: MLIR pass pipeline "all" failed
failed to raise func: ... {
  %0 = affine.parallel (%arg2) = (0) to (3) reduce ("mulf") -> (f64) {
    ...
  }
}
caused by: "failed to run pass manager on module"
```

## Related issues

Related affine-reduction raising work is tracked in EnzymeAD/Enzyme-JAX [#560](https://github.com/EnzymeAD/Enzyme-JAX/issues/560) and [#566](https://github.com/EnzymeAD/Enzyme-JAX/issues/566). Those reports cover subtraction/general reduction support; this reproducer isolates `reduce ("mulf")`.

Pangoraw:


wsmoses (commented):
can you add isolated tests for each case?

wsmoses (commented):
lumping into the existing functions is just a lot to look through

wsmoses (approved):

PR #2614, opened 2026-07-13
Use absl call once, instead of std

wsmoses:
cc @giordano

### DIFF
--- a/src/enzyme_ad/jax/xla_ffi/register.cc
+++ b/src/enzyme_ad/jax/xla_ffi/register.cc
@@ -1,5 +1,6 @@
 #include <mutex>
 
+#include "absl/base/call_once.h"
 #include "cuda/register.h"
 #include "host/register.h"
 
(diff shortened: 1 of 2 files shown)


giordano:
I believe you also need to add
```
"@com_google_absl//absl/base",
```
around https://github.com/EnzymeAD/Enzyme-JAX/blob/1f137356f94e74ae9f57b68096bb942fde807c10/src/enzyme_ad/jax/BUILD#L65

src/enzyme_ad/jax/xla_ffi/register.cc
```diff
@@ -1,5 +1,6 @@
 #include <mutex>
```
giordano: mutex isn't needed with this change
```suggestion
```

giordano (approved):
This is good, as far as I understand.

wsmoses:
@giordano mingw build here succeeds still...

giordano:
Uh, that's good then, I guess?

giordano:
It did fail earlier today in https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/29262347113?pr=2595, so passing here is good sign.

PR #2615, opened 2026-07-13
try to bump libtpu

wsmoses:
Updated reactant_commit to include 'wsmoses-patch-18' and removed 'multipad'.


wsmoses:
@dkytezab this is blocked by https://github.com/EnzymeAD/Enzyme-JAX/pull/2618

PR #2619, opened 2026-07-14
LLVMToMemrefAccess: accept explicit argument type contracts

smart-moomoo:
LLVM-imported C++ kernels are standalone functions. The existing pass only discovers callees of enzymexla kernel and JIT calls, then recovers their memref types from ranked tensor operands. Imported kernels have no such callers, and their LLVM pointer arguments no longer carry shape information.

We treat functions with an enzymexla.memref_type argument attribute as rewrite candidates and use that contract to replace the corresponding LLVM pointer with a ranked memref while preserving the existing caller-derived recovery path. Also, we preserve unannotated and non-pointer arguments, retain unrelated argument attributes, and consume the type contract when rebuilding the function. 

The PR and some PRs that we have only cover ABI-boundary contract, w/o/ general LLVM pointer type inference. So for future integration, we can absorb internal pointers, dynamic layouts, or aliasing information, and integrate Enzyme TypeAnalysis.

Add a lit test covering a standalone imported kernel, exact shape recovery, contract removal, and preservation of llvm.readonly.

Tests:
  bazel test //test/lit_tests:raising_llvm_to_memref_access_arg_attrs.mlir.test

### DIFF
--- a/src/enzyme_ad/jax/Passes/LLVMToMemrefAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToMemrefAccess.cpp
@@ -6,6 +6,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/MLIRContext.h"
+#include "src/enzyme_ad/jax/Dialect/Dialect.h"
 #include "src/enzyme_ad/jax/Dialect/Ops.h"
 
 #define DEBUG_TYPE "llvm-to-memref-access"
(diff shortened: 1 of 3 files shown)


smart-moomoo:
Some background for the PR (and some other PRs that we have):

- We're sasa's students who want to have an well-maintained (in terms of LLVM version) Polygeist in our downstream. 
- I have a fork that can partly work and I'm still evolving that. So we hope that we can progressively upstream some of our fixes to both help improve this project and to reduce our maintainence cost.
- My understanding is that this repos has the Polygeist lib working with almost up-to-date llvm version but is not end-to-end working for the cpp - llvm - mlir pipeline. If that's correct, I hope I can contribute to that aspect. 
- Feel free to give us feedbacks (e.g. if it's appropriate to put PRs for fixing Polygeist here)

wsmoses:
@ftynse @ivanradanov @ImanHosseini can you take a look

smart-moomoo:
Ping.

src/enzyme_ad/jax/Passes/LLVMToMemrefAccess.cpp
ImanHosseini, line 205: Add a check for `callee.isExternal()` to skip external llvm.function declarations. (otherwise you cant call getFunctionBody())

ImanHosseini (changes requested):

ImanHosseini:
Thanks for working on this @smart-moomoo . Sorry for the delay in reviewing.

PR #2625, opened 2026-07-15
Generalize JIT call fusion using SSA dependencies

yanzin00:
## Summary

This PR adds a generic `fuse-jit-calls` pass that combines directly SSA-connected `enzymexla.jit_call` operations into a single LLVM wrapper and replaces the original calls with one fused JIT call.

- For each JIT call, the pass examines its operands to find producer JIT calls and its results to find consumer JIT calls. This discovers the connected SSA dependency graph, including chains and DAGs containing more than two calls.

- Operands defined outside the discovered group become arguments of the fused call. Results used only by another JIT call in the group become internal dependencies and are mapped directly between the cloned wrappers. Results that still have users outside the group remain outputs of the fused call.
- Before modifying the IR, the pass validates that fusion is safe and supported. Currently, all calls must be in the same block, operands must dominate the fused call, and fusion must not move calls across unrelated side effects. The referenced LLVM wrappers must use the supported single-block, pointer-only, non-vararg, void-returning ABI. Calls with unsupported backend, layout, argument, or result metadata are not fused. Each JIT result must currently have a supported flat `output_operand_alias`.

## Test

The pass is tested in: `test/lit_tests/mpi/fuse_jit_calls.mlir`

The test first runs `lower-enzymexla-mpi` to lower the MPI operations into `enzymexla.jit_call` operations and LLVM wrapper functions. It then runs `fuse-jit-calls` and verifies that the dependent `MPI_Irecv` and `MPI_Wait` JIT calls are replaced by one fused JIT call.

### IR before `fuse-jit-calls`
```mlir
// -----// IR Dump Before FuseJITCallsPass (fuse-jit-calls) //----- //
module {
  llvm.mlir.global external constant @MPI_INT() {addr_space = 0 : i32} : !llvm.ptr
  llvm.mlir.global external constant @MPI_COMM_WORLD() {addr_space = 0 : i32} : !llvm.ptr
  llvm.func @MPI_Irecv(!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
  llvm.func @enzymexla_wrapper_MPI_Irecv_MPI_INT(%arg0: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg1: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg2: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg3: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg4: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}) attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.mlir.addressof @MPI_INT : !llvm.ptr
    %2 = llvm.load %arg1 : !llvm.ptr -> i32
    %3 = llvm.load %arg2 : !llvm.ptr -> i32
    %4 = llvm.load %arg3 : !llvm.ptr -> i32
    %5 = llvm.call @MPI_Irecv(%arg0, %2, %1, %3, %4, %0, %arg4) : (!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
    llvm.return
  }
  llvm.func @MPI_Wait(!llvm.ptr, !llvm.ptr) -> i32
  llvm.func @enzymexla_wrapper_MPI_Wait(%arg0: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}) attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %c1_i32 = arith.constant 1 : i32
    %0 = llvm.alloca %c1_i32 x !llvm.array<6 x i32> : (i32) -> !llvm.ptr
    %1 = llvm.call @MPI_Wait(%arg0, %0) : (!llvm.ptr, !llvm.ptr) -> i32
    llvm.return
  }
  func.func @main(%arg0: tensor<5xf64> {enzymexla.memory_effects = ["read", "write", "allocate", "free"], tf.aliasing_output = 0 : i32}) -> tensor<5xf64> attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %c = stablehlo.constant dense<-1> : tensor<i32>
    %c_0 = stablehlo.constant dense<5> : tensor<i32>
    %c_1 = stablehlo.constant dense<42> : tensor<i32>
    %c_2 = stablehlo.constant dense<1> : tensor<i32>
    %0 = stablehlo.transpose %arg0, dims = [0] : (tensor<5xf64>) -> tensor<5xf64>
    %1:2 = enzymexla.jit_call @enzymexla_wrapper_MPI_Irecv_MPI_INT (%0, %c_0, %c_2, %c_1, %c) {output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [0], operand_index = 0, operand_tuple_indices = []>, #stablehlo.output_operand_alias<output_tuple_indices = [1], operand_index = 4, operand_tuple_indices = []>]} : (tensor<5xf64>, tensor<i32>, tensor<i32>, tensor<i32>, tensor<i32>) -> (tensor<5xf64>, tensor<i32>)
    enzymexla.jit_call @enzymexla_wrapper_MPI_Wait (%1#1) : (tensor<i32>) -> ()
    %2 = stablehlo.transpose %1#0, dims = [0] : (tensor<5xf64>) -> tensor<5xf64>
    return %2 : tensor<5xf64>
  }
}
```
Before fusion, `MPI_Irecv` and `MPI_Wait` are represented by separate JIT calls. The request result `%1#1` from `MPI_Irecv` is passed directly to `MPI_Wait`, while the receive buffer `%1#0` remains externally used by `stablehlo.transpose`.

### IR `after fuse-jit-calls`
```mlir
// -----// IR Dump After FuseJITCallsPass (fuse-jit-calls) //----- //
module {
  llvm.mlir.global external constant @MPI_INT() {addr_space = 0 : i32} : !llvm.ptr
  llvm.mlir.global external constant @MPI_COMM_WORLD() {addr_space = 0 : i32} : !llvm.ptr
  llvm.func @MPI_Irecv(!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
  llvm.func @__enzyme_fused_enzymexla_wrapper_MPI_Irecv_MPI_INT_enzymexla_wrapper_MPI_Wait(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: !llvm.ptr, %arg3: !llvm.ptr, %arg4: !llvm.ptr) {
    %c1_i32 = arith.constant 1 : i32
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.mlir.addressof @MPI_INT : !llvm.ptr
    %2 = llvm.load %arg1 : !llvm.ptr -> i32
    %3 = llvm.load %arg2 : !llvm.ptr -> i32
    %4 = llvm.load %arg3 : !llvm.ptr -> i32
    %5 = llvm.call @MPI_Irecv(%arg0, %2, %1, %3, %4, %0, %arg4) : (!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
    %6 = llvm.alloca %c1_i32 x !llvm.array<6 x i32> : (i32) -> !llvm.ptr
    %7 = llvm.call @MPI_Wait(%arg4, %6) : (!llvm.ptr, !llvm.ptr) -> i32
    llvm.return
  }
  llvm.func @enzymexla_wrapper_MPI_Irecv_MPI_INT(%arg0: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg1: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg2: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg3: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}, %arg4: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}) attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.mlir.addressof @MPI_INT : !llvm.ptr
    %2 = llvm.load %arg1 : !llvm.ptr -> i32
    %3 = llvm.load %arg2 : !llvm.ptr -> i32
    %4 = llvm.load %arg3 : !llvm.ptr -> i32
    %5 = llvm.call @MPI_Irecv(%arg0, %2, %1, %3, %4, %0, %arg4) : (!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
    llvm.return
  }
  llvm.func @MPI_Wait(!llvm.ptr, !llvm.ptr) -> i32
  llvm.func @enzymexla_wrapper_MPI_Wait(%arg0: !llvm.ptr {enzymexla.memory_effects = ["read", "write", "allocate", "free"]}) attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %c1_i32 = arith.constant 1 : i32
    %0 = llvm.alloca %c1_i32 x !llvm.array<6 x i32> : (i32) -> !llvm.ptr
    %1 = llvm.call @MPI_Wait(%arg0, %0) : (!llvm.ptr, !llvm.ptr) -> i32
    llvm.return
  }
  func.func @main(%arg0: tensor<5xf64> {enzymexla.memory_effects = ["read", "write", "allocate", "free"], tf.aliasing_output = 0 : i32}) -> tensor<5xf64> attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %c = stablehlo.constant dense<-1> : tensor<i32>
    %c_0 = stablehlo.constant dense<5> : tensor<i32>
    %c_1 = stablehlo.constant dense<42> : tensor<i32>
    %c_2 = stablehlo.constant dense<1> : tensor<i32>
    %0 = stablehlo.transpose %arg0, dims = [0] : (tensor<5xf64>) -> tensor<5xf64>
    %1 = enzymexla.jit_call @__enzyme_fused_enzymexla_wrapper_MPI_Irecv_MPI_INT_enzymexla_wrapper_MPI_Wait (%0, %c_0, %c_2, %c_1, %c) {output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [], operand_index = 0, operand_tuple_indices = []>]} : (tensor<5xf64>, tensor<i32>, tensor<i32>, tensor<i32>, tensor<i32>) -> tensor<5xf64>
    %2 = stablehlo.transpose %1, dims = [0] : (tensor<5xf64>) -> tensor<5xf64>
    return %2 : tensor<5xf64>
  }
}
```

After fusion, both wrapper bodies are cloned into a single fused LLVM wrapper, and the two original JIT calls are replaced with a single fused JIT call. The `MPI request` becomes an internal pointer shared between `MPI_Irecv` and `MPI_Wait`, so it is removed from the JIT results. The receive buffer remains externally visible, and its alias to operand 0 is preserved.
- The original LLVM wrappers remain temporarily and may be removed by a later symbol DCE pass if they are no longer referenced.

---

### MPI Waitall Variadic

- Modified MPI Waitall to accept variadic request operands, allowing us to detect JIT-to-JIT dependencies and fuse them.
- The modified MPI Waitall packs the request handles into the contiguous temporary array required by MPI, calls MPI_Waitall, and copies the updated handles back to the original operand buffers.

### test coverage

- Added a lit test verifying that two MPI_Irecv calls and their variadic MPI_Waitall dependency are fused into a single JIT call.
- Verified that the fused wrapper preserves the correct execution order: both receives produce request handles before MPI_Waitall consumes them.
- Added a regression test for Reactant [issue #2591](https://github.com/EnzymeAD/Reactant.jl/issues/2591), ensuring the completed receive buffer is returned by the fused JIT call before being unpacked.
- Verified that output aliases and results are preserved and that the original standalone Irecv and Waitall JIT calls are removed.
- As an intentionally scoped first implementation, JIT fusion currently excludes LAPACK/BLAS calls because their lowerings attach `operand_layouts` and `result_layouts`, which `FuseJITPass` does not accept.

### DIFF
--- a/src/enzyme_ad/jax/Passes/FuseJITPass.cpp
+++ b/src/enzyme_ad/jax/Passes/FuseJITPass.cpp
@@ -15,12 +15,12 @@
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
 
-#define DEBUG_TYPE "fuse-jit-calls"
+#define DEBUG_TYPE "fuse-jit"
 
 namespace mlir {
 namespace enzyme {
 
-#define GEN_PASS_DEF_FUSEJITCALLSPASS
+#define GEN_PASS_DEF_FUSEJITPASS
 #include "src/enzyme_ad/jax/Passes/Passes.h.inc"
 
 namespace {
@@ -32,7 +32,7 @@ struct JITFusionInfo {
   SmallVector<Value> fusedReturns;
 };
 
-static constexpr llvm::StringLiteral FusedNamePrefix = "__enzyme_fused_";
+static constexpr llvm::StringLiteral FusedNamePrefix = "fused__";
 
 static bool isGeneratedFusedCall(enzymexla::JITCallOp call) {
   // Prevent the greedy driver from folding a generated call again.
@@ -394,7 +394,7 @@ getFusedSideEffectFreeAttr(ArrayRef<enzymexla::JITCallOp> fusionCalls) {
   return UnitAttr();
 }
 
-struct FuseJITCallsPattern : public OpRewritePattern<enzymexla::JITCallOp> {
+struct FuseJITPattern : public OpRewritePattern<enzymexla::JITCallOp> {
   using OpRewritePattern<enzymexla::JITCallOp>::OpRewritePattern;
 
   LogicalResult matchAndRewrite(enzymexla::JITCallOp jitCallOp,
@@ -534,13 +534,13 @@ struct FuseJITCallsPattern : public OpRewritePattern<enzymexla::JITCallOp> {
   }
 };
 
-struct FuseJITCallsPass : public impl::FuseJITCallsPassBase<FuseJITCallsPass> {
+struct FuseJITPass : public impl::FuseJITPassBase<FuseJITPass> {
   void runOnOperation() override {
     ModuleOp module = getOperation();
     MLIRContext *context = &getContext();
     RewritePatternSet patterns(context);
 
-    patterns.add<FuseJITCallsPattern>(context);
+    patterns.add<FuseJITPattern>(context);
 
     if (failed(applyPatternsGreedily(module, std::move(patterns)))) {
       signalPassFailure();
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
@@ -1276,17 +1276,21 @@ struct MPIWaitallOpLowering : public OpRewritePattern<enzymexla::MPIWaitallOp> {
       auto i32Type = IntegerType::get(context, 32);
 
       std::string mpiFunctionName = "MPI_Waitall";
+      auto requests = op.getRequests();
+      unsigned numRequests = requests.size();
 
       // Generate the enzymexla_wrapper LLVM function body
-      std::string wrapperFunctionName = "enzymexla_wrapper_" + mpiFunctionName;
+      std::string wrapperFunctionName = "enzymexla_wrapper_" + mpiFunctionName +
+                                        "_" + std::to_string(numRequests);
 
       if (!moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(wrapperFunctionName)) {
         OpBuilder::InsertionGuard guard(rewriter);
         rewriter.setInsertionPointToStart(moduleOp.getBody());
 
         // Create the wrapper function decl
+        SmallVector<Type> wrapperArgumentTypes(numRequests, llvmPtrType);
         auto funcType = LLVM::LLVMFunctionType::get(
-            llvmVoidType, {llvmPtrType, llvmPtrType}, false);
+            llvmVoidType, wrapperArgumentTypes, false);
 
         auto wrapperFunc = LLVM::LLVMFuncOp::create(
             rewriter, op.getLoc(), wrapperFunctionName, funcType);
@@ -1302,18 +1306,31 @@ struct MPIWaitallOpLowering : public OpRewritePattern<enzymexla::MPIWaitallOp> {
         rewriter.setInsertionPointToStart(entryBlock);
 
         // Add argument-level memory effects attribute to all arguments
-        for (unsigned i = 0; i < 2; ++i) {
+        for (unsigned i = 0; i < numRequests; ++i) {
           wrapperFunc.setArgAttr(i, "enzymexla.memory_effects",
                                  memoryEffectsAttr);
         }
 
-        // Get the function argument
-        Value countPtr = entryBlock->getArgument(0);
-        Value requestPtr = entryBlock->getArgument(1);
+        Value count = arith::ConstantOp::create(
+            rewriter, op.getLoc(), i32Type,
+            rewriter.getI32IntegerAttr(static_cast<int32_t>(numRequests)));
 
-        // Load the count value
-        Value count =
-            LLVM::LoadOp::create(rewriter, op.getLoc(), i32Type, countPtr);
+        // Pack the scalar requests into the native contiguous request array.
+        Value requestsPtr = LLVM::AllocaOp::create(rewriter, op.getLoc(),
+                                                   llvmPtrType, i32Type, count);
+        SmallVector<Value> requestElementPtrs;
+        requestElementPtrs.reserve(numRequests);
+        for (unsigned index = 0; index < numRequests; ++index) {
+          Value requestPtr = entryBlock->getArgument(index);
+          Value request =
+              LLVM::LoadOp::create(rewriter, op.getLoc(), i32Type, requestPtr);
+          Value requestElementPtr = LLVM::GEPOp::create(
+              rewriter, op.getLoc(), llvmPtrType, i32Type, requestsPtr,
+              ArrayRef<LLVM::GEPArg>{static_cast<int32_t>(index)});
+          LLVM::StoreOp::create(rewriter, op.getLoc(), request,
+                                requestElementPtr);
+          requestElementPtrs.push_back(requestElementPtr);
+        }
 
         // Allocate a count x !llvm.array<6 x i32> for the array of statuses
         // Size of status is implem dependendent, 6 should cover the max
@@ -1328,7 +1345,16 @@ struct MPIWaitallOpLowering : public OpRewritePattern<enzymexla::MPIWaitallOp> {
         // TODO returns i32 error code which we're ignoring here
         LLVM::CallOp::create(rewriter, op.getLoc(), TypeRange{i32Type},
                              SymbolRefAttr::get(context, mpiFunctionName),
-                             ValueRange{count, requestPtr, statusPtr});
+                             ValueRange{count, requestsPtr, statusPtr});
+
+        // MPI_Waitall sets completed requests to MPI_REQUEST_NULL.
+        for (unsigned index = 0; index < numRequests; ++index) {
+          Value requestPtr = entryBlock->getArgument(index);
+          Value requestElementPtr = requestElementPtrs[index];
+          Value request = LLVM::LoadOp::create(rewriter, op.getLoc(), i32Type,
+                                               requestElementPtr);
+          LLVM::StoreOp::create(rewriter, op.getLoc(), request, requestPtr);
+        }
 
         LLVM::ReturnOp::create(rewriter, op.getLoc(), ValueRange{});
       }
@@ -1345,14 +1371,11 @@ struct MPIWaitallOpLowering : public OpRewritePattern<enzymexla::MPIWaitallOp> {
                                  funcType, LLVM::Linkage::External);
       }
 
-      // Get all orinigal op operands
-      auto opOperands = op.getOperands();
-
       // Call the LLVM function with enzymexla.jit_call
       enzymexla::JITCallOp::create(
           rewriter, op.getLoc(), TypeRange{},
-          mlir::FlatSymbolRefAttr::get(context, wrapperFunctionName),
-          ValueRange{opOperands}, rewriter.getStringAttr(""),
+          mlir::FlatSymbolRefAttr::get(context, wrapperFunctionName), requests,
+          rewriter.getStringAttr(""),
           /*operand_layouts=*/nullptr,
           /*result_layouts=*/nullptr,
           /*arg_attrs=*/nullptr,
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
@@ -16,7 +16,8 @@
 // #include "llvm/Support/LogicalResult.h"
 // #include "llvm/Support/MathExtras.h"
 // #include <algorithm>
-// #include <cstdint>
+#include <cstdint>
+#include <limits>
 
 namespace mlir {
 namespace enzyme {
@@ -768,9 +769,12 @@ struct MPIRecvOpLowering : public OpRewritePattern<enzymexla::MPIRecvOp> {
 struct MPIIsendOpLowering : public OpRewritePattern<enzymexla::MPIIsendOp> {
 
   std::string backend;
+  int32_t &nextRequestStorageInitializer;
   MPIIsendOpLowering(std::string backend, MLIRContext *context,
+                     int32_t &nextRequestStorageInitializer,
                      PatternBenefit benefit = 1)
-      : OpRewritePattern(context, benefit), backend(backend) {}
+      : OpRewritePattern(context, benefit), backend(backend),
+        nextRequestStorageInitializer(nextRequestStorageInitializer) {}
 
   LogicalResult matchAndRewrite(enzymexla::MPIIsendOp op,
                                 PatternRewriter &rewriter) const override {
@@ -910,10 +914,16 @@ struct MPIIsendOpLowering : public OpRewritePattern<enzymexla::MPIIsendOp> {
       // Get all orinigal op operands
       auto opOperands = op.getOperands();
 
-      // Create a constant tensor to hold request
+      // Each isend/irecv ops needs it's own storage for requests.
+      // Initialize constants with unique values so that folding doesn't
+      // cause other ops to resuse the same buffer
+      if (nextRequestStorageInitializer ==
+          std::numeric_limits<int32_t>::min())
+        return rewriter.notifyMatchFailure(
+            op, "exhausted distinct MPI request storage initializers");
       auto tensorType = RankedTensorType::get({}, i32Type);
-      auto constantAttr =
-          DenseIntElementsAttr::get(tensorType, ArrayRef<int32_t>{-1});
+      auto constantAttr = DenseIntElementsAttr::get(
+          tensorType, ArrayRef<int32_t>{nextRequestStorageInitializer--});
       Value constantTensor = stablehlo::ConstantOp::create(
           rewriter, op.getLoc(), tensorType, constantAttr);
 
@@ -954,9 +964,12 @@ struct MPIIsendOpLowering : public OpRewritePattern<enzymexla::MPIIsendOp> {
 struct MPIIrecvOpLowering : public OpRewritePattern<enzymexla::MPIIrecvOp> {
 
   std::string backend;
+  int32_t &nextRequestStorageInitializer;
   MPIIrecvOpLowering(std::string backend, MLIRContext *context,
+                     int32_t &nextRequestStorageInitializer,
                      PatternBenefit benefit = 1)
-      : OpRewritePattern(context, benefit), backend(backend) {}
+      : OpRewritePattern(context, benefit), backend(backend),
+        nextRequestStorageInitializer(nextRequestStorageInitializer) {}
 
   LogicalResult matchAndRewrite(enzymexla::MPIIrecvOp op,
                                 PatternRewriter &rewriter) const override {
@@ -1097,10 +1110,16 @@ struct MPIIrecvOpLowering : public OpRewritePattern<enzymexla::MPIIrecvOp> {
       // Get all orinigal op operands
       auto opOperands = op.getOperands();
 
-      // Create a constant tensor to hold request
+      // Each isend/irecv ops needs it's own storage for requests.
+      // Initialize constants with unique values so that folding doesn't
+      // cause other ops to resuse the same buffer
+      if (nextRequestStorageInitializer ==
+          std::numeric_limits<int32_t>::min())
+        return rewriter.notifyMatchFailure(
+            op, "exhausted distinct MPI request storage initializers");
       auto tensorType = RankedTensorType::get({}, i32Type);
-      auto constantAttr =
-          DenseIntElementsAttr::get(tensorType, ArrayRef<int32_t>{-1});
+      auto constantAttr = DenseIntElementsAttr::get(
+          tensorType, ArrayRef<int32_t>{nextRequestStorageInitializer--});
       Value constantTensor = stablehlo::ConstantOp::create(
           rewriter, op.getLoc(), tensorType, constantAttr);
 
@@ -1754,14 +1773,17 @@ struct LowerEnzymeXLAMPIPass
   void runOnOperation() override {
     auto context = getOperation()->getContext();
     RewritePatternSet patterns(context);
+    int32_t nextRequestStorageInitializer = -1;
 
     patterns.add<MPICommRankOpLowering>(backend, context);
     patterns.add<MPICommSizeOpLowering>(backend, context);
     patterns.add<MPIBarrierOpLowering>(backend, context);
     patterns.add<MPISendOpLowering>(backend, context);
     patterns.add<MPIRecvOpLowering>(backend, context);
-    patterns.add<MPIIsendOpLowering>(backend, context);
-    patterns.add<MPIIrecvOpLowering>(backend, context);
+    patterns.add<MPIIsendOpLowering>(backend, context,
+                                     nextRequestStorageInitializer);
+    patterns.add<MPIIrecvOpLowering>(backend, context,
+                                     nextRequestStorageInitializer);
     patterns.add<MPIWaitOpLowering>(backend, context);
     patterns.add<MPIWaitallOpLowering>(backend, context);
     patterns.add<MPIAllreduceOpLowering>(backend, context);
--- a/test/lit_tests/fuse_jit/mpi_independent-components.mlir
+++ b/test/lit_tests/fuse_jit/mpi_independent-components.mlir
@@ -20,9 +20,9 @@ module {
 
 // CHECK-LABEL: func.func @main
 // CHECK-NOT:     enzymexla.jit_call @enzymexla_wrapper_MPI_Irecv_MPI_INT
-// CHECK:         %[[FIRST:.*]] = enzymexla.jit_call @fused__enzymexla_wrapper_MPI_Irecv_MPI_INT_enzymexla_wrapper_MPI_Wait
+// CHECK:         %[[FIRST:.*]] = enzymexla.jit_call @fused__enzymexla_wrapper_MPI_Irecv_MPI_INT_enzymexla_wrapper_MPI_Wait{{(_1)?}}
 // CHECK-SAME:    output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [], operand_index = 0, operand_tuple_indices = []>]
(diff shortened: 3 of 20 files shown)


romanlee:
Addresses https://github.com/EnzymeAD/Reactant.jl/issues/2591

romanlee:
Nice work on this!

It might be good at least initially for this pass to additionally filter for only Irecv, Isend, Wait and Waitall. Later on this filter could be removed to make it more general as you have it now. That way we can make sure we fix MPI and don't risking breaking a whole bunch of other cases before rolling it out to be more general.

Also I might suggest calling it `FuseJITPass`/`fuse-jit` etc instead. This fits slightly better alongside `LowerJITPass`.

For the tests:
* I think it would be good to add a test starting directly from after `lower-enzymexla-mpi` (what you call "before IR before `fuse-jit-calls`" above) to isolate the pass further. 
* It would be good to add a test with multiple comm calls and a waitall, eg isend + irecv + waitall.

Otherwise I think this looks pretty good. You said you're still working on some implementation details?

test/lit_tests/mpi/fuse_jit_calls.mlir
```diff
@@ -0,0 +1,33 @@
+
+// RUN: enzymexlamlir-opt --pass-pipeline="builtin.module(lower-enzymexla-mpi{backend=cpu},fuse-jit-calls)" %s | FileCheck %s --check-prefix=CPU
```
mofeing: the `CPU` prefix is usually used when we must test multiple backends (e.g. CPU vs CUDA), which is not what you're testing here. you should use the `CHECK` prefix instead.

```suggestion
// RUN: enzymexlamlir-opt --pass-pipeline="builtin.module(lower-enzymexla-mpi{backend=cpu},fuse-jit-calls)" %s | FileCheck %s
```

mofeing (commented):
> It might be good at least initially for this pass to additionally filter for only Irecv, Isend, Wait and Waitall. Later on this filter could be removed to make it more general as you have it now. That way we can make sure we fix MPI and don't risking breaking a whole bunch of other cases before rolling it out to be more general.

I actually advised @yanzin00 against that. The reason being that the filtering makes the implementation more complex when we are the only current users yet and can be enabled with an option in Reactant's `CompileOption` construct as an opt-in. So we are not breaking anything.

Agreed about the tests. Move the test out of the `lit_tests/mpi/` folder as it will soon be applied to other contexts (you can still use the examples from MPI lowering as tests, there's no problem on that).

We must also test the edge cases that we spoke about last time (i.e. no results, no arguments, fusion but intermediate variable used elsewhere, etc.).

test/lit_tests/mpi/fuse_jit_calls.mlir
yanzin00: Thank you, I've added it in commit [8b1693b](https://github.com/EnzymeAD/Enzyme-JAX/commit/8b1693bdf84c22fe3b29ebe5c550cf55c7b4ebbf)

test/lit_tests/fuse_jit/mpi_isend-irecv-waitall-fusion.mlir
```diff
@@ -0,0 +1,35 @@
+// RUN: enzymexlamlir-opt --pass-pipeline="builtin.module(lower-enzymexla-mpi{backend=cpu},fuse-jit)" %s | FileCheck %s
+
+// A mixed nonblocking component must fuse as one unit while retaining lexical
+// Isend -> Irecv -> Waitall order in the generated wrapper.
```
romanlee, line 4: Out of curiosity, what does this mean? Lexical order must remain the same? Why?

test/lit_tests/fuse_jit/mpi_independent-components.mlir
```diff
@@ -0,0 +1,28 @@
+// RUN: enzymexlamlir-opt --pass-pipeline="builtin.module(lower-enzymexla-mpi{backend=cpu},fuse-jit)" %s | FileCheck %s
+
+// Two independent communication chains must remain separate fusion components.
```
romanlee, line 3: why can't these all be fused together?

romanlee:
I brought this up earlier, but what's the reasoning for having all the test go through `lower-enzymexla-mpi` first? I feel like that would make the tests easier to read / would isolate this pass better to just go directly from the result of that, and similar for the eg `lower-enzymexla-blas` etc cases.

test/lit_tests/fuse_jit/mpi_nonblocking-order.mlir
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// CHECK-NOT: llvm.func @fused__
+// CHECK-LABEL: func.func @rank0
+// CHECK: %[[SEND_REQUEST:.*]] = enzymexla.jit_call @enzymexla_wrapper_MPI_Isend_MPI_INT
+// CHECK-NEXT: %[[RECV:.*]]:2 = enzymexla.jit_call @enzymexla_wrapper_MPI_Irecv_MPI_INT
+// CHECK-NEXT: enzymexla.jit_call @enzymexla_wrapper_MPI_Wait (%[[SEND_REQUEST]])
+// CHECK-NEXT: enzymexla.jit_call @enzymexla_wrapper_MPI_Send_MPI_INT
+// CHECK-NEXT: enzymexla.jit_call @enzymexla_wrapper_MPI_Wait (%[[RECV]]#1)
+// CHECK-NEXT: return %[[RECV]]#0
```
romanlee, line 32: When I run enzymemliropt on this I get 
```
romanlee@hydra:~/codes/Enzyme-JAX$ bazel-bin/enzymexlamlir-opt --pass-pipeline="builtin.module(lower-enzymexla-mpi{backend=cpu},fuse-jit)" test/lit_tests/fuse_jit/mpi_nonblocking-order.mlir

...

  func.func @rank0(%arg0: tensor<5xf64>, %arg1: tensor<5xf64>, %arg2: tensor<1xi32>) -> tensor<5xf64> attributes {enzymexla.memory_effects = ["read", "write", "allocate", "free"]} {
    %c = stablehlo.constant dense<1> : tensor<i32>
    %c_0 = stablehlo.constant dense<5> : tensor<i32>
    %c_1 = stablehlo.constant dense<10> : tensor<i32>
    %c_2 = stablehlo.constant dense<20> : tensor<i32>
    %c_3 = stablehlo.constant dense<30> : tensor<i32>
    %c_4 = stablehlo.constant dense<-1> : tensor<i32>
    %0 = enzymexla.jit_call @enzymexla_wrapper_MPI_Isend_MPI_INT (%arg0, %c_0, %c, %c_1, %c_4) {output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [], operand_index = 4, operand_tuple_indices = []>]} : (tensor<5xf64>, tensor<i32>, tensor<i32>, tensor<i32>, tensor<i32>) -> tensor<i32>
    %1:2 = enzymexla.jit_call @enzymexla_wrapper_MPI_Irecv_MPI_INT (%arg1, %c_0, %c, %c_2, %c_4) {output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [0], operand_index = 0, operand_tuple_indices = []>, #stablehlo.output_operand_alias<output_tuple_indices = [1], operand_index = 4, operand_tuple_indices = []>]} : (tensor<5xf64>, tensor<i32>, tensor<i32>, tensor<i32>, tensor<i32>) -> (tensor<5xf64>, tensor<i32>)
    enzymexla.jit_call @enzymexla_wrapper_MPI_Wait (%0) : (tensor<i32>) -> ()
    enzymexla.jit_call @enzymexla_wrapper_MPI_Send_MPI_INT (%arg2, %c, %c, %c_3) : (tensor<1xi32>, tensor<i32>, tensor<i32>, tensor<i32>) -> ()
    enzymexla.jit_call @enzymexla_wrapper_MPI_Wait (%1#1) : (tensor<i32>) -> ()
    return %1#0 : tensor<5xf64>
  }
}

```
It looks like nothing gets fused here. Don't we need everything to be fused here? Ie, doesn't this leave us vulnerable to the same issue the fusion pass is trying to fix?

test/lit_tests/fuse_jit/mpi_irecv-isend-waitall-fusion.mlir
romanlee, line 1: Have you tested this end to end with Reactant yet? I ask because when I run the pass on this file the generated IR looks weird to me.

```
romanlee@hydra:~/codes/Enzyme-JAX$ bazel-bin/enzymexlamlir-opt --pass-pipeline="builtin.module(lower-enzymexla-mpi{backend=cpu},fuse-jit)" test/lit_tests/fuse_jit/mpi_irecv-isend-waitall-fusion.mlir
module {
  llvm.func @MPI_Irecv(!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
  llvm.func @fused__enzymexla_wrapper_MPI_Irecv_MPI_INT_enzymexla_wrapper_MPI_Isend_MPI_INT_enzymexla_wrapper_MPI_Waitall_2(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: !llvm.ptr, %arg3: !llvm.ptr, %arg4: !llvm.ptr) {
    %c2_i32 = arith.constant 2 : i32
    %0 = llvm.mlir.addressof @MPI_COMM_WORLD : !llvm.ptr
    %1 = llvm.mlir.addressof @MPI_INT : !llvm.ptr
    %2 = llvm.load %arg1 : !llvm.ptr -> i32
    %3 = llvm.load %arg2 : !llvm.ptr -> i32
    %4 = llvm.load %arg3 : !llvm.ptr -> i32
    %5 = llvm.call @MPI_Irecv(%arg0, %2, %1, %3, %4, %0, %arg4) : (!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
    %6 = llvm.load %arg1 : !llvm.ptr -> i32
    %7 = llvm.load %arg2 : !llvm.ptr -> i32
    %8 = llvm.load %arg3 : !llvm.ptr -> i32
    %9 = llvm.call @MPI_Isend(%arg0, %6, %1, %7, %8, %0, %arg4) : (!llvm.ptr, i32, !llvm.ptr, i32, i32, !llvm.ptr, !llvm.ptr) -> i32
    %10 = llvm.alloca %c2_i32 x i32 : (i32) -> !llvm.ptr
    %11 = llvm.load %arg4 : !llvm.ptr -> i32
    llvm.store %11, %10 : i32, !llvm.ptr
    %12 = llvm.load %arg4 : !llvm.ptr -> i32
    %13 = llvm.getelementptr %10[1] : (!llvm.ptr) -> !llvm.ptr, i32
    llvm.store %12, %13 : i32, !llvm.ptr
    %14 = llvm.alloca %c2_i32 x !llvm.array<6 x i32> : (i32) -> !llvm.ptr
    %15 = llvm.call @MPI_Waitall(%c2_i32, %10, %14) : (i32, !llvm.ptr, !llvm.ptr) -> i32
    %16 = llvm.load %10 : !llvm.ptr -> i32
    llvm.store %16, %arg4 : i32, !llvm.ptr
    %17 = llvm.load %13 : !llvm.ptr -> i32
    llvm.store %17, %arg4 : i32, !llvm.ptr
    llvm.return
  }
...
```
Both Irecv and Isend use arg4 for the request, then it's loaded twice into the request buffer, but both loads happen after Isend. Shouldn't there be a load after irecv but before isend?

romanlee:
Re the lit tests:
* Curious what others' thoughts are, but I wonder if the checks are too general. What I prefer to do is: test end-to-end with Reactant, make sure it works, dump the `code_hlo optimize=false`, and then use that as the CHECK in the lit test.
* It feels like there might be more lit tests than necessary. Do we need all of these? For instance, is there a good reason to have all those different combinations of isend, irecv, wait and waitall? I wonder if one or two would provide all the coverage we need

test/lit_tests/fuse_jit/mpi_isend-irecv-waitall-fusion.mlir
yanzin00: Yes, by lexical order I mean the order of the original operations in the block:` Isend -> Irecv -> Waitall`

test/lit_tests/fuse_jit/mpi_independent-components.mlir
yanzin00: Mostly because there isn't an SSA dependency connecting the two chains. Just to check that the fuse will not fuse jits that are not dependencies of each other.

test/lit_tests/fuse_jit/mpi_nonblocking-order.mlir
yanzin00: In that case, those JIT calls are not fused because fusing them could change the MPI ordering and cause a deadlock. For example:
```CPP
if (rank == 0) {
    MPI_Isend(A, ..., 1, TAG_A, comm, &sendA);
    MPI_Irecv(B, ..., 1, TAG_B, comm, &recvB);

    MPI_Wait(&sendA, MPI_STATUS_IGNORE);

    // This operation enables rank 1 to send B.
    MPI_Send(control, ..., 1, TAG_CONTROL, comm);

    MPI_Wait(&recvB, MPI_STATUS_IGNORE);
}

if (rank == 1) {
    MPI_Recv(A, ..., 0, TAG_A, comm, MPI_STATUS_IGNORE);

    // Do not send B until the control message arrives.
    MPI_Recv(
        control,
        ...,
        0,
        TAG_CONTROL,
        comm,
        MPI_STATUS_IGNORE
    );

    MPI_Send(B, ..., 0, TAG_B, comm);
}
```
However, I agree that in cases like this, where another effectful JIT call or operation appears between Irecv and Wait and prevents them from being safely fused. Leaving them unfused may still expose us to the same issue.

test/lit_tests/fuse_jit/mpi_irecv-isend-waitall-fusion.mlir
yanzin00: I have not tested this end to end with Reactant yet; that is what I am working on now. But good catch, I hadn't noticed this issue. If both operations share `%arg4`, the request produced by `Irecv` must be loaded before `Isend` overwrites it. I think the better fix is to give each request separate internal storage (e.g., %arg4 and %arg5), rather than depending on an intermediate load from shared storage. Thanks!

yanzin00:
I initially ran `lower-enymexla-mpi` before the fusion pass because I believed the MPI operations needed to be lowered to JIT calls for the pass to recognize them. I now understand that this is not necessary, and I agree that several of the lit tests are redundant. I’ll simplify them to keep only the combinations needed for coverage. Thanks also for the suggestion to test end-to-end with Reactant and use the real MLIR produced by `code_hlo(optimize=false)` in the lit test; that should make the tests more representative and focused.

test/lit_tests/fuse_jit/mpi_nonblocking-order.mlir
romanlee: The above doesn't have a deadlock though. As long as you don't reorder any MPI ops then you should be fine. And yeah, we need to fuse everything here otherwise we didn't solve the problem.

test/lit_tests/fuse_jit/mpi_isend-irecv-waitall-fusion.mlir
romanlee: Why must it remain the same though? In this case you could reorder Isend/Irecv without any issue.

In any case, it's fine that the order is kept, I just wouldn't say in the comment that the lexical order must be kept

```suggestion
// A mixed nonblocking component must fuse as one unit
```

test/lit_tests/fuse_jit/mpi_irecv-isend-waitall-fusion.mlir
romanlee: Got it. I think we should probably wait to merge until it's working, just to ensure that what we're checking in is correct code/minimize future noise

test/lit_tests/fuse_jit/mpi_independent-components.mlir
romanlee: In principle though there's no reason they couldn't be fused, right? If so, I would probably just remove the comment

```suggestion
```

romanlee:
> I initially ran lower-enymexla-mpi before the fusion pass because I believed the MPI operations needed to be lowered to JIT calls for the pass to recognize them. I now understand that this is not necessary, 

Well, they do need to be lowered to jit calls for the pass to recognize them, right? What I'm saying is, in the lit test, start with jit call ops instead of mpi ops, and then rather than applying both lower-enzymexla-mpi and fuse-jit passes in your tests, just apply the latter. Does that make sense?

> and I agree that several of the lit tests are redundant. I’ll simplify them to keep only the combinations needed for coverage. Thanks also for the suggestion to test end-to-end with Reactant and use the real MLIR produced by code_hlo(optimize=false) in the lit test; that should make the tests more representative and focused.

sg

test/lit_tests/fuse_jit/mpi_nonblocking-order.mlir
yanzin00: Are you suggesting that, when another compatible jit_call appears between the SSA-connected calls, the pass should include it in the fusion group even if it has no SSA dependency on them, while preserving the original lexical order?
I think that could preserve the MPI operation ordering without adding MPI-specific logic or Deadlocks, but I want to confirm that this is what you have in mind.

test/lit_tests/fuse_jit/mpi_irecv-isend-waitall-fusion.mlir
yanzin00: Sure, sounds good!

yanzin00:
Yes, make sense!

src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
mofeing, line 1: the `MPI_Waitall` fix now lives in #3100

src/enzyme_ad/jax/Passes/FuseJITPass.cpp
```diff
@@ -0,0 +1,554 @@
...
+  SmallVector<Value> fusedReturns;
+};
+
+static constexpr llvm::StringLiteral FusedNamePrefix = "fused__";
+
+static bool isGeneratedFusedCall(enzymexla::JITCallOp call) {
+  // Prevent the greedy driver from folding a generated call again.
+  return call.getFn().getRootReference().getValue().starts_with(
+      FusedNamePrefix);
+}
```
mofeing, line 41: is there a reason why not to recursively fold already fused jit calls? or is this to prevent problems because some jit calls have already been folded?

test/lit_tests/mpi/waitall.mlir
mofeing, line 1: same as above, the fix lives now in #3100 so just revert changes to this file

src/enzyme_ad/jax/Passes/FuseJITPass.cpp
```diff
@@ -0,0 +1,554 @@
...
+// Describes one direct SSA-connected JIT-call component and its IR boundary.
+struct JITFusionInfo {
+  SmallVector<enzymexla::JITCallOp> fusionCalls;
+  SmallVector<Value> fusedArgs;
+  SmallVector<Value> fusedReturns;
+};
+
+static constexpr llvm::StringLiteral FusedNamePrefix = "fused__";
+
+static bool isGeneratedFusedCall(enzymexla::JITCallOp call) {
```
mofeing, line 37: why are all the helper functions `static`? is it to avoid linking name collisions with other files?

test/lit_tests/fuse_jit/mpi_nonblocking-order.mlir
romanlee: Yes. At least in this case, we need them to be fused. This relates to Sergios comment below about recursive folding I think.

src/enzyme_ad/jax/Passes/FuseJITPass.cpp
yanzin00: The overall idea was to prevent external linking, but I didn't know that the `namespace` already does that. So there's no need to use `static`.

src/enzyme_ad/jax/Passes/FuseJITPass.cpp
yanzin00: The current collector finds the complete, directly SSA-connected component in one rewrite, so recursive fusion is not necessary for that specific case. That said, I agree that we can change it to allow already fused calls to participate in a later fusion, so we can include jit calls that don't have SSA dependencies, as Roman and I discussed above.

PR #2631, opened 2026-07-16
FullReduceTransposeReshape rewrite

### LINKED ISSUE #3039: Silent wrong result with raise_first=true: fused sum(dest .* weights) disagrees with the (correct) dest the same compiled call wrote
## Summary

`@compile raise=true raise_first=true` of a function that (a) launches a KA kernel doing one data-dependent gather-store into `dest` and then (b) returns `sum(dest .* weights)` produces a **silently wrong scalar** when all three `ndrange` dims are ≥ 2. The `dest` buffer written by the **same compiled call is correct** — recomputing `sum(dest .* weights)` on the host from the returned `dest` gives the right answer. Only the fused weighted reduction is wrong. `sum(dest)` (no broadcast multiply) is correct, and making any dimension 1 makes everything correct.

**Enzyme turned out to be irrelevant.** This was originally filed as a wrong `Enzyme.autodiff(ReverseWithPrimal, ...)` primal (the original reproducer is in the edit history). Minimizing it showed the Enzyme-free raised forward of the same composed function already returns the identical wrong scalar; the "wrong reverse primal" was just this forward miscompile observed through the AD pipeline.

Reproduced on Reactant 0.2.271 (2026-07-11).

## Environment

Julia 1.12.5 · Reactant 0.2.271 · CUDA 6.2.0 · KernelAbstractions 0.9.42 · CPU backend (CUDA loaded for ReactantCUDAExt)

## Reproducer

```julia
using Reactant
using CUDA  # only so ReactantCUDAExt loads; everything runs on the CPU backend
using KernelAbstractions
import KernelAbstractions as KA

@assert Reactant.is_extension_loaded(Val(:CUDA))

@kernel function gather_kernel!(dest, ref, level)
    i, k, ch = @index(Global, NTuple)
    row = ifelse(@inbounds(level[i, k]) > 0.5, 2, 1)
    @inbounds dest[i, k, ch] = ref[i, row, ch]
end

function weighted_sum(dest, ref, level, weights, n)
    gather_kernel!(KA.get_backend(dest))(dest, ref, level; ndrange = (n, 2, 2))
    return sum(dest .* weights)
end

for n in (1, 2)
    level = [0.3 0.7; 0.35 0.75][1:n, :]                       # k=1 picks row 1, k=2 picks row 2
    ref = Float64[r + 2 * (c - 1) for i in 1:n, r in 1:2, c in 1:2]
    weights = reshape(0.1 .* (1:(4n)), n, 2, 2)

    eager = weighted_sum(zeros(n, 2, 2), ref, level, weights, n)

    r_dest = Reactant.to_rarray(zeros(n, 2, 2))
    r_ref = Reactant.to_rarray(copy(ref))
    r_level = Reactant.to_rarray(copy(level))
    r_w = Reactant.to_rarray(copy(weights))
    compiled = Reactant.@compile raise=true raise_first=true weighted_sum(r_dest, r_ref, r_level, r_w, n)
    raised = Float64(compiled(r_dest, r_ref, r_level, r_w, n))

    host_redo = sum(Array(r_dest) .* weights)  # dest written by the same call, reduced on host
    println("n=$n  eager=$eager  raised=$raised  host recompute from returned dest=$host_redo")
end
```

## Expected behavior

The raised scalar matches eager at both sizes.

## Actual behavior

```text
n=1  eager=3.0  raised=3.0  host recompute from returned dest=3.0
n=2  eager=11.0  raised=10.0  host recompute from returned dest=11.0
```

At `n=2` the compiled call writes the correct `dest` (`[1,1,2,2,3,3,4,4]`) yet returns `10.0` instead of `11.0` — the value it returns is inconsistent with the buffer it just wrote.

## What minimization showed is (and is not) required

- **Required:** the data-dependent gather (`row` from an `ifelse` on `level`), the leading parallel index on the gathered array (`ref[i, row, ch]`; a column-independent `ref[row, ch]` is correct), all three `ndrange` dims ≥ 2, the broadcast-multiply inside the reduction (`sum(dest .* weights)`; plain `sum(dest)` is correct), and `raise_first=true` (plain `raise=true` is correct; `sync=true` is irrelevant).
- **Not required:** Enzyme (any mode), the binary-search loop, interpolation arithmetic, division, `Duplicated` shadows, prologue zeroing of `dest` — the original reproducer's whole radiative-transfer dressing.

## Related issues

Distinct from #3038 (forward multi-element gather channel collapse): there the kernel-written buffer itself is wrong. Here the kernel-written buffer is correct and only the fused reduction over it is wrong — consistent with the reduction re-deriving the kernel result (rather than reading the stored buffer) and getting a different answer. #3038 does not involve a consumer of the buffer inside the compiled region.

Pangoraw:
Closes https://github.com/EnzymeAD/Reactant.jl/issues/3039.

The pattern was invalid if the transformation was not the same on both sides of a binary op.


wsmoses (approved):
LGTM in a quick look but you should also have avik give it a once over too

PR #2635, opened 2026-07-18
Simplify real, imag, and abs of complex exp

### LINKED ISSUE #1990: imag, real, abs of `exp(complex number)`

SuryanshSS1011:
Fixes #1990.

Adds three algebraic simplifications to `EnzymeHLOOpt` for `real`, `imag`, and `abs` of a complex `exp`, matching the identities in the issue:

```
real(exp(x)) -> exp(real(x)) * cos(imag(x))
imag(exp(x)) -> exp(real(x)) * sin(imag(x))
abs(exp(x))  -> exp(real(x))
```

Each pattern matches when the operand is a complex-typed `stablehlo.exponential`, guarding on the element type so the non-complex `real`/`imag` cases are left to `RealOpCanon`/`ImagOpCanon`.

Test: `test/lit_tests/realimagabsexp.mlir`.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -15117,6 +15117,11 @@ struct RealOfExpSimplify final
     if (!exp)
       return failure();
 
+    // Only rewrite when the exp feeds nothing but this op, otherwise the
+    // original exp stays live and we just add redundant work.
+    if (!llvm::hasSingleElement(exp->getUsers()))
+      return failure();
+
     auto loc = op.getLoc();
     auto re = stablehlo::RealOp::create(rewriter, loc, exp.getOperand());
     auto im = stablehlo::ImagOp::create(rewriter, loc, exp.getOperand());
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -15103,6 +15103,75 @@ struct ImagOpCanon final
...
 
+// real(exp(x)) -> exp(real(x)) * cos(imag(x))
+struct RealOfExpSimplify final
+    : CheckedOpRewritePattern<stablehlo::RealOp, RealOfExpSimplify> {
+  using CheckedOpRewritePattern::CheckedOpRewritePattern;
+
+  LogicalResult matchAndRewriteImpl(stablehlo::RealOp op,
+                                    PatternRewriter &rewriter) const {
+    if (!isa<ComplexType>(op.getOperand().getType().getElementType()))
+      return failure();
```
wsmoses, line 15114: can you add an additional check that the only user of the exp is this realop? [eg if its used elsewhere we might as well avoid the redundant work]

SuryanshSS1011: Done, added the single-user guard so the rewrite only fires when the `exp` feeds nothing but this op, otherwise it bails and leaves the original `exp` untouched. I applied it to the `imag` and `abs` patterns too since the same redundant-work idea holds there. I also added a negative test where the `exp` is also returned directly so it is not rewritten.

PR #2637, opened 2026-07-18
Extend DotGeneralSimplify to handle broadcasted scalars

### LINKED ISSUE #1475: Extend DotGeneralSimplify to handle Broadcasted Scalars
xref https://github.com/EnzymeAD/Enzyme-JAX/pull/1473#pullrequestreview-3332130471

binoygeorge97:
Fixes #1475.

### Problem

`DotGeneralSimplify` rewrites `dot_general(splat, X)` into a `reduce` over the
contracting dims, scaled by the splat value and broadcast to the result shape.
Detection matched a splat operand only via `m_Constant`, so a
`broadcast_in_dim` of a rank-0 **runtime** scalar (e.g. a function argument)
was missed — no canonicalizer can fold it into a splat constant, so the full
`dot_general` survived even though it is still `s * reduce(X)`.

### Change

- Add `extractBroadcastScalar`, which looks through a rank-0 `broadcast_in_dim`
  (empty `broadcast_dimensions`) to its scalar operand `Value`. Mirrors the
  look-through in `extractSplatInt`, but returns a `Value` since the runtime
  case has no compile-time attribute.
- Detection now accepts either a splat constant or such a broadcast, on either
  operand. The constant path is unchanged; the runtime path scales the reduce
  by a broadcast of the scalar `Value` and always emits the multiply, relying
  on the existing downstream canonicalizer to drop it when the value is 1.
- The `dot_general` element-type constraint guarantees the scalar's type
  matches the reduce result, so no `convert` is needed.

### Tests

`test/lit_tests/dot_general_bcast_scalar.mlir` covers a broadcast of a constant
scalar (already folded by the pipeline) plus runtime scalars on the LHS and RHS
operands, both of which now simplify away the `dot_general`.

Before (runtime scalar, LHS):
```mlir
func.func @bcast_rank0_runtime(%s: tensor<f32>, %arg1: tensor<1024x32xf32>) -> tensor<24x32xf32> {
    %ones = stablehlo.broadcast_in_dim %s, dims = [] : (tensor<f32>) -> tensor<24x1024xf32>
    %r = stablehlo.dot_general %ones, %arg1, contracting_dims = [1] x [0], precision = [DEFAULT, DEFAULT] : (tensor<24x1024xf32>, tensor<1024x32xf32>) -> tensor<24x32xf32>
    return %r : tensor<24x32xf32>
}
```

After:
```mlir
func.func @bcast_rank0_runtime(%arg0: tensor<f32>, %arg1: tensor<1024x32xf32>) -> tensor<24x32xf32> {
    %cst = stablehlo.constant dense<0.000000e+00> : tensor<f32>
    %0 = stablehlo.reduce(%arg1 init: %cst) applies stablehlo.add across dimensions = [0] : (tensor<1024x32xf32>, tensor<f32>) -> tensor<32xf32>
    %1 = stablehlo.broadcast_in_dim %0, dims = [1] : (tensor<32xf32>) -> tensor<24x32xf32>
    %2 = stablehlo.broadcast_in_dim %arg0, dims = [] : (tensor<f32>) -> tensor<24x32xf32>
    %3 = stablehlo.multiply %1, %2 : tensor<24x32xf32>
    return %3 : tensor<24x32xf32>
}
```

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -9648,6 +9648,19 @@ struct BroadcastIotaSimplify
   }
 };
 
+
+
+// Returns the rank-0 scalar operand of a broadcast_in_dim that splats a scalar
+// across its entire result (empty broadcast_dimensions <=> rank-0 input), else
+// a null Value. The scalar may be a runtime value, hence a Value rather than an
+// attribute (cf. extractSplatInt).
+static Value extractBroadcastScalar(Value v) {
+  if (auto bcast = v.getDefiningOp<stablehlo::BroadcastInDimOp>())
+    if (bcast.getBroadcastDimensions().empty())
+      return bcast.getOperand();
+  return Value();
+}
+
 struct DotGeneralSimplify
     : public CheckedOpRewritePattern<stablehlo::DotGeneralOp,
                                      DotGeneralSimplify> {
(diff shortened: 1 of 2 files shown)


mofeing:
This is already implemented in #2216, which also handles more general cases like contraction of broadcasted dims on tensors (not only scalars).

PR #2661, opened 2026-07-23
Implement RegionBranchOpInterface for gpu_wrapper

pengmai:


test/lit_tests/diffrules/enzymexla/gpu_wrapper.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: enzymexlamlir-opt %s --enzyme --canonicalize --remove-unnecessary-enzyme-ops --enzyme-simplify-math --split-input-file | FileCheck %s
+// RUN: enzymexlamlir-opt %s --enzyme=dataflow --canonicalize --remove-unnecessary-enzyme-ops --enzyme-simplify-math --split-input-file | FileCheck %s
```
wsmoses: what if we ran both dataflow and not dataflow [with two run lines]?

wsmoses (approved):
lgtm regardless

test/lit_tests/diffrules/enzymexla/gpu_wrapper.mlir
pengmai: This is done

PR #2663, opened 2026-07-23
Try to use multithreaded

wsmoses:
@giordano x/ref https://github.com/julia-actions/cache/pull/244

Previous load was 5min on ci


wsmoses:
@giordano so this is the net combination of all my current julia cache prs, together we take ~5min -> ~1min for depot restore, so that's a significant win imo

PR #2664, opened 2026-07-24
Enable experimental remote repo contents cache

wsmoses:
Added experimental_remote_repo_contents_cache option to public and jll caches.

This hopefully avoids issues where github file access is slow or down

only in bazel 9+ it seems: https://github.com/bazelbuild/bazel/commit/b8589c3b278e3f5cee6ef85b0dcabb1cdcd69839


giordano:
I presume we need to add this file to https://github.com/EnzymeAD/Enzyme-JAX/blob/b4cf9532132e5dd046ad873bb3f95ef1e834cb5c/.github/workflows/build.yml#L7-L17 and/or https://github.com/EnzymeAD/Enzyme-JAX/blob/b4cf9532132e5dd046ad873bb3f95ef1e834cb5c/.github/workflows/build-reactantjll.yml#L18-L25 to actually run anything, and see if this works

wsmoses:
yeah we should independently add it to there

PR #2667, opened 2026-07-24
Macos: fixup tag

wsmoses:
@giordano so i think the issue is we added amdgpu by default in the list, but maybe didn't adjust this

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -80,7 +80,7 @@ jobs:
           echo $TAG
           sed -i.bak "s~version = \"[0-9.]*\"~version = \"$TAG\"~g" BUILD
           cat BUILD
-          if [[ "${{ matrix.os }}" == *"macOS"* || "${{ matrix.os }}" == *"macos"* ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/tag.yml
```diff
@@ -80,10 +80,16 @@ jobs:
           echo $TAG
           sed -i.bak "s~version = \"[0-9.]*\"~version = \"$TAG\"~g" BUILD
           cat BUILD
-          if [[ "${{ matrix.os }}" == *"x86"* ]]; then
-            sed -i.bak "s~\[\"X86\", \"AArch64\", \"AMDGPU\", \"NVPTX\"\]~\[\"X86\", \"NVPTX\"\]~g" workspace.bzl
+          if [[ "${{ matrix.os }}" == *"macOS"* || "${{ matrix.os }}" == *"macos"* ]]; then
```
giordano: The [`runner`](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context) context is more reliable to identify the macOS operating system, independently of how you spell the runs-on label
```suggestion
          if [[ "${{ runner.os }}" == "macOS" ]]; then
```

PR #2669, opened 2026-07-24
[enzyme] treat scf.if as a single statement for the purpose of atomic removal

ftynse:
This is conservatively correct in the polyhedral world.

Assisted-by: GPT 5.3

### DIFF
--- a/src/enzyme_ad/jax/Passes/RemoveAtomics.cpp
+++ b/src/enzyme_ad/jax/Passes/RemoveAtomics.cpp
@@ -339,74 +339,90 @@ bool isAccessRacy(ScopStmt &stmtA, MemoryAccess *accessA, ScopStmt &stmtB,
   return isAccessRacy(schedule.get_root().child(0), deps);
 }
 
+ScopStmt *getStatementContaining(IslScop &scop, Operation *op) {
+  for (ScopStmt &stmt : scop)
+    if (stmt.getOperation() == op || stmt.getOperation()->isAncestor(op))
+      return &stmt;
+  return nullptr;
+}
+
+MemoryAccess *getAtomicArrayWrite(enzyme::AffineAtomicRMWOp rmw,
+                                  ScopStmt &stmt) {
+  MemoryAccess *write = nullptr;
+  for (MemoryAccess *ma : stmt) {
+    if (ma->Kind != MemoryAccess::MT_Array || !ma->isWrite())
+      continue;
+    if (ma->AI->val != rmw.getMemref())
+      continue;
+    if (write)
+      return nullptr;
+    write = ma;
+  }
+  return write;
+}
+
 // Determines whether it is safe to "remove", i.e. convert an atomic rmw to
 // non-atomic read-modify-store.
 //
-// This is safe when there is no racy access w.r.t. to the rmw.
-bool isSafeToRemoveAtomicImpl(enzyme::AffineAtomicRMWOp rmw, IslScop &scop) {
+// This is safe when there is no other racy atomic rmw w.r.t. this rmw. Regular
+// load/store races are ignored because those races already exist in the input.
+bool isSafeToRemoveAtomicImpl(enzyme::AffineAtomicRMWOp rmw, IslScop &scop,
+                              ArrayRef<enzyme::AffineAtomicRMWOp> rmws) {
   LDBG() << "Handling rmw: " << rmw;
-  ScopStmt &rmwStmt = scop.getStatement(rmw);
-  MemoryAccess *theArrayWrite = nullptr;
-  for (MemoryAccess *ma : rmwStmt)
-    if (ma->Kind == MemoryAccess::MT_Array && ma->isMustWrite())
-      theArrayWrite = ma;
+  ScopStmt *rmwStmt = getStatementContaining(scop, rmw);
+  if (!rmwStmt) {
+    LDBG() << "Failed to find statement for rmw";
+    return false;
+  }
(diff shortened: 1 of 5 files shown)


test/lit_tests/remove_atomic_rmw.mlir
```diff
@@ -802,3 +802,44 @@ module {
...
+      affine.parallel (%arg2, %arg3) = (0, 0) to (4, 5) {
+        %2 = "enzymexla.pointer2memref"(%arg1) : (!llvm.ptr) -> memref<?xi8>
+        %3 = affine.load %2[%arg2] : memref<?xi8>
+        %4 = arith.extui %3 : i8 to i32
+        %5 = arith.andi %4, %c1_i32 : i32
+        %6 = arith.cmpi eq, %5, %c0_i32 : i32
+        %7 = scf.if %6 -> (f32) {
+          scf.yield %cst : f32
+        } else {
+          %9 = affine.load %0[%arg2 + %arg3 * 5 + 16] : memref<?xf32>
```
wsmoses, line 975: so this loads %0[%arg2 + %arg3 * 5 + 16]

the atomicrmw does

%0[%arg2 + %arg3 * 5 + 14] += ...

given that arg2 ranges from 0-4 I don't think it is legal to lower this atomic?

if the load was on a different aliasing result or the bounds were different it would be?

src/enzyme_ad/jax/Passes/RemoveAtomics.cpp
```diff
@@ -323,6 +323,15 @@ bool isSafeToRemoveAtomicImpl(enzyme::AffineAtomicRMWOp rmw, IslScop &scop) {
       if (!mayAlias(thisArray, theArray))
         continue;
 
+      // A may-write to an aliasing memref (e.g. effects summarized from
+      // conditional regions) is too imprecise to prove non-racy, so keep the
```
wsmoses: Shouldn't this be handled by the following two helpers?

```
      if (thisArray != theArray) {
        if (isAnyInstanceRacy(rmwStmt, stmt)) {
          LDBG() << "Found racy write to an aliasing array: "
                 << *stmt.getOperation();
          return false;
        }
      } else {
        if (isAccessRacy(rmwStmt, theArrayWrite, stmt, ma)) {
          LDBG() << "Found racy write to the same array: "
                 << *stmt.getOperation();
          return false;
        }
      }
```

wsmoses:
@copilot rebase this onto current main

wsmoses (approved):
manually rebased this, I don't totally follow how we check if the other is atomic or not?

we also have a non-affine atomic rmw too that we need to check against [and the llvm/memref one]

ftynse:
> manually rebased this, I don't totally follow how we check if the other is atomic or not?

We go through the list of all atomics around here: https://github.com/EnzymeAD/Enzyme-JAX/pull/2669/changes#diff-5cef3446c77396051782bed61d2497420fafee1d1fec8025ccd066f99f48fb1cR388

> we also have a non-affine atomic rmw too that we need to check against [and the llvm/memref one]

Those are "regular" side-effecting operations, we don't treat them specially, so they are already covered by conservative default. We only need special casing for normal load/store that we decided not preclude removal of atomics. I'll add a test.

There may be a slightly more advanced version of this where we understand other atomics but assume they "may-access" the entire memref + aliases instead of the specific subscript (typical polyhedral overapproximation), but I'll leave that for a follow-up.

test/lit_tests/remove_atomic_rmw.mlir
```diff
@@ -675,7 +787,6 @@ module {
 // CHECK:             affine.parallel (%[[VAL_1:.*]]) = (0) to (4) {
 // CHECK:               affine.parallel (%[[VAL_2:.*]]) = (0) to (4) {
 // CHECK:                 affine.store %[[ARG0]], %[[ARG1]]{{\[}}%[[VAL_1]], %[[VAL_2]] + 4] : memref<?x?xf32>
-// CHECK:                 %[[AFFINE_ATOMIC_RMW_0:.*]] = enzyme.affine_atomic_rmw addf %[[ARG0]], %[[ARG1]], (#[[$ATTR_7]]) {{\[}}%[[VAL_1]], %[[VAL_2]]] : (f32, memref<?x?xf32>) -> f32
```
wsmoses, line 678: we should restore these checks?

test/lit_tests/remove_atomic_rmw.mlir
```diff
@@ -25,7 +24,9 @@ module {
 // CHECK:               affine.store %[[CONSTANT_0]], %[[ARG4]]{{\[}}%[[VAL_2]]] : memref<?xf32>
 // CHECK:               %[[MULF_1:.*]] = arith.mulf %[[LOAD_2]], %[[ARG0]] : f32
 // CHECK:               %[[MULF_2:.*]] = arith.mulf %[[LOAD_2]], %[[LOAD_1]] : f32
-// CHECK:               %[[AFFINE_ATOMIC_RMW_0:.*]] = enzyme.affine_atomic_rmw addf %[[MULF_1]], %[[ARG2]], (#[[$ATTR_0]]) {{\[}}%[[VAL_2]]] : (f32, memref<?xf32>) -> f32
+// CHECK:               %[[LOAD_3:.*]] = affine.load %[[ARG2]]{{\[}}%[[VAL_2]]] : memref<?xf32>
```
wsmoses, line 27: can we do a test with an enzyme.affine atomic and enzyme.non affine atomic

PR #2694, opened 2026-07-28
raise math.atan2

Pangoraw:


wsmoses (approved):
double check the atan(y, x) convention is the same one between the math dialect and nv_atan2, otherwise lgtm

Pangoraw:
StableHLO Atan2 is atan2 from IEEE-754 (atan2(y,x) is atan(y/x))
The nv_atan2 does not really have a spec apart from the singularity __nv_atan2(0, 1) returns +0. but nv_atan2(y, x) is atan(y/x)
math.atan2(y,x) is also somehow underspeced but it's math.atan(y/x). so i think the raising is correct.

wsmoses:
cool sgtm [just wanted to double check]

PR #2695, opened 2026-07-28
Allow equivalent affine maps for for to while raising

Pangoraw:
with https://github.com/EnzymeAD/Enzyme-JAX/pull/2666, we exercise this path more often.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -1183,6 +1183,27 @@ static LogicalResult tryRaisingForOpToStableHLOUnroll(
   return success();
 }
 
+static bool equivUpToConstant(const affine::AffineValueMap &a,
+                              const affine::AffineValueMap &b) {
+  if (a.getOperands() != b.getOperands())
+    return false;
+
+  auto amap = a.getAffineMap(), bmap = b.getAffineMap();
+
+  if (amap.getNumDims() != bmap.getNumDims() ||
+      amap.getNumSymbols() != bmap.getNumSymbols() ||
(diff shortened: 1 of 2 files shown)


wsmoses (commented):
I don't follow, why is this safe?

Pangoraw:
once loaded, the constant offsets of a value's map are not used anywhere. just for each dimension, whether the stride (constant multiplied by the dimension) is positive/negative and whether or not the result is depending on a dimension.

```
    %0 = affine.load %arg0[27, %arg2 + 8, %arg3 + 8] : memref<36x112x208xf64, 1>
    %1 = affine.for %arg4 = 0 to 19 iter_args(%arg5 = %0) -> (f64) {
      %2 = affine.load %arg0[-%arg4 + 26, %arg2 + 8, %arg3 + 8] : memref<36x112x208xf64, 1>
      %3 = affine.load %arg1[-%arg4 + 19, %arg2, %arg3] : memref<20x96x192xf64, 1>
      %4 = arith.mulf %3, %arg5 {fastmathFlags = #llvm.fastmath<none>} : f64
      %5 = arith.subf %2, %4 {fastmathFlags = #llvm.fastmath<none>} : f64
      affine.store %5, %arg0[-%arg4 + 26, %arg2 + 8, %arg3 + 8] : memref<36x112x208xf64, 1>
      affine.yield %5 : f64
    }
```

here
```
%0: (%arg2, %arg3) -> (%arg2 + 8, %arg3 + 8)
%5: (%arg2, %arg3) -> (%arg2, %arg3)
```

if we were to alignMemory, it would be a no-op.

wsmoses (approved):
okay sure, can you add a comment saying that?

or even perhaps making the helper function called, equivalentUnderAlignMemory or something [with the explanation

PR #2696, opened 2026-07-28
fix WhileLoopInfo limit & step outer region check

Pangoraw:
the check was inverted, but working fine when the two regions are the
same.


src/enzyme_ad/jax/Implementations/WhileLoopInfo.cpp
```diff
@@ -191,29 +191,30 @@ int64_t WhileLoopInfo::getConstantNumIters() {
...
-                                      mapping.lookupOrDefault(start)),
-        stepVal);
   }
 
+  auto opReg = op->getParentRegion();
+  // Limit and Step are only usable here if they are defined in a region that
+  // encloses the while op -- a region of our own is not enough, that is the
+  // Condition/Block case.
+  if (!limit.getParentRegion()->isAncestor(opReg) ||
+      (step && !step.getParentRegion()->isAncestor(opReg))) {
```
Pangoraw, line 205: swapped the two regions here for the check to make sense (step / limit must dominate the op).

PR #2697, opened 2026-07-28
Add flag to disable speculation of partial if bodies

pengmai:
Disabling this pattern leads to LBM finding a significantly better min-cut solution. cc: @Pangoraw

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeLoops.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeLoops.cpp
@@ -889,6 +889,7 @@ public:
   }
 };
 
+template <bool Speculate>
 class PartialIfToSelect final : public OpRewritePattern<scf::IfOp> {
 public:
   using OpRewritePattern<scf::IfOp>::OpRewritePattern;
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/Passes/CanonicalizeLoops.cpp
```diff
@@ -908,6 +909,12 @@ class PartialIfToSelect final : public OpRewritePattern<scf::IfOp> {
     Block *elseBlock = ifOp.elseBlock();
     auto thenYield = cast<scf::YieldOp>(thenBlock->getTerminator());
     auto elseYield = cast<scf::YieldOp>(elseBlock->getTerminator());
+    // If speculation is disabled, only transform if ops that only contain
+    // yields
+    if (!Speculate && (thenBlock->getOperations().size() != 1 ||
```
wsmoses: I think we can do a bit better and we can just do this after the :

```
        if (!ifOp->isAncestor(op))
          return v;
```
in recomputeOutsideIf below. 

That way we can still if->select for the yielded values which are defined outside, even if it contains other values in the if statement that shouldn't be

wsmoses (approved):
can you add a quick test to make sure it has the desired behavior?

otherwise lgtm

PR #2699, opened 2026-07-29
Adds cuda backend support to MPI Allreduce via NCCL

romanlee:
This PR Adds cuda backend support to MPI Allreduce via ncclAllReduce.

We could either merge this PR roughly as is once we're happy with it, or wait until cuda support for the rest of the MPI ops is implemented. But I wanted to check in at this point to make sure we're happy with the design before moving forward.

In particular, changes were made to some core infrastructure:
* `LowerJIT.cpp`: Previously it was assumed that one only needed to replace `GetStreamOp` with the actual cuda stream when there was a kernel call, which were identified with criteria that don't catch the NCCL case. But we also need to make this replacement for the NCCL case, so we generalize the `LowerJITPass` to catch the NCCL case as well.
* `gpu.cc`: Previously it was assumed that there was always a `CuFunc`, which isn't true for the NCCL case. We generalize the `XLA_FFI_Handler`'s `initialize` and `execute` accordingly

A couple other design choices to note:
* Previously, we passed in the count via an op arg. Now we infer the count from the send/recv bufs, but therefore require those bufs to be statically shaped. 
* The address of the nccl communicator comes in via a pass option, which isn't the most robust but is probably ok for now

### DIFF
--- a/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
@@ -1,5 +1,6 @@
 // #include "mhlo/IR/hlo_ops.h"
 #include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
@@ -26,6 +27,44 @@ namespace enzyme {
 } // namespace mlir
 
 using namespace mlir;
+
+static FailureOr<int32_t> mapMPIToNCCLDatatype(enzymexla::MPIDatatype dt) {
+  switch (dt) {
+  case enzymexla::MPIDatatype::MPI_INT32_T:
+  case enzymexla::MPIDatatype::MPI_INT:
+    return 2; // ncclInt32
+  case enzymexla::MPIDatatype::MPI_UINT32_T:
+  case enzymexla::MPIDatatype::MPI_UNSIGNED:
+    return 3; // ncclUint32
+  case enzymexla::MPIDatatype::MPI_INT64_T:
+  case enzymexla::MPIDatatype::MPI_LONG_LONG_INT:
+    return 4; // ncclInt64
+  case enzymexla::MPIDatatype::MPI_UINT64_T:
+  case enzymexla::MPIDatatype::MPI_UNSIGNED_LONG_LONG:
+    return 5; // ncclUint64
+  case enzymexla::MPIDatatype::MPI_FLOAT:
+    return 7; // ncclFloat32
+  case enzymexla::MPIDatatype::MPI_DOUBLE:
+    return 8; // ncclFloat64
+  default:
+    return failure();
+  }
+}
+
+static FailureOr<int32_t> mapMPIToNCCLRedOp(enzymexla::MPIOp op) {
+  switch (op) {
+  case enzymexla::MPIOp::MPI_SUM:
+    return 0; // ncclSum
+  case enzymexla::MPIOp::MPI_PROD:
+    return 1; // ncclProd
+  case enzymexla::MPIOp::MPI_MAX:
+    return 2; // ncclMax
+  case enzymexla::MPIOp::MPI_MIN:
+    return 3; // ncclMin
+  default:
+    return failure();
+  }
+}
 
 struct MPICommRankOpLowering
     : public OpRewritePattern<enzymexla::MPICommRankOp> {
@@ -1374,9 +1413,10 @@ struct MPIAllreduceOpLowering
     : public OpRewritePattern<enzymexla::MPIAllreduceOp> {
 
   std::string backend;
-  MPIAllreduceOpLowering(std::string backend, MLIRContext *context,
+  size_t ncclCommPtr;
+  MPIAllreduceOpLowering(std::string backend, size_t ncclCommPtr, MLIRContext *context,
                          PatternBenefit benefit = 1)
-      : OpRewritePattern(context, benefit), backend(backend) {}
+      : OpRewritePattern(context, benefit), backend(backend), ncclCommPtr(ncclCommPtr) {}
 
   LogicalResult matchAndRewrite(enzymexla::MPIAllreduceOp op,
                                 PatternRewriter &rewriter) const override {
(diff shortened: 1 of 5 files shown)


src/enzyme_ad/jax/gpu.cc
```diff
@@ -131,10 +133,13 @@ XLA_FFI_Error *execute(XLA_FFI_CallFrame *call_frame) {
         &reinterpret_cast<XLA_FFI_Buffer *>(call_frame->args.args[i])->data;
   }
 
+  void *cufunc = nullptr;
   auto *execution_state = reinterpret_cast<xla::ffi::ExecutionState *>(
       internal_api->XLA_FFI_INTERNAL_ExecutionState_Get(
           ctx, XLA_FFI_ExecutionStage_INITIALIZE));
-  auto cufunc = (void *)execution_state->Get<CuFuncWrapper>().value();
+  if (execution_state != nullptr && execution_state->IsSet()) {
```
wsmoses: So I don't think this is actaully needed/the right way to do things, essentially here we set up an ffi to always call a function with also an init. We can always leave things as the empty fn if desired, or here just pass nullptr from init to execute if there's no shared initialization we need to preserve

romanlee: Ok let me investigate this

romanlee: Ok, I've reverted gpu.cc, and now we handle the creation of an init for the nccl case upstream in LowerJIT.cpp. How does that look?

Also, in LowerJIT.cpp, we previously were branching on `backend==cuda` (aka the parameter `useCudaABI`) in LowerJITPass to determine if we needed to do cuda-y things (setting up the stream and the init). My fear was that this was too general, however, so in [07e01ff](https://github.com/EnzymeAD/Enzyme-JAX/pull/2699/commits/07e01ffc5fce12e1de5d1a6362935077605ce6e4) I narrowed the scope to branch on a `require_cuda_abi` attr of the wrapper func. Let me know your thoughts on this/if you would prefer a different way of doing things.

src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
```diff
@@ -1729,7 +1903,100 @@ struct LowerEnzymeXLAMPIPass
...
+          << "lower-enzymexla-mpi must run on a builtin.module";
+      signalPassFailure();
+      return;
+    }
+
+    auto context = module->getContext();
+
+    if (backend == "cuda" && !ncclCommPtr) {
+      bool hasLowerableMPIOp = false;
+      module.walk([&](Operation *op) {
```
wsmoses: slightly easier, rather than having a captured bool you can return interrupt if it is something desired, and advance otherwise, and the result of module.walk().interrupted() will say if it has something bad

src/enzyme_ad/jax/Passes/LowerJIT.cpp
```diff
@@ -1094,7 +1137,8 @@ struct LowerJITPass
           symbolTable, op.getLoc(), fn, jit, op, openmp, cuResultHandlerPtr,
           cuStreamSynchronizePtr, indexBitWidth, cubinTriple, cubinChip,
           cubinFeatures, cubinFormat, cuOptLevel, toolkitPath, linkFilesArray,
-          debug, hasReturn, dump_final_module);
+          debug, hasReturn, dump_final_module,
+          fn->hasAttr("enzymexla.requires_cuda_abi"));
```
wsmoses: I think we can avoid this

wsmoses: actually fine i think we go for this for now, but change this to "enzymexla.device_abi" as the attr name and "cuda" as the value, rather than all in the name

src/enzyme_ad/jax/Passes/LowerJIT.cpp
```diff
@@ -426,6 +439,46 @@ CallInfo CompileHostModule(std::string &key, mlir::ModuleOp modOp,
...
+  auto helper = LLVM::LLVMFuncOp::create(builder, loc, "EnzymeJaXEmptyGPUInit",
+                                         initTy, LLVM::Linkage::External);
+  auto initfn = LLVM::LLVMFuncOp::create(builder, loc, "nv_func_init", initTy,
+                                         LLVM::Linkage::External);
+
+  auto blk = new Block();
+  initfn.getRegion().push_back(blk);
+  builder.setInsertionPointToEnd(blk);
+  SmallVector<mlir::Value> args;
+  auto cufunc = LLVM::CallOp::create(builder, loc, helper, args)->getResult(0);
```
wsmoses: why bother having the EnzymeJaXEmptyGPUInit for now if it's empty, you can just return nullptr itself?

romanlee: I don't think you can directly return nullptr because that then breaks inside gpu.cc — either in setting or getting the execution state as I recall (can try to repro but having hydra issues currently). Note currently we're not passing a nulltpr directly, we're passing a non-null ptr to an empty Cufunc wrapper.

romanlee: Alright yeah, if we try to do something like in https://github.com/EnzymeAD/Enzyme-JAX/pull/2699/commits/f461cbbc3355d5e3dc07f50ac33cedf9ad2f88b1 we get this error
```
  error copying to host: UNKNOWN: XLA FFI call failed: Bad StatusOr access: NOT_FOUND: State is not set [executable_name='reactant_Allreduce_']
  Type: Int32: Test Failed at /mnt2/romanlee/codes/Reactant.jl/test/integration/mpi.jl:165
    Expression: Array(result) == expected
     Evaluated: Int32[1, 0, 0, 0, 608408720] == Int32[2, 2, 2, 2, 2]
```
ie I think when cufunc is null, the execution state doesn't actually get set, and then this fails when you try to get it during execution

src/enzyme_ad/jax/Passes/LowerEnzymeXLAMPI.cpp
romanlee: Thanks, done

src/enzyme_ad/jax/Passes/LowerJIT.cpp
wsmoses: okay i think that's just that xla doesn't support the nullptr there. How about returning inttoptr 1 to ptr?

romanlee: Yeah I agree.

Returning inttoptr 1 to ptr like this https://github.com/EnzymeAD/Enzyme-JAX/pull/2699/commits/2f583bfd0cc99fe061877ff61837bfb6ae703ae5 ?  That results in a segfault in `__libc_free`. 

Are you just trying to avoid having the extra/possibly not necessary `EnzymeJAXEmptyGPUInit` func definition and associated plumbing? I think the above segfault means that XLA is ultimately going to try and `delete` anything we pass it through `Set`, so I'm not sure we can get away with avoiding `EnzymeJaxEmptyGPUInit` without doing something pretty weird looking in `insertEmptyGPUInit`?

romanlee: Fixed in b24492f0f8df92a0a43c914ce86162af65fb76b1

wsmoses (approved):

PR #2700, opened 2026-07-29
Tessera: result/output arguments

jessicacotturone21:
Add support for output arguments, like those used in MPFR. Can now mark an argument as being a result of the operation in the `tessera_op` annotation and it will be treated as a return value of the tessera.call (in addition to any existing return values) to allow for pattern matching. 

The marking logic has been changed so that instead of just specifying an argument as "byref," we now give an argument the marker `:val` to indicate that the argument needs to be lifted from a pointer to a value in the tessera raising pass, and then we can specify the direction as either in, inout, or out (e.g. `rop:val=out`). Each of these 3 cases have a different behavior in the tessera raising pass.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
@@ -10,6 +10,8 @@
 #include "mlir/Interfaces/FunctionInterfaces.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 
+#include <optional>
+
 using namespace mlir;
 using namespace mlir::enzyme::tessera;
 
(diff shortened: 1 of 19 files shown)


src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
```diff
@@ -140,52 +142,176 @@ Attribute DefineOp::getSretAttr() {
...
-  assert(argIdx < types.size() && "argIdx out of bounds in getByRefType");
-  auto attr = types[argIdx];
-  if (!isa<TypeAttr>(attr))
+// The direction string of a lifted argument, or empty if it is not lifted.
+static StringRef getArgDir(DefineOp op, unsigned argIdx) {
+  auto dict = getArgModeDict(op, argIdx);
+  if (!dict)
+    return StringRef();
+  if (auto dirAttr = dyn_cast_or_null<StringAttr>(dict.get(argModeDirField)))
+    return dirAttr.getValue();
```
wsmoses, line 208: do you want to check here that dirAttr is a string you expect [maybe even converting to an enum for downstream], and not just "foobar"

src/enzyme_ad/jax/Dialect/Tessera/Ops.cpp
```diff
@@ -140,52 +142,176 @@ Attribute DefineOp::getSretAttr() {
...
+    auto typeAttr = dyn_cast_or_null<TypeAttr>(dict.get(argModeTypeField));
+    if (!typeAttr)
+      return emitOpError("argModes entry ")
+             << i << " must have a '" << argModeTypeField << "' TypeAttr";
+
+    auto dirAttr = dyn_cast_or_null<StringAttr>(dict.get(argModeDirField));
+    if (!dirAttr)
+      return emitOpError("argModes entry ")
+             << i << " must have a '" << argModeDirField << "' StringAttr";
+
```
wsmoses, line 291: minor, if we make this an enum [or enumattr] then

wsmoses (approved):

PR #2713, opened 2026-07-30
Fix affine.if yield captures in isValidSymbolInt (+ then/else guard)

wsmoses:
Follow-up to #2712, which made `getAllOps` see the captures on an if's terminator when choosing where to hoist. That exposed a latent crash on a neighbouring input. Two commits.

### 1. Read the else yield in `MoveSelectToAffine`'s and-operand guard

The guard that skips already-canonical `affine.if` booleans when splitting an `and`/`or` condition read both `tval` and `fval` from `getThenBlock()`, so `tval == fval` always and neither `continue` condition (One/Zero, Zero/One) could ever hold. Such operands got rewritten on every sweep, so `applyPatternsGreedily` never converges — a silent `signalPassFailure()` with no diagnostic (bare `exit 1`, no output).

Latent on `main`: the first conversion loop handles these conditions whole, so the and/or fallback is not reached with `affine.if` operands. Commit 2 makes it reachable.

### 2. Account for terminator captures in `isValidSymbolInt`

`isValidSymbolInt` judges an `scf.if`/`affine.if` by its condition/operands plus its body ops via `without_terminator()` — the same blind spot #2712 fixed in `getAllOps`. An `affine.if` whose only outside references are on its `affine.yield` therefore looks like a valid affine symbol, so `legalCondition`/`isValidIndex` green-light raising a conditional keyed off it. `AffineApplyNormalizer::fix` then correctly reports it cannot legalize that symbol by returning null, but the caller routes null through `llvm_unreachable("cannot move")` — UB under `-c opt`, where the compiler drops the null check and passes a null `Value` straight to `isValidSymbolInt`:

```
Program received signal SIGSEGV
0x... in isValidSymbolInt(mlir::Value, bool, mlir::Region*)
rdi 0x0     <- null Value
#1  AffineApplyNormalizer::AffineApplyNormalizer(...)
#4  MoveIfToAffine::matchAndRewrite(mlir::scf::IfOp, mlir::PatternRewriter&)
```

This checks the yielded values too, so the legality predicate and the hoister agree. The conditional is then left as an `scf.if`, which is correct — in the new test it is keyed off a value yielded out of an `if` on the induction variable, i.e. a dim, never a symbol.

Note dropping `without_terminator()` outright does not work: the body loop uses the `Operation*` overload, which means "is this op's result a symbol", and a yield has no results and is not in the whitelist — every `scf.if`/`affine.if` would fail the check. Confirmed: that variant regresses `affinecfg_noselect` and `raiseimmerse`.

### Behaviour on the two reduced inputs

| build | `select_hoist_capture` | `select_yield_dim` |
|---|---|---|
| before #2712 | dominance error | dominance error |
| #2712 only | ok | **segfault** |
| with this | ok | ok |

`affinecfg_select_hoist_capture`'s expectations are updated: the two selects now merge into one `affine.if` that stays beside the `scf.if` it yields, and the inner conditional stays an `scf.if` since its `scf.if` is keyed off a non-affine `ptrtoint`.

### Testing

`bazel test //test/lit_tests/... --cache_test_results=no`: 1083/1086, and 1082/1085 for commit 1 alone. The 3 failures are pre-existing on `main` and unrelated (uncommitted WIP tests in my working tree: `communication_dusN`, `communication_dusS`, `raising_affine_to_stablehlo_ab_tend`).

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -2451,7 +2451,7 @@ struct MoveSelectToAffine : public OpRewritePattern<arith::SelectOp> {
           auto idx = cast<OpResult>(opv.get()).getResultNumber();
           auto tval = cast<AffineYieldOp>(midIf.getThenBlock()->getTerminator())
                           .getOperand(idx);
-          auto fval = cast<AffineYieldOp>(midIf.getThenBlock()->getTerminator())
+          auto fval = cast<AffineYieldOp>(midIf.getElseBlock()->getTerminator())
                           .getOperand(idx);
           if (matchPattern(tval, m_One()) && matchPattern(fval, m_Zero()))
             continue;
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -100,15 +100,7 @@ bool isValidSymbolInt(Operation *defOp, bool recur, Region *scope) {
     }
     if (auto ifOp = dyn_cast<scf::IfOp>(defOp)) {
       if (isValidSymbolInt(ifOp.getCondition(), recur, scope)) {
-        if (llvm::all_of(ifOp.thenBlock()->without_terminator(),
-                         [&](Operation &o) {
-                           return isValidSymbolInt(&o, recur, scope);
-                         }) &&
-            llvm::all_of(ifOp.elseBlock()->without_terminator(),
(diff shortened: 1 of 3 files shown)


wsmoses:
Marking draft: my `bazel test` verification was wrong. The "1083/1086" run reused cached results (only 3 tests actually executed). With `--cache_test_results=no` this change regresses two tests — `raising_affinecfg_noselect` and `raising_raiseimmerse` — both by making `--affine-cfg` fail to converge (silent `signalPassFailure` from `applyPatternsGreedily`). Investigating.

PR #2720, opened 2026-08-01
Update default pipeline

pengmai:
* configure round trip options struct
* symbol-dce before and after gpu-kernel-outlining
* lower-affine before differentiation

@Pangoraw the additional `symbol-dce` resolves the GPU module without target issue.

### DIFF
--- a/src/enzyme_ad/jax/raise.cpp
+++ b/src/enzyme_ad/jax/raise.cpp
@@ -11,6 +11,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "src/enzyme_ad/jax/raise.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Pass/PassRegistry.h"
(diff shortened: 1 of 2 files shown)


wsmoses (approved):
Can you turn split multi results and pre AD lower affine into options too, otherwise LGTM

src/enzyme_ad/jax/raise.cpp
```diff
@@ -131,15 +133,26 @@ extern "C" std::string runLLVMToMLIRRoundTrip(std::string input,
...
+
+      pass_pipeline += "enzyme{";
+      if (options->dataflow)
+        pass_pipeline += "dataflow ";
+      if (options->markReadonly)
+        pass_pipeline += "markReadonly";
+      pass_pipeline += "},canonicalize,";
+      if (options->splitMultiResults)
+        pass_pipeline += "split-multi-results,";
+      pass_pipeline += "remove-unnecessary-enzyme-ops,flatten-enzyme-caches,enzyme-simplify-math,"
```
wsmoses, line 148: in a follow up we should perhaps consider doing the same for flatten-enzyme-caches if useful to compare with

wsmoses (approved):

PR #2722, opened 2026-08-01
Fix affine if handling

wsmoses:


wsmoses:
@dkytezab this should fix the thing you hit [i think]

PR #2725, opened 2026-08-01
[Polymer] Bound the scf.if ancestor check to the scop in IslScopBuilder

### LINKED ISSUE #2724: Atomic removal for LBM
Module: https://gist.github.com/pengmai/d45d16a0db500413f634861cd4e8d9ea
Passes: `--affine-cfg --remove-atomics`

Fails with the message:
```
stmt not found
UNREACHABLE executed at src/enzyme_ad/jax/polymer/mlir/lib/Conversion/Polymer/Support/IslScop.cc:347!
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace and instructions to reproduce the bug.
```

wsmoses:
Fixes #2724.

`--remove-atomics` crashed with `stmt not found` / `UNREACHABLE executed at IslScop.cc:347` on the LBM kernel from the issue.

### Root cause

`IslScopBuilder::build`, under `allowScfIfConditionalWritesAsMay` (which `--remove-atomics` passes as `true`), drops statements nested in an `scf.if` — the `scf.if` has a `ScopStmt` of its own that models their effects, and counting both would double-count the accesses:

```cpp
llvm::erase_if(scop->stmts, [&](ScopStmt &stmt) {
  return stmt.getOperation()->getParentOfType<scf::IfOp>() != nullptr;
});
```

`getParentOfType<scf::IfOp>()` walks ancestors **without stopping at the scop root**. In the reported module the scop root — the `enzymexla.gpu_wrapper` — is itself nested:

```
func.func -> scf.if -> scf.for -> enzymexla.gpu_wrapper   <- scop root
```

so every statement reported an `scf.if` ancestor: the one wrapping the entire kernel, which is outside the scop and has no `ScopStmt`. `gatherStmts` built 454 statements and `erase_if` deleted all 454.

The crash is the downstream symptom. `ScopStmt`s constructor stamps `polymer.stmt.name` onto its op, and erasing from the `stmts` vector does not remove that attribute. `buildSequenceSchedule` dispatches on exactly that attribute, so it called `buildLeafSchedule` on an op whose statement no longer existed, and `getIslStmt` hit the `llvm_unreachable`.

### Fix

Bound the ancestor search to the scop, so an `scf.if` only suppresses nested statements when it is itself inside the scop.

On the reported kernel the scop goes from 0 statements back to 64 (the genuinely `scf.if`-nested ones are still dropped, as intended), so the dependence analysis runs on real data instead of crashing, and all 38 `atomic_rmw`s are removed.

### Test

New case in `test/lit_tests/remove_atomic_rmw.mlir` — a `gpu_wrapper` inside `scf.for` inside `scf.if`, mirroring the nesting from the issue. Verified it is a real regression test: with the fix reverted it fails with exactly `stmt not found` / `UNREACHABLE ... IslScop.cc:347`. It also asserts the atomic is still removed, so it would catch a "fix" that dodged the crash by leaving the scop empty.

Full `bazel test -c dbg //test/lit_tests/...` shows no regressions.

### DIFF
--- a/src/enzyme_ad/jax/polymer/mlir/lib/Conversion/Polymer/Support/IslScop.cc
+++ b/src/enzyme_ad/jax/polymer/mlir/lib/Conversion/Polymer/Support/IslScop.cc
@@ -1845,9 +1845,13 @@ IslScopBuilder::build(Operation *f, bool allowScfIfConditionalWritesAsMay) {
 
   if (allowScfIfConditionalWritesAsMay) {
     // The enclosing scf.if statement models effects for its nested ops.
-    // Drop nested statements here to avoid double-counting accesses.
+    // Drop nested statements here to avoid double-counting accesses. Only an
+    // scf.if inside the scop has a statement of its own to do that modelling:
+    // one enclosing the whole scop is not part of it, and treating it as the
+    // modelling statement would drop every statement we just gathered.
     llvm::erase_if(scop->stmts, [&](ScopStmt &stmt) {
-      return stmt.getOperation()->getParentOfType<scf::IfOp>() != nullptr;
+      auto ifOp = stmt.getOperation()->getParentOfType<scf::IfOp>();
+      return ifOp && f->isProperAncestor(ifOp);
     });
   }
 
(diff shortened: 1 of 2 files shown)


wsmoses:
@ftynse this fixes the issue @pengmai was hitting. I'm going to go ahead and merge as it seems reasonable, but please review post commit

ftynse (approved):
Sounds plausible

PR #2740, opened 2026-08-03
Extend CSE coverage to batch_norm, unary math, convolution, fft and slicing

wsmoses:
## Problem

`CSE<T>` (`EnzymeHLOOpt.cpp`) is registered for an explicit list of op types. Ops off that list have **no working CSE at all**, because the only fallback is MLIR's own CSE — and that compares the full discardable attribute dictionary.

That matters here because the `enzymexla.{non_negative,no_nan,finite,symmetric_matrix,...}` attributes are *memoized analysis results*: `GuaranteedResultAnalysisBase::setGuaranteedInIR` writes the conclusion back onto the defining op so later queries can read it via `lookupGuaranteedFromIR`. Once an analysis has queried one of two otherwise identical ops, MLIR's CSE considers them inequivalent.

`CSE<T>` passes `OperationEquivalence::IgnoreDiscardableAttrs` and is immune — so **coverage is the only thing that matters**.

## Where it showed up

`grad(DGCNN)` from Reactant.jl's `benchmark/nn` contained three copies of an 8.4M-element `batch_norm_training` over the same data:

```mlir
%25        : tensor<64x131072xf32>
%output_20 = batch_norm_training(%25, %arg7, %arg8) <{feature_index = 0 : i64}>
%26        = transpose %25, dims = [1,0]
%output_23 = batch_norm_training(%26, %arg7, %arg8) <{feature_index = 1 : i64}>
%output_26 = batch_norm_training(%25, %arg7, %arg8) <{feature_index = 0 : i64}>
             {enzymexla.non_negative = [#enzymexla<guaranteed NOTGUARANTEED>, ...]}
```

`%output_20` and `%output_26` have byte-identical operands and properties. They differ **only** by the cached analysis result — which records a *negative* answer, so it carries no information whatsoever — and were still not merged.

Adding `batch_norm_training` takes that module from **8 → 6** `batch_norm_training` ops.

The remaining pair (`%output_23`) is the same normalization expressed through a transpose: batchnorm over `feature_index=0` of `%25` and over `feature_index=1` of `transpose(%25)` are by definition identical. Removing that needs a transpose/batch_norm commutation pattern rather than CSE, and is left for a follow-up.

## What's added

`batch_norm_{training,inference,grad}`, the unary math ops (`exp`, `expm1`, `log`, `log1p`, `tanh`, `logistic`, `sqrt`, `rsqrt`, `cbrt`, `sine`, `cosine`, `atan2`, `floor`, `ceil`, `round`, `sign`, `is_finite`), `clamp`, `remainder`, the bitwise ops, `convolution`, `fft`, `dynamic_slice`, `dynamic_update_slice` and `reverse`.

## Testing

- New lit test `cse_guaranteed_analysis_attrs.mlir` covering the batchnorm case, unary math, and a negative control (differing `epsilon` must still block CSE).
- All 1090 lit tests pass.
- Three autobatching goldens updated. **All three were encoding a missed CSE:**
  - `higher_order_post_diff.mlir` had two literally identical `stablehlo.logistic %17` ops.
  - `elementwise_loop.mlir` had a duplicated `dynamic_slice`+`slice` chain. Autobatching hoists the `cosine` and the `sine` out of the loop, and each hoist independently re-materialises the chain feeding it:

    ```
    before                                after
      %0 = dynamic_slice %arg0, %c_3        %0 = dynamic_slice %arg0, %c_3
      %1 = slice %0 [0:10:3]                %1 = slice %0 [0:10:3]
      %2 = cosine %1                        %2 = cosine %1
      %3 = dynamic_slice %arg0, %c_3        %3 = sine %1
      %4 = slice %3 [0:10:3]
      %5 = sine %4
    ```
  - `concatreducebatching.mlir` is a renumbering only (same op count).


wsmoses:
for obvious reasons this impacts both mlir and raising papers

vimarsh6739 (approved):

PR #2744, opened 2026-08-03
Fuse XLA wrappers and lift host loops into StableHLO

vimarsh6739:
LBM's raised updates execute as separate XLA calls inside a host time loop. This pass combines compatible sequential wrappers and moves eligible `scf.for` and `affine.for` loops into one StableHLO function containing a `stablehlo.while`.

Alias analysis must prove that the wrappers refer to the same pairwise-disjoint buffers. Loop lifting requires an unused induction variable and otherwise pure operand setup. Static bounds become tensor constants; dynamic bounds are copied through registered temporary device allocations. Affine lower/upper bound maps preserve max/min semantics and non-unit steps, and SCF signed/unsigned comparisons are preserved.

This produces one XLA execution for the loop. GPU kernel fusion remains a backend decision: the current LBM 20-step run launches 441 GPU kernels (43 body kernels across 10 iterations, plus 11 condition evaluations). The current XLA GPU backend excludes `while` from ordinary fusion, so a single CUDA kernel needs further backend work.

Validation:

- Built `//:enzymexlamlir-opt` and `//:reactant-clang` against the local staging sources.
- Four megakernelization LIT tests pass, covering sequential fusion, SCF loops, affine bounds and rejection cases, and the fresh LBM host loop shape and runtime lowering.
- Ran `Enzyme-GPU-Tests/run_lbm.sh xla-gpu 20` on an RTX 5090. Generated MLIR contains one wrapper around both updates and the full time loop; the host entry point has no time-loop backedge.
- Compared all 6,480,000 output float32 values against CUDA: maximum absolute difference `5.958245310466737e-7`, zero values outside `1e-6 + 1e-5 * abs(reference)`.
- Confirmed the actual GPU launch count with Nsight Systems.

### DIFF
--- a/src/enzyme_ad/jax/Passes/XLAMegakernelize.cpp
+++ b/src/enzyme_ad/jax/Passes/XLAMegakernelize.cpp
@@ -1,4 +1,4 @@
-//===- XLAWrapperMegakernelize.cpp - Fuse raised XLA wrappers ------------===//
+//===- XLAMegakernelize.cpp - Fuse raised XLA wrappers -------------------===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -23,7 +23,7 @@
 
 namespace mlir {
 namespace enzyme {
-#define GEN_PASS_DEF_XLAWRAPPERMEGAKERNELIZEPASS
+#define GEN_PASS_DEF_XLAMEGAKERNELIZEPASS
 #include "src/enzyme_ad/jax/Passes/Passes.h.inc"
 } // namespace enzyme
 } // namespace mlir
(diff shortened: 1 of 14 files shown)


src/enzyme_ad/jax/Passes/XLAMegakernelize.cpp
```diff
@@ -0,0 +1,262 @@
...
+    return false;
+
+  Block &body = function.getBody().front();
+  auto returnOp = dyn_cast<func::ReturnOp>(body.getTerminator());
+  return returnOp && returnOp.getNumOperands() == type.getNumResults();
+}
+
+/// Fuse two wrappers separated only by memory-effect-free bookkeeping. The
+/// initial rule deliberately requires alias analysis to prove that both
+/// wrappers cover exactly the same unique, pairwise-disjoint buffers. This
```
wsmoses, line 143: we don't need the two to cover the same unique parwise-disjoint buffers. We simply need to prove that for all buffers a in A it is either guaranteed to not alias any buffer b in B or is guaranteed to alias a buffer b [and thus we can pass a single buffer for both args, or a new arg]

PR #2770, opened 2026-08-04
canonicalize-scf-for: a call that stays keeps what it names

wsmoses:
Fixes the LBM build failure on https://github.com/EnzymeAD/Reactant/pull/65:

```
error: 'llvm.call' op '_ZZ26CUDA_LBM_kernel_loop_inneriPfS_E19__enzyme_set_mincutm'
       does not reference a symbol in the current scope
```

`ForLoopApplyEnzymeAttributes` reads the setting off an `__enzyme_set_mincut` / `__enzyme_set_checkpointing` call, puts it on the enclosing loop, and erases both the call and the declaration. The mincut branch `continue`d after setting the attribute, skipping `rewriter.eraseOp(user)` — but `rewriter.eraseOp(func)` at the end ran unconditionally. So whenever a mincut call was still inside an `scf.for`/`scf.while` when the pattern fired, the declaration went and the call stayed, pointing at nothing.

Reduced to twelve lines, on main before this change:

```mlir
scf.for %i = %c0 to %n step %c1 : i32 {
  llvm.call @_Z26__enzyme_set_mincutm(%m) : (i64) -> ()
}
llvm.func @_Z26__enzyme_set_mincutm(i64)
```
→ `error: 'llvm.call' op ... does not reference a symbol in the current scope`

The same shape is reachable from the non-constant path, which also `continue`s. So the declaration is now erased only once every call to it has been, and a use the pattern could not read keeps both.

This is not new to the commit range being bumped — it needs only that the call still be in a recognised loop at that point, and the existing `raise-loop-attributes.mlir` covers checkpointing but never mincut. `mincut_attr.mlir` adds the off / on / dynamic cases. Suite is 1103/1103.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -3448,9 +3448,15 @@ struct ForLoopApplyEnzymeAttributes
       return success();
     }
 
+    // The calls whose attribute has been read out and put on the loop, which
+    // are the ones there is no longer anything to say through.
+    SmallVector<Operation *> read;
+    unsigned called = 0;
+
     for (auto use : *uses) {
       auto user = use.getUser();
       assert(isa<LLVM::CallOp>(user));
+      called++;
 
       APInt ckptType;
       if (!matchPattern(user->getOperand(0), m_ConstantInt(&ckptType))) {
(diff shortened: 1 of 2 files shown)


wsmoses:
Confirmed end-to-end against the configuration CI builds.

Reproduced locally first — my earlier "builds fine" was against a stale `libClangReactant-24.so` that predated the `enzyme_set_mincut` attribute registration, so the call was never emitted. With the plugin rebuilt, LBM fails with exactly the CI error:

```
error: 'llvm.call' op '_ZZ26CUDA_LBM_kernel_loop_inneriPfS_E19__enzyme_set_mincutm'
       does not reference a symbol in the current scope
lbm.cu:120:6: error: Reactant: failed to run mlir passes
```

With this change, `make -B FORWARD=no -j8` in `LBM/` completes clean (rc=0, no errors), which is the `MINCUT=yes` default CI uses.

One thing surfaced behind it: `MINCUT=no` now gets past the dangling symbol and fails later with `'affine.load' op operand cannot be used as a dimension id`. That is a separate issue in the disable-mincut path, previously masked by this one — it is not exercised by CI, which builds the default.

Pangoraw (approved):

PR #2786, opened 2026-08-06
Hand the round-tripped module back as bitcode

wsmoses:
The module crosses the boundary between the Reactant pass and `runLLVMToMLIRRoundTrip` as bytes either way — the interface is `extern "C" std::string runLLVMToMLIRRoundTrip(std::string input, ...)`. It was being spelled as textual LLVM IR at both ends, which means four full print/parse cycles per compile:

| step | where |
|---|---|
| print module → string | `Enzyme.cpp` `raw_string_ostream ss(MStr); ss << M;` |
| `parseIR` string → module | `raise.cpp` |
| print module → string | end of `runLLVMToMLIRRoundTrip` |
| `parseIR` string → module | `Enzyme.cpp`, then `Linker::linkInModule` |

Bitcode is the cheaper and more faithful spelling. On MFEM's dFEM tests, the same module measured:

| | textual | bitcode |
|---|---|---|
| size | 46.3 MB | **8.0 MB** (5.8×) |
| parse (full materialization) | 0.90 s | **0.47 s** |
| parse + write | 1.15 s | 0.76 s |

More to the point, it does not depend on a printer and a parser agreeing about syntax — which a textual hand-off does depend on, and which matters here because the plugin and `libRaise.so` are built separately and connected through `dlsym`.

**Compatibility is free in both directions.** The reader on each side is `llvm::parseIR`, which sniffs the bitcode magic and dispatches to `parseBitcodeFile`:

```cpp
if (isBitcode((const unsigned char *)Buffer.getBufferStart(),
              (const unsigned char *)Buffer.getBufferEnd())) {
  ... parseBitcodeFile(Buffer, Context, Callbacks);
}
return parseAssembly(...);
```

So only the two writers change; a new Reactant reads an old `libRaise`, and an old Reactant reads a new one.

**This is not a compile-time change, and I want to be clear about that.** Controlled A/B on MFEM's `tests/unit/dfem/test_diffusion.cpp`: **279.09 s bitcode vs 279.05 s textual**. The serialization is ~2 s of that compile. The round trip's cost is elsewhere — for a 675 s compile of `test_functional_gradient.cpp` it decomposes as ~70 s LLVM→MLIR translation, 420 s MLIR pass pipeline (220 s of it canonicalize), ~80 s MLIR→LLVM translation, and ~2 s of serialization.

Verified: the plugin and `libRaise.so` build, a full round trip through bitcode produces correct code (MFEM `ParMesh` construction gives the same result as an unplugged build), and MFEM dFEM TUs compile unchanged.


wsmoses:
@copilot fix the format

PR #2787, opened 2026-08-06
while: Use generic checkpointing implementation

Pangoraw:
Requires https://github.com/EnzymeAD/Enzyme/pull/3092.


src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
```diff
@@ -720,640 +636,571 @@ class AutoDiffWhileRev
...
+  //===--------------------------------------------------------------------===//
+
+  static StringRef enableCheckpointingAttrName() {
+    return "enzymexla.enable_checkpointing";
+  }
 
-    indexTensor =
-        setIndex(builder, indexTensor, capo, innerBody->getArgument(0));
+  static StringRef binomialCheckpointingAttrName() {
+    return "enzymexla.binomial_checkpointing";
```
wsmoses: in the future can we just make this an enzyme rather than enzymexla attr [for consistency throughout]

wsmoses (approved):
lgtm, but see follow up attr comment

Pangoraw:
Reactant integration failing is expected due to missing https://github.com/EnzymeAD/Reactant.jl/pull/3180 which we should merge as we update the E-JAX commit.

PR #2799, opened 2026-08-07
polygeist-mem2reg: nocapture does not say the callee will not write

wsmoses:
`forwardStoreToLoad` dropped a call from the clobber set as soon as the slot was passed at a `nocapture` position. nocapture only says the callee does not hold on to the pointer; an out parameter is nocapture **and written through** before the call returns. The slot's value is unknown after such a call — only `readonly` (or `readnone`) alongside it lets a store be forwarded past.

Also require the attribute at every position the value is passed, not just some position that happens to carry it.

Found via MFEM: `ParMesh::ParMesh(comm, Mesh&, ...)` passes a `Table*` slot to `BuildLocalBoundary(..., Table *&)` and `FindSharedEdges(..., Table *&)`. Forwarded across both, the constructor read back its stale initial pointer instead of the Table the callees built, and the AMGF solver test crashed dereferencing it.

The forwarding cases in `mem2reg_capturing.mlir` now say `llvm.readonly` where they mean it; the new case is the out parameter.

### DIFF
--- a/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
+++ b/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
@@ -2090,13 +2090,18 @@ bool isCallNonCapturing(CallOpInterface callOp, Value val,
                              LLVM::LLVMDialect::getNoCaptureAttrName());
 }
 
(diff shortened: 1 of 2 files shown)


wsmoses:
@copilot fix the format

PR #2802, opened 2026-08-07
llvm-to-affine-access: distribute a select over both addends, not one

wsmoses:
`SelectCSE` turns `select(c, a+b, x+y)` into `select(c,a,x) + select(c,b,y)`. The second select was built from the false side's operand **twice** — `select(c, y, y)` — so whichever way the condition went, the false branch's addend was the one added.

MFEM's Hilbert-curve element ordering (`NCMesh::GridSfcOrdering3D`), all select chains over direction deltas, came out as a different curve on every odd-sized grid; `Make3D` laid out elements with it, and the mismatch between the ordering and everything derived from it corrupted the heap far from the cause (`HypreBoomerAMG` and friends crashed in `STable3D::Push`).

Also gates the affine spellings of division and remainder on a provably non-negative numerator: sdiv rounds toward zero and floordiv toward minus infinity, srem takes its numerator's sign and mod has none, and the unsigned forms read a negative number as enormous. Only an arithmetic shift right is a floor division whatever the sign. A numerator that cannot be shown non-negative stays untranslated and the access lowers as a plain memref access.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -6472,6 +6472,10 @@ bool valueCmp(Cmp cmp, Value bval, ValueOrInt val) {
   }
 
   if (cmp == Cmp::GE && !val.isValue && val.i_val == 0) {
+    // A zero extension leaves the sign bit clear whatever it widened.
+    if (bval.getDefiningOp<arith::ExtUIOp>() ||
+        bval.getDefiningOp<LLVM::ZExtOp>())
+      return true;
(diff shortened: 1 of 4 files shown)


wsmoses:
Reworked per review: dropped the bespoke `isKnownNonNegative` and reused `valueCmp(Cmp::GE, v, 0)` for the numerator gates. Also taught `valueCmp` that a zero extension (`arith.extui`/`llvm.zext`) is non-negative by construction, which it did not know. All 180 raising/mem2reg/affinecfg lit tests pass and the MFEM Hilbert-ordering probe still comes out byte-identical to plain clang.

PR #2823, opened 2026-08-09
Add enzymexla.math.fmuladd: keep fmuladd's fuse-if-profitable contract through the round trip

### LINKED ISSUE #2821: Round trip strengthens llvm.fmuladd into llvm.fma → libm fma() calls on non-FMA targets → ~3x slower plain codegen
## Root cause (proven): the round trip strengthens `llvm.fmuladd` into `llvm.fma`

Two passes compose into a silent semantic strengthening of every multiply-add:

1. **`libdevice-funcs-raise`**: `llvm.intr.fmuladd` → `math.fma` (so the MLIR AD table has a rule for it)
2. **`convert-polygeist-to-llvm`** (upstream `MathToLLVM::FmaOpLowering`): `math.fma` → **`llvm.intr.fma`**

`llvm.fmuladd` means *"fuse only if profitable, else emit mul+add"*. `llvm.fma` **requires** single-rounding fused semantics. On a target without FMA units — any plain `-O3` build without `-mfma`/`-march=native`, like this benchmark — the only correctly-rounded lowering of `llvm.fma` is a **libm `fma()` call per multiply-add**.

The MFEM diffusion kernel has 32 multiply-adds per quadrature point → 32 libm calls per iteration. That is directly the `__fma_fma3` at 30–55% of the runtime profile, it blocks SLP vectorization (can't pack around calls), and it forces caller-saved-xmm spills around every call (the 133 `%rsp` ops vs baseline's 2). The delinearized index math I originally blamed is a red herring — post-round-trip InstCombine folds it completely (`(q*72) sdiv 8` → `mul nsw 9`; final IR has zero `sdiv`).

**Decisive experiment** — the *same* round-tripped `kern` IR, only the target attr changed:

| `llc -O3` on round-tripped kern | libm fma calls | stack mem-ops | `vfmadd` |
|---|---|---|---|
| generic x86-64 | **32** | **133** | 0 |
| `-mattr=+fma` | 0 | 2 | 26 |

With FMA hardware the round-tripped code is fine — which is why this only bites default `-O3` builds.

## Single dependency-free MWE

One `.ll` file, two identical loops differing **only** in the intrinsic (verify with stock `opt`/`llc`, no other deps):

```llvm
target triple = "x86_64-unknown-linux-gnu"
define void @loop_fmuladd(ptr noalias %a, ptr noalias %b, ptr noalias %c, ptr noalias %d, i64 %n) {
entry:
  br label %loop
loop:
  %i = phi i64 [ 0, %entry ], [ %i.next, %loop ]
  %pa = getelementptr inbounds double, ptr %a, i64 %i
  %pb = getelementptr inbounds double, ptr %b, i64 %i
  %pc = getelementptr inbounds double, ptr %c, i64 %i
  %pd = getelementptr inbounds double, ptr %d, i64 %i
  %va = load double, ptr %pa, align 8
  %vb = load double, ptr %pb, align 8
  %vc = load double, ptr %pc, align 8
  %r = call double @llvm.fmuladd.f64(double %va, double %vb, double %vc)   ; <— ONLY difference
  store double %r, ptr %pd, align 8
  %i.next = add nuw nsw i64 %i, 1
  %done = icmp eq i64 %i.next, %n
  br i1 %done, label %exit, label %loop
exit:
  ret void
}
; @loop_fma: byte-identical except  %r = call double @llvm.fma.f64(...)
declare double @llvm.fmuladd.f64(double, double, double)
declare double @llvm.fma.f64(double, double, double)
```

`opt -passes='default<O3>' | llc -O3` (generic x86-64):

| | libm calls | packed SIMD | stack mem-ops |
|---|---|---|---|
| `@loop_fmuladd` | 0 | 4 | 0 |
| `@loop_fma` | **3** | **0** | **10** |

With `llc -mattr=+fma` both are clean (`vfmadd*`) — same signature as the full kernel.

And the round trip performs exactly this rewrite (3-line probes):
- `enzymexlamlir-opt --libdevice-funcs-raise` on `llvm.intr.fmuladd(...)` → `math.fma`
- `enzymexlamlir-opt --convert-polygeist-to-llvm=backend=cpu` on `math.fma` → `llvm.intr.fma`

## Fix

Lower `math.fma` back to **`llvm.intr.fmuladd`** in `convert-polygeist-to-llvm` (i.e., don't use the upstream `FmaOpLowering` for the CPU path), or record on raise that the op came from `fmuladd` and restore it on lowering. Round-tripping must not strengthen contraction semantics the source never asked for. (`math.fma`'s documented semantics likely do mean strict fma, so the raise side arguably needs a `math.fmuladd`-style distinction or an attribute — maintainer's call which end to fix.)

## Benchmark context (unchanged)
`bench_dfem_linearized_simple` order 3: plain-codegen path 2.3–3.5× slower than `clang -O3 + ClangEnzyme`; Enzyme-AD path competitive (1.0–1.4×); `linf_error=0` everywhere. Toolchain: Enzyme-JAX `main` @ed3b56d45, Enzyme @ab0bd5b8.

(Earlier revisions of this issue blamed un-promoted allocas, then the delinearized index math — both disproven by the standalone IR: there are no allocas at any stage, and the index math folds completely. The fmuladd→fma strengthening is the whole story.)

wsmoses:
Fixes #2821.

The round trip raised `llvm.intr.fmuladd` to `math.fma` (`libdevice-funcs-raise`) and lowered `math.fma` to `llvm.intr.fma` (upstream `MathToLLVM` via `convert-polygeist-to-llvm`) — a silent semantic **strengthening**. `fmuladd` only *permits* fusing (else an ordinary mul+add); `math.fma`/`llvm.fma` *require* the single rounding, which a target without FMA units honors with a **libm `fma()` call per multiply-add**. On MFEM's dFEM diffusion benchmark (plain `-O3`, no `-mfma`) that put 30–55% of runtime into `__fma_fma3`, blocked SLP vectorization around the 32 calls per quadrature point, and spilled caller-saved xmm registers around each — the ~3× plain-codegen slowdown of #2821.

The math dialect has no permission-form op (`math.fma` documents its lowering is *guaranteed* to produce `llvm.fma`), so this adds one:

- **`enzymexla.math.fmuladd`** (`FloatLike ×3 → FloatLike`, `Pure/SameOperandsAndResultType/Elementwise`), documented as the `fmuladd` contract carried through the round trip.
- **Raise**: `libdevice-funcs-raise` raises `llvm.intr.fmuladd` to it (+ declares the dialect dependency). `llvm.intr.fma` and libm `fma()` still raise to the strict `math.fma` — the previously-conflated distinction is restored.
- **Lower**: `convert-polygeist-to-llvm` lowers it back to `llvm.intr.fmuladd`; `math.fma → llvm.intr.fma` unchanged.
- **Derivative rule** (`EnzymeXLADerivatives.td`): mirrors `math.fma`'s — adjoints `(g·y, g·x, g)` as arith ops; forward from summed reverse. Verified fwd (`ẋy + xẏ + ż`, primal stays fmuladd) and rev.
- **StableHLO paths**: `affine-to-stablehlo` tensorizes it like the other ternary elementwise ops; `arith-raise` splits it into `stablehlo.multiply + add` (stablehlo has no fused form — required for strict fma, permitted for fmuladd alike).

**Results** on the reduced diffusion kernel (`inv`/`transpose`/`det` of a 3×3, plain loop), `objdump` of `kern`:

| | insns | libm fma calls | `mulpd` | stack mem-ops |
|---|---|---|---|---|
| stock `clang -O3` | 167 | 0 | 16 | 2 |
| round trip, before | 260 | **32** | 0 | **133** |
| round trip, **after** | **168** | **0** | **16** | **2** |

Plain codegen returns to parity with stock clang.

Tests: `diffrules/enzymexla/fmuladd.mlir` (fwd+rev), `lower_fmuladd.mlir` (permission → `llvm.intr.fmuladd`, strict `math.fma` → `llvm.intr.fma`), `raising/affine_to_stablehlo_fmuladd.mlir`, and the `libdevice_raise.mlir` expectation updated. Full lit suite 1137/1137. MFEM end-to-end (punit_tests + `bench_dfem_linearized_simple`) rebuild in progress; results to follow in a comment.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -1298,6 +1298,31 @@ def ReluOp: EnzymeXLA_Op<"math.relu", [Pure, SameOperandsAndResultType, Elementw
   }];
 }
 
+def FMulAddOp: EnzymeXLA_Op<"math.fmuladd", [Pure, SameOperandsAndResultType, Elementwise]> {
+  let summary = "Multiply-add that may, but need not, be fused";
+
+  let description = [{
+    Carries the semantics of the `llvm.fmuladd` intrinsic through the round
+    trip: compute `a * b + c`, fused into a single rounding only when the
+    target can do so profitably, otherwise as an ordinary multiply and add.
+    This is a *permission* to fuse, unlike `math.fma`, whose single rounding
+    is required and whose honest lowering on a target without FMA units is a
+    libm call. Raised from `llvm.intr.fmuladd`; lowered back to the same.
+  }];
+
+  let arguments = (ins
+    FloatLike:$a,
+    FloatLike:$b,
+    FloatLike:$c
+  );
+
+  let results = (outs FloatLike:$result);
+
+  let assemblyFormat = [{
+    $a `,` $b `,` $c attr-dict `:` type($result)
+  }];
+}
+
 def TGammaOp: EnzymeXLA_Op<"math.tgamma", [Pure, SameOperandsAndResultType, Elementwise]> {
   let summary = "Computes the true gamma function";
 
--- a/src/enzyme_ad/jax/Implementations/EnzymeXLADerivatives.td
+++ b/src/enzyme_ad/jax/Implementations/EnzymeXLADerivatives.td
@@ -229,6 +229,16 @@ def : EnzymeXLADerivative<"SincOp", (Op $x),
   (Sinc (Shadow $x))
 >;
 
+// Scalar op from the raised-LLVM path (FloatLike, not tensors), so its
+// adjoints are arith ops -- the same rule as math.fma, whose value it shares.
+def : EnzymeXLADerivative<"FMulAddOp", (Op $x, $y, $z),
+  [
+    (CheckedMulF (DiffeRet), $y),
+    (CheckedMulF (DiffeRet), $x),
+    (DiffeRet)
+  ]
+>;
+
 def CoscDerivTerm : SubRoutine<(Op $diffret, $x),
   (
     Select 
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3103,7 +3103,7 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
   }
 
   // ternary ops
-  if (isa<arith::SelectOp, math::FmaOp>(op)) {
+  if (isa<arith::SelectOp, math::FmaOp, enzymexla::FMulAddOp>(op)) {
     assert(op->getNumOperands() == 3 && op->getNumResults() == 1);
 
     Value a = mapping.lookup(op->getOperand(0)),
--- a/src/enzyme_ad/jax/Passes/ArithRaising.cpp
+++ b/src/enzyme_ad/jax/Passes/ArithRaising.cpp
@@ -19,6 +19,7 @@
 #include "mlir/IR/IRMapping.h"
 #include "mlir/Interfaces/ControlFlowInterfaces.h"
 #include "mlir/Interfaces/FunctionInterfaces.h"
+#include "src/enzyme_ad/jax/Dialect/Ops.h"
 #include "src/enzyme_ad/jax/Passes/Passes.h"
 #include "src/enzyme_ad/jax/Utils.h"
 #include "xla/mlir_hlo/mhlo/IR/hlo_ops.h"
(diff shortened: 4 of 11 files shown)


wsmoses:
MFEM end-to-end results with this PR (merged-main toolchain, `REACTANT_LOWER_INVOKE=0`):

**Correctness**: `punit_tests` = **All tests passed (88,493 assertions in 103 test cases)** — exact baseline parity, and the new derivative rule is exercised by the dFEM AD tests.

**`bench_dfem_linearized_simple`** (order 3, `-w10 -i100`, stable over 3 trials, `linf_error=0` everywhere), ours/vanilla-clang+ClangEnzyme:

| case | before this PR | after |
|---|---|---|
| quad forward global (plain codegen) | 2.29× | **0.59×** |
| quad deriv  global (plain codegen) | 2.66× | **0.80×** |
| hex forward global (plain codegen) | 3.46× | **0.70×** |
| hex deriv  global (plain codegen) | 2.83× | **0.73×** |
| quad/hex local (Enzyme AD, fwd+deriv) | 1.0–1.4× | **1.00×** (parity) |

The former 2.3–3.5× plain-codegen slowdown is gone; with the fmuladd contract preserved, the round-tripped kernels now run **25–40% faster than stock `clang -O3`** on the global path, and the Enzyme-AD path sits at exact parity.

Kernel-level: `objdump` of the reduced diffusion qfunc is at parity with stock clang (168 vs 167 insns — the +1 is loop-header bookkeeping from the scf-normalized IV; FP opcode mix identical: 16 `mulpd`, 27 `mulsd`, 0 libm calls, 2 stack ops).

Issue #2825, opened 2026-08-09
dFEM TUs take ~10min to compile: 58% is 18 single-threaded module-scope canonicalize runs over a 62MB module


wsmoses:
## Where the ~10 minutes go (measured on the worst TU)

`tests/unit/dfem/test_functional_gradient.cpp`, merged-main toolchain, plain `-O3`:

| stage | time |
|---|---|
| plain clang -O3 (no plugin), whole TU | 38.6s |
| **with plugin, whole TU** | **656s (10:56)** |
| — MLIR pipeline (offline replay, `--mlir-timing`) | **443s** |
| — remainder (import + translate-back + link + clang backend on merged module) | ~175s |

The TU imports as a **62MB module, 3,029 functions, 188 `__enzyme_*` call sites** (grows to 112MB after inlining).

**Pass ranking** (of 443s): `CanonicalizerPass` **257s (58%)** across 18 module-scope runs (~8.5s each); `LLVMToAffineAccess` 61s (14%); `AffineCFG` 17s; the Differentiate pass itself only 8.7s. Everything else ≤2%.

**Inside canonicalize the profile is flat** — no hot pattern: storage-uniquer/hashing ~5%, malloc/free ~4.5%, region-simplification liveness (`propagateLiveness`/`eraseUnreachableBlocks`/`SuccessorRange`) ~6%, the rest spread across the greedy driver. The cost is simply *module size × run count*, and — decisive — **User ≈ Wall (462s vs 443s): the entire pipeline is effectively single-threaded**, because every canonicalize (and the other heavy passes) runs at module scope.

## Fix directions, ranked by measured/estimated payoff

1. **Run the heavy passes nested per-function so the pass manager parallelizes them.** `builtin.module(llvm.func(canonicalize), func.func(canonicalize))` measured 5.78s wall / 8.19s user vs 8.22s/8.52s module-scope on the same dump (both including ~3.5s single-threaded parse; the canonicalize portion roughly halves, in-process would gain more). Speedup is currently capped by a few giant dFEM kernel functions — but combined with (2)/(3) and 192 cores this is the largest lever. Needs an audit of which of the 18 runs actually require module scope (canonicalize itself doesn't; interplay is with `symbol-dce`, which can stay module-level).
[...]

PR #2826, opened 2026-08-09
Incremental canonicalization: skip functions unchanged since their last run

wsmoses:
First fix from the compile-time investigation in #2825. **Draft** pending the remaining validation (below).

## Why the stock canonicalizer is slow here

The round-trip pipeline canonicalizes the module ~17 times, and the stock cost is **independent of how much changed**:
- the greedy driver re-seeds **every op, every run**: pattern-list lookup, failed `matchAndRewrite`s (affine ones speculatively build maps through the mutex-sharded uniquer just to say no), `fold()` attempts — measured ~1.7µs/op of pure failed-match work;
- its in-loop `simplifyRegions` rebuilds reachability, a **liveness fixpoint** (`DenseSet<Value>` per operand), and **identical-block hashing/comparison** over every block, from scratch, once per driver iteration — and any simplification re-triggers a full module sweep.

On MFEM's `test_functional_gradient` (62MB, 3,029 functions, 452K blocks — 19,108 `llvm.invoke`s in EH functions the raising never touches, per the #2800 design): **257s of the 443s pipeline**, the six post-affine runs at 25.6–28.3s *each*, almost all of it concluding nothing changed. Catch2's thousands of near-identical landingpad blocks are pathological for the block-merging hash clusters.

## The pass

`canonicalize-incremental` canonicalizes each top-level region-holding op separately (in parallel via `parallelForEach`, which also confines region simplification per-op) and remembers in a process-side table an `OperationFingerPrint` of the op when it was last left canonical; a matching fingerprint skips the op wholesale. The fingerprint compare **is** the correctness guard — a stale table entry (erased op, reused allocation) can only cause a skip when content is identical, which is exactly when skipping is right. No IR attributes, so nothing reaches translation. `raise.cpp` swaps the 21 module-scope `canonicalize` pipeline entries for it (the `inline` option string, func-scoped `postpasses` canonicalize, `canonicalize-loops`, `canonicalize-scf-for` are untouched).

## Measured (offline replay, same module, `--mlir-timing`)

| | stock | incremental |
|---|---|---|
| pipeline total | 443s | **205s** |
| canonicalize aggregate | 257s | **15.3s** |

## Caveat (in the pass doc)

A pattern whose match depends on state *outside* the op — another symbol changing while the caller's body didn't — can be missed where a module-scope sweep would catch it. That's a missed cleanup, never a miscompile; canonicalization is fixpoint-bounded best-effort and nothing downstream may rely on it firing.

## Validation status

- [x] lit suite 1134/1134 with the pass linked and smoke of the skip path
- [ ] output-equivalence diff vs stock pipeline on the big TU (baseline replay running)
- [ ] real plugin recompile timing of `test_functional_gradient` (expect ~656s → ~420s; the ~175s import/translate/backend tail is untouched, tracked in #2825)
- [ ] MFEM `punit_tests` 103/103


wsmoses:
Validation progress:
- **Real plugin recompile of `test_functional_gradient.cpp`: 10:56 → 5:20 (656s → 320s, 2.05× end-to-end)**, peak RSS 2.0GB (unchanged). The remaining ~115s beyond the frontend's 38s is the import/translate/backend tail plus the reduced pipeline — consistent with the offline replay (443s → 205s).
- Output equivalence: stock vs incremental pipeline outputs on the same TU are byte-identical in size; 730 of 1.6M lines differ, and normalizing `^bbN`/`%N` numbering shows every differing line present in both — pure block-renumbering artifacts of per-function vs module-scope region simplification, structurally identical IR.
- lit 1134/1134.
- MFEM full rebuild + `punit_tests` with this plugin: in progress, will report here.

wsmoses:
Validation complete — all checklist items green:

- **MFEM `punit_tests`: All tests passed (88,493 assertions in 103 test cases)** after a full plugin rebuild of the library + suite.
- Validation earned its keep: the first punit rebuild caught a real bug — the default `GreedyRewriteConfig` runs **aggressive** region simplification, whose identical-block merging adds successor operands for the values merged blocks differed in, and `llvm.invoke` cannot carry an `index`-typed successor operand (`mesh.cpp`, `AddTriangleFaceElement`). The canonicalizer defaults to **Normal** for exactly this reason; the pass now matches (`826b1f2e3`).
- Final numbers with the corrected config: pipeline **443s → 211.7s**, canonicalize aggregate **257s → 15.3s**, real end-to-end recompile of `test_functional_gradient.cpp` **10:56 → 5:20 (2.05×)**. `llvm-to-affine-access` (63s) is now the dominant pipeline cost — tracked separately in #2825.

Marking ready for review.

wsmoses:
Reworked per review: the process-side `Operation* -> OperationFingerPrint` table was unsafe — an op erased by a later pass leaves a dangling key for whatever is allocated at the same address to find, and the table grows without bound across a build's modules. The stamp now lives in an `enzymexla.canonical_fp` attribute on the op itself, so it dies with the op.

`OperationFingerPrint` exposes no byte accessor, so the stamp is a SHA1 in its spirit (pointers of everything uniqued + `hashProperties()` over the nested walk); the root's discardable dictionary is hashed entry-by-entry so the stamp attribute itself stays out of the stamp, and pointer instability only errs toward re-canonicalizing. The round trip strips stamps before LLVM translation, and skip/canonicalize counts are pass statistics.

Verified: 3000-func module through the pass 3×: run 1 canonicalizes 3000, runs 2–3 skip all 6000. Full lit suite 1134/1134.

wsmoses:
Pushed the anchor fix: a `FunctionOpInterface` anchor (the postpasses case) is stamped and canonicalized directly as a single unit; any other non-single-block-symbol-table anchor likewise. Module anchors now also run their regionless children (globals, declarations) through one batched greedy pass — previously those were never canonicalized at all. The stamp is now a 64-bit `hash_combine` instead of SHA1.

Attribution measurements for how much of the speedup is work reduction vs parallelism are running on the real dFEM TU (shipped / forced-no-skip / stock canonicalizer through the full pipeline) — will post the split here.

wsmoses:
Attribution results (full raising pipeline on the real dFEM TU, `test_functional_gradient`, 23 canonicalize runs, per-pass wall timing):

| pipeline config | canonicalize total | whole compile |
|---|---|---|
| stock `canonicalize` (serial, module-scope) | **329.7s** | 10:03 |
| incremental, skip **disabled** (parallel per-func + confined region-simplify) | **47.7s** | 5:19 |
| incremental, skip enabled (shipped) | **49.5s** | 5:27 |

So the honest split: **effectively all of the 6.9× canonicalize win is parallelism + per-op confinement; the fingerprint skip contributes ~nothing** (it's ~2s *slower*, within noise — the stamp walk costs about what the failed-match sweep it avoids costs, and the expensive giant kernel functions are exactly the ones the intermediate passes keep touching, so they never skip; the 3277 functions that do skip are the small EH ones whose parallel cost was already negligible). Steady-state microbenchmarks on the imported module agree: second-run stock 0.90s vs incremental 0.39s with skip and 0.39s without, all 3277 functions confirmed skipping.

Given that, I'd recommend simplifying the PR to drop the fingerprint/stamp machinery entirely and keep what actually pays: parallel per-function canonicalization with region simplification confined per unit (plus the regionless-children batch). Happy to push that simplification if you agree.

wsmoses:
Done — the pass is now `canonicalize-parallel`, with the fingerprint/stamp machinery removed entirely per the attribution data: each top-level region-holding op gets its own greedy driver in parallel (which also confines per-iteration region simplification to that op), regionless children are one batched run, and a `FunctionOpInterface`/non-symbol-table anchor is canonicalized directly. No attributes, nothing to strip before translation, and the correctness surface the stamps carried (dangling pointers, stamp-in-fingerprint, cross-process identity) is gone with them.

Steady-state on the real module is identical to the stamped version (0.41s vs 0.39s); full lit 1134/1134. Running the MFEM punit gate against this final shape next.

wsmoses:
MFEM punit gate on the final `canonicalize-parallel` shape: full test rebuild through the raising path (`REACTANT_LOWER_INVOKE=0`) — **All tests passed (88,493 assertions in 103 test cases)**, byte-for-byte the campaign baseline. The PR is ready from my side.

PR #2827, opened 2026-08-09
llvm-to-affine-access: index access builders by base pointer

wsmoses:
Second fix from the compile-time investigation in #2825 (companion to #2826, independent of it).

A converted access can be the base pointer of other accesses, whose builders need the replacement said to them. The conversion loop said it to **every** builder for **every** conversion — and `maybeReplaceBase` is a single comparison, so this was a full quadratic scan over the access set. `perf` on MFEM's `test_functional_gradient` put **69% of the pass's self time** in this loop: 48s of the 63s pass, the largest pipeline cost left after #2826.

The fix indexes builders by base pointer once and tells only the bucket that cares, moving it under the new base. Same replacements by construction — `maybeReplaceBase` was a no-op for every builder whose base didn't match.

Measured on the same dump: first-run pass time **48s → 6.9s (7×)**. lit suite 1134/1134 (run with this change). Combined with #2826, the full-pipeline replay number will be posted on #2825.

### DIFF
--- a/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
@@ -1911,6 +1911,25 @@ convertLLVMToAffineAccess(Operation *op,
   }
 
   IRMapping mapping;
+  // A converted access can be the base pointer of other accesses, whose
+  // builders then need the replacement said to them. Saying it to every
+  // builder for every conversion is quadratic in the number of accesses;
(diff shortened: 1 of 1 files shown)


wsmoses:
Validation: full MFEM rebuild (library + suite) with this change combined with #2826 — **`punit_tests`: All tests passed (88,493 assertions in 103 test cases)**, lit 1134/1134. Combined offline pipeline replay on `test_functional_gradient`: **443s → 171.4s (2.6×)**; this pass alone 61s → 21s across both runs (first run 48s → 6.9s).

PR #2828, opened 2026-08-09
Verify the round-trip module once, not after every pass

wsmoses:
Third fix from #2825 (independent of #2826/#2827).

perf on a real plugin compile of MFEM's `test_functional_gradient` showed `mlir::verify → verifySymbolTable` at **8% of the entire clang invocation**: the pass manager re-verifies the whole module (symbol-table walk + dominance over every op) after each of the pipeline's ~65 passes — mostly re-proving the unchanged exception-handling functions well-formed after passes that never looked at them.

Offline replay on the same TU (with #2826+#2827 in): **171.4s → 114.1s** with inter-pass verification off — a third of the remaining pipeline. Cumulative across the three PRs: **443s → 114s (3.9×)**.

The change verifies **once after the pipeline** — malformed IR still fails with proper diagnostics rather than inside the LLVM translator — and `DEBUG_REACTANT` keeps per-pass verification, which is exactly when you want to bisect a miscompile to a pass. lit is unaffected (`enzymexlamlir-opt` keeps mlir-opt's own `--verify-each` semantics; this touches only the plugin's PassManager in `raise.cpp`). MFEM `punit_tests` validation running; result to follow.

### DIFF
--- a/src/enzyme_ad/jax/raise.cpp
+++ b/src/enzyme_ad/jax/raise.cpp
@@ -244,9 +244,9 @@ extern "C" std::string runLLVMToMLIRRoundTrip(std::string input,
   // dominance check over every operation -- after every pass. Over this
   // pipeline's ~65 passes on a large TU that is a third of the pipeline's
   // wall time, re-proving unchanged exception-handling functions well-formed.
-  // Verify once at the end instead (below); DEBUG_REACTANT keeps the
-  // per-pass verification for debugging miscompiles to a pass.
-  if (!getenv("DEBUG_REACTANT"))
+  // Verify once at the end instead (below); options->verifyEach restores the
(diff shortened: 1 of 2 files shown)


wsmoses:
Validation: full MFEM rebuild + **`punit_tests`: All tests passed (88,493 assertions in 103 test cases)** with this change (verifier-only, on stock main). The end-of-pipeline verification correctly remains the failure gate — malformed IR still errors with diagnostics before translation.

wsmoses:
Revised per review: the control is now `MLIRRoundTripOptions::verifyEach` (default **false**) instead of the environment variable — the plugin's designated-initializer construction in Reactant gets the false default with no change there; a Reactant-side flag can re-enable per-pass verification for pass bisection when wanted. Final-verification gate before translation unchanged.

PR #2829, opened 2026-08-10
CUDA host round trip: lower affine_atomic_rmw, drop unresolved stub modules

wsmoses:
Fixes the compiler-explorer crash on a six-line `__enzyme_autodiff` CUDA kernel (segfault in ReactantNewPM), reproduced and debugged locally. Three independent defects in the host-side round trip:

1. **`enzyme.affine_atomic_rmw` was never lowered.** The adjoint of a device store is an atomic accumulate; the op survived to `convert-polygeist-to-llvm`, whose source materialization built a `MemRefDescriptor` from a null converted type and segfaulted (the CE stack). Enzyme's `lower-affine-atomic-rmw` now runs alongside the rest of the affine lowering. (LBM never hit this because it runs with `remove-atomics`.)
2. **Unresolved-stub `gpu.module`s reached registration and serialization.** A kernel the raising can't resolve leaves a module wrapping an external `reactant$` declaration — no device code, and the original clang fatbin still carries the kernel. It was serialized anyway: ptxas got handed the host's target-cpu (`Value 'x86-64' is not defined for option 'gpu-name'`), or a `"binaryAttr"` placeholder reached the runtime (`device kernel image is invalid`). Such modules are now erased before conversion.
3. **`getGPUTarget` hardened** to never admit a non-`sm_` chip into the NVVM target attribute.

After the fixes the kernel compiles and runs end-to-end on an RTX 5090 (sm_120, CUDA 12.9) and matches LLVM-Enzyme's output exactly (`y=1.96, d_x=2.8` for `x=1.4`).

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -4816,27 +4816,6 @@ struct ConvertPolygeistToLLVMPass
     op->getResult(0).getType(); });
         */
 
-    // A kernel the raising could not resolve leaves behind a gpu.module
-    // whose body is a wrapper around an external `reactant$` declaration:
-    // there is no device code in it to register or serialize -- the
-    // original clang fatbin still carries that kernel -- and its target
-    // attribute is whatever the host stub wore. Drop such modules whole
-    // before they grow a bogus registration or reach gpu-module-to-binary.
-    {
-      SmallVector<gpu::GPUModuleOp> unresolved;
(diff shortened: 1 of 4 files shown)


wsmoses:
Both review points addressed: the pipeline comment is gone, and the `ConvertPolygeistToLLVM.cpp` erasure is **dropped** — an A/B with just that hunk reverted shows the MWE still compiles and runs correctly (`d_x=2.8`, clean launch). It was never load-bearing: the stub modules are created during the conversion itself (after the pre-conversion erasure ran), and with `getGPUTarget` refusing non-`sm_` chips their serialization is benign. The PR is now just the two real fixes: `lower-affine-atomic-rmw` in the pipeline and the target hardening.

PR #2832, opened 2026-08-10
parallel-lower: repair yields whose inlined callee cannot return

wsmoses:
Follow-up to #2829 (these commits landed on that branch after it merged). Root-causes and fixes the remaining MFEM CUDA compile crash, per-review without any LLVM patch:

**Bug**: `parallel-lower`'s GPU-inline wrapping puts a call in an `scf.execute_region` that yields the call's results — but when the inlined callee never returns (an assert/error path that prints and traps, as MFEM's device bounds checks do), the yield is left with nothing to forward: *"region branch point has 0 operands, but region successor needs 1 inputs"*. With pipeline verification off (#2828 default) the malformed op survived until `scf.execute_region`'s inlining canonicalization walked it into upstream's unimplemented `defaultReplBuilderFn` and segfaulted.

**Fix**: control cannot reach such a yield, so its operands are said as poison (`repairNonReturningYields`, applied at all three wrap sites).

**Sixteen-line reproducer** (crashes before, compiles+runs after):
```cuda
#include <cstdio>
__device__ void check(int i, int n) {
  if (i >= n) { printf("out of bounds %d\n", i); __builtin_trap(); }
}
__global__ void fill(int *a, int n) { int i = threadIdx.x; check(i, n); a[i] = i; }
int main() { int *a; cudaMalloc(&a, 128); fill<<<1,32>>>(a, 32); cudaDeviceSynchronize(); }
```

Validated: the reproducer runs, the `__enzyme_autodiff` CUDA MWE still matches LLVM-Enzyme exactly, and MFEM's `sparsemat.cpp` (the TU that surfaced this) compiles. Also adds `REACTANT_DUMP_IMPORTED` for dumping the pipeline's exact input module, which is how this was localized.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ParallelLower.cpp
+++ b/src/enzyme_ad/jax/Passes/ParallelLower.cpp
@@ -23,6 +23,7 @@
 #include "mlir/Dialect/LLVMIR/ROCDLDialect.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/IR/Dominance.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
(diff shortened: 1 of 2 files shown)


wsmoses:
Dropped the `raise.cpp` diff — the PR is now only the ParallelLower fix. On the test: I have not yet managed to carve a lit-sized MLIR reproducer. Automated reduction of the real 30k-line pre-parallel-lower module (greedy top-level deletion with 'repair fires' as the criterion, then per-launcher symbol-closure probes) bottoms out at ~200 interdependent top-level ops — the trigger needs a specific constellation of the launch chain, the reactant$ stub renaming, and the registration machinery that hand-written modules so far don't recreate (a directly-written guarded `vprintf`+trap in a `gpu.launch`, with or without a callee wrapper, wraps through a different path and stays well-formed). The end-to-end reproducer is solid — the 16-line CUDA file in the description crashes before this change and runs after — and I'll keep chiseling at an in-tree MLIR test as a follow-up.

wsmoses:
Split this PR per review: this branch now carries only the parallel-lower noreturn-yield repair, with a lit test (the hand-written reproducer needed the callee to return a value and hold a device intrinsic — that's why earlier reduction attempts bottomed out). The other fixes moved to their own PRs, each with tests:
- #2834 affine-cfg: never insert index casts after a terminator
- #2835 convert-parallel-to-gpu: survive kernels without an exact or recorded shape
- #2836 never memref.cast across memory spaces

The comdat handling moved to EnzymeAD/Enzyme#3132 (the derivative clone drops the attribute at creation).

PR #2836, opened 2026-08-10
Never memref.cast across memory spaces

wsmoses:
Split out of #2832 per review.

Two sites patched users of a rebuilt allocation with a `memref.cast` into another memory space, which the op cannot express — only disabled inter-pass verification let it through, and MFEM's dfem CUDA tests crashed downstream on it.

- convert-parallel-to-gpu's shared-memory promotion launders the space-5 alloca through `memref2pointer`/`pointer2memref` (their simplification refuses to fold across spaces; `memref.memory_space_cast` doesn't work here because upstream canonicalization bubbles it into consumers and recreates the invalid cast; the lowering emits the addrspacecast).
- llvm-to-affine-access already went through the pointer for element-type changes; a memory-space difference now takes the same path.

Lit tests for both sites.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1740,9 +1740,12 @@ struct ParallelToGPULaunch : public OpRewritePattern<enzymexla::GPUWrapperOp> {
             memref::AllocaOp::create(rewriter, alloca.getLoc(), type);
         if (auto align = alloca.getAlignment())
           newAlloca.setAlignment(*align);
-        auto cast = memref::CastOp::create(rewriter, alloca.getLoc(),
-                                           alloca.getType(), newAlloca);
-        it->replaceAllUsesWith(cast);
+        // memref.cast cannot change the memory space; that is what
+        // memref.memory_space_cast is for, and with C-style memrefs it
+        // lowers to an addrspacecast.
+        auto cast = memref::MemorySpaceCastOp::create(
+            rewriter, alloca.getLoc(), alloca.getType(), newAlloca);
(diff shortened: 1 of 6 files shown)


wsmoses:
Switched to `memref::MemorySpaceCastOp` as suggested — my earlier note about canonicalization recreating the invalid cast was wrong: the corruption came from a bad `replaceAllUsesWith` overload use in the first attempt, not from upstream folding (the bubble-down pass isn't in our pipeline and its passthrough impl retypes correctly). Both sites now emit `memory_space_cast`; the element-type-change case in llvm-to-affine-access still goes through the pointer as before. Tests updated.

wsmoses:
Went back to the pointer laundering after testing `memory_space_cast` end-to-end: the op itself works and my earlier folding claim was wrong in the details, but canonicalization (legitimately) folds the space cast into its consumers — loads, stores and casts absorb it and retype to space-annotated memrefs, 2000+ of them on the failing TU — and convert-polygeist-to-llvm's type converter returns null for those, which `MemRefDescriptor::fromStaticShape` then crashes on. Making the polygeist lowering handle numbered memory spaces is the real fix but a separate, larger change; the commit message documents this. The pointer ops don't get folded through (their simplification refuses cross-space folds) and lower to the addrspacecast.

wsmoses:
Done — kept `memref.memory_space_cast` at both sites and added the missing piece in convert-polygeist-to-llvm instead: with C-style memrefs a memref is a bare pointer in its memory space, so the cast lowers to `llvm.addrspacecast` (`CMemorySpaceCastOpLowering`). The earlier crash was unconverted `memory_space_cast` consumers making the upstream *descriptor-based* source materialization run on bare-pointer conversions. Full failing-TU pipeline now passes with the space casts flowing through; lit test for the lowering included.

PR #2840, opened 2026-08-10
convert-parallel-to-gpu: stop erasing values the parallel still reads

wsmoses:
Four correctness fixes in the kernel-normalization patterns, found raising MFEM's lor_batched.cpp as CUDA (which crashed the greedy driver on freed IR).

1. **Dangling erase** — `parallelize-block-ops` replaced uses only *inside* the parallel body, then erased the originals; the parallel's own bounds, the terminator, and users nested in other blocks kept pointing at freed ops. Ops with an unreachable reader now stay put (always sound), and the wrapper-body callers of `getDirectlyNestedSingleParallel` pass the `allowIndexComputation` flag it has always had so the leftover bound computation is expected.
2. **`isAncestor` → `isProperAncestor`** — `isAncestor` counts the op itself, so values read by the parallel's *own bounds* were misclassified as "used inside the body".
3. **Two mutate-before-bail convergence bugs** — `parallel-to-gpu-launch` created a constant before the checks that decline an unconvertible wrapper, and `parallelize-block-ops` built its trailing guard `if` before knowing there was trailing work. Each marks the IR changed on every scan, so the greedy driver never converges and the pass fails with **no diagnostic at all** (silent `rc=1`).
4. **`handle-wrapper-root-ops` loop** — it treated the parallel's bound computation as root ops to handle, rebuilding the identical wrapper shape forever.

Plus a better `Unhandled unrealized conversion cast` diagnostic: it now prints the cast, its enclosing function, and its users. That error is how a `gpu_wrapper` no pattern converted announces itself at lowering, and the bare message gave nothing to work from.

Verified: MFEM's mesh.cpp, tmop 302, dfem test_mass, the multikernel and hybridization repros all still lower cleanly; lit test for the pin case included.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -471,7 +471,9 @@ struct SplitParallelOp : public OpRewritePattern<enzymexla::GPUWrapperOp> {
 
   LogicalResult matchAndRewrite(enzymexla::GPUWrapperOp wrapper,
                                 PatternRewriter &rewriter) const override {
-    scf::ParallelOp pop = getDirectlyNestedSingleParallel(wrapper.getBody());
+    scf::ParallelOp pop = getDirectlyNestedSingleParallel(
+        wrapper.getBody(), /*allowAllocas*/ false,
+        /*allowIndexComputation*/ true);
     if (!pop)
       return failure();
     bool child = false;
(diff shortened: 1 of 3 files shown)


wsmoses:
Split per review into #2841 (diagnostic) and #2843 (the erase fix + the launch convergence fix it needs to be testable). Three further changes from this PR are dropped: writing a standalone test for each showed no demonstrable effect in isolation.

wsmoses:
Correction: the split PRs are #2841 (diagnostic) and #2842 (erase fix + launch convergence fix).

PR #2842, opened 2026-08-10
convert-parallel-to-gpu: stop erasing values the parallel still reads

wsmoses:
Split out of #2840 per review (2 of 2).

`parallelize-block-ops` replaced uses only *inside* the parallel body, then erased the originals — so the parallel's own bounds, the block terminator, and users nested in other blocks kept pointing at freed ops. Raising MFEM's lor_batched.cpp as CUDA crashed the greedy driver folding one such corpse (an `arith.cmpi` whose real reader was an `scf.condition` elsewhere). Ops with an unreachable reader now stay put.

Two subtleties: a use by the parallel op *itself* is one of its bounds, which the body's copy does not replace, so the walk asks `isProperAncestor` not `isAncestor`; and an all-pinned block reports no change rather than claiming one.

**Why the launch fix is in the same PR:** `parallel-to-gpu-launch` created its constant before the checks that decline an unconvertible wrapper — exactly the wrappers this fix now correctly produces. An op created before a bail marks the IR changed on every scan, so the driver never converges and the pass fails with no diagnostic at all. Without it the included lit test hangs instead of passing; the two are not separable.

I dropped three further changes from #2840 for lack of a demonstrable test (an empty-tail guard, a handle-wrapper-root-ops skip, and `allowIndexComputation` at the wrapper-body callers) — every small case I built for them behaves identically with and without.

Verified against MFEM's mesh.cpp, tmop 302, and the multikernel repro. test_mass fails identically on pristine main (a pre-existing `gpu.launch` async-token error), so it is unaffected by this change.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1037,6 +1037,43 @@ struct ParallelizeBlockOps : public OpRewritePattern<scf::ParallelOp> {
       }
     };
 
+    // Two things read values of this block from outside the inner parallel's
+    // body, where moving their computation inside cannot rewrite them: the
+    // parallel's own bounds, and the block's terminator -- which, when the
+    // parallel sits in a loop body rather than directly in the grid parallel,
+    // is an scf.condition or scf.yield reading values computed beside it.
+    // Only the uses within the body are replaced, so erasing what those read
+    // leaves them pointing at a corpse. They, and everything they are in turn
+    // computed from in this block, have to stay where they are.
+    DenseSet<Operation *> pinned;
+    {
+      SmallVector<Value> todo(pop->getOperands().begin(),
+                              pop->getOperands().end());
+      llvm::append_range(todo, outerBlock->getTerminator()->getOperands());
+      while (!todo.empty()) {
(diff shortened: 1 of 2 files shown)


wsmoses:
Rewrote the pinning as you suggested: a worklist seeded from what is read outside the body, marking defining ops and pushing their operands, skipping values from enclosing regions, asserting the ops are region-free. Half the code, and the `isProperAncestor` subtlety disappears with the user-walking.

One thing the rewrite exposed that my original version was papering over: seeded from `pop->getOperands()` alone, lor_batched crashed again (139). `outerBlock` is `pop->getBlock()`, which is *not* always the grid parallel's body — when the inner parallel sits in a loop nested in the grid parallel, it is that loop's body and its terminator is an `scf.condition`. The op that crashed lor_batched is an `arith.cmpi` that condition reads, which is not reachable from the bounds. So the seed is both:

```cpp
SmallVector<Value> todo(pop->getOperands().begin(), pop->getOperands().end());
llvm::append_range(todo, outerBlock->getTerminator()->getOperands());
```

With the terminator seeded, lorb goes 139 → 1 (the clean unconvertible-wrapper path). Verified: pin-bounds, shared-memory-alloca and blocksize-fallback lit tests pass; mesh.cpp, tmop 302 and the multikernel repro all still lower cleanly.

PR #2845, opened 2026-08-11
convert-parallel-to-gpu: interchange a guard with the parallel it guards

wsmoses:
Stacked on #2844.

Splitting a kernel whose thread count does not divide its work leaves the block parallel *inside* the guard its bounds need, and the launch wants it directly in the grid parallel's body. So the wrapper is never converted: it survives every kernel pattern, has its parallels serialized away later, and reaches the LLVM lowering as an unhandled unrealized conversion cast — the failure raising MFEM's lor_batched.cpp as CUDA.

Interchanging the two is sound because the condition is computed outside the block parallel and is therefore uniform across threads: running the parallel and having each thread skip the body is exactly what the guard means. A pure prologue inside the guard comes along (it can run unguarded); anything else leaves the guard alone.

**This is half of lor_batched:** it takes that TU's surviving wrappers from 32 to 16 (16 launches created). The remainder hold their thread parallel inside a *grid-stride `scf.for`* within the guard, which needs the loop and the parallel interchanged — a real loop transformation with its own legality conditions, and a design question I would rather raise than answer unilaterally. No regressions: test_mass, tmop 302, mesh.cpp and the multikernel repro all still lower cleanly.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1602,6 +1602,72 @@ struct RemovePolygeistGPUWrapperOp : public OpRewritePattern<OpType> {
   }
 };
 
+/// parallel {
+///   ...
+///   if (c) { parallel { A() } }
+/// }
+/// ->
+/// parallel {
+///   ...
+///   parallel { if (c) { A() } }
+/// }
+///
+/// The launch wants the block parallel directly in the grid parallel's body,
+/// but the guard its bounds need is left wrapped around it. The condition is
+/// computed outside the block parallel, so it is the same for every thread:
+/// running the parallel and having each thread skip the body is what the
+/// guard already means.
+struct InterchangeGuardedParallel : public OpRewritePattern<scf::IfOp> {
+  using OpRewritePattern<scf::IfOp>::OpRewritePattern;
+  const char *PATTERN = "interchange-guarded-parallel";
+  LogicalResult matchAndRewrite(scf::IfOp ifOp,
+                                PatternRewriter &rewriter) const override {
+    if (!ifOp.getElseRegion().empty() || ifOp->getNumResults() != 0)
+      return failure();
+    auto parent = dyn_cast<scf::ParallelOp>(ifOp->getParentOp());
+    if (!parent || !parent->getParentOfType<enzymexla::GPUWrapperOp>())
+      return failure();
+    auto *thenBlock = ifOp.thenBlock();
+    auto pop =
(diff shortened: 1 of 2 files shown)


wsmoses:
Correction to the description above — you're right that the nesting is from the original kernel, not from our splitter. MFEM's `forall.hpp`:

```cuda
template <typename BODY> __global__ static
void CuKernel2D(const int N, BODY body) {
   const int k = blockIdx.x*blockDim.z + threadIdx.z;
   if (k >= N) { return; }   // <- this is the guard
   body(k);                  // <- the thread-parallel work
}

template <typename BODY> __global__ static
void CuKernel3D(const int N, BODY body) {
   for (int k = blockIdx.x; k < N; k += gridDim.x) { body(k); }
}
```

So the guard is the raised form of the source's early return on the grid index (the trace's condition is `%grid_iv < %N`, not a thread-count remainder), and `body(k)` is where the discovered thread parallelism sits. The splitter's own tail guard (createSplitOp, line ~952) is emitted *inside* the block parallel and was never the problem.

That also sharpens what is left: the remaining wrappers are `CuKernel3D`, whose grid-stride loop is likewise straight from the source. Undoing it is not general loop interchange but recognising a strip-mined grid dimension — a loop over `blockIdx` stepping by `gridDim` — and mapping `k` back onto the grid, which is legal by the same construction that makes the original kernel correct. I'll fix the description's attribution.

wsmoses:
Closing this — you're right that interchanging is the wrong move, and checking it properly also shows it earns nothing.

On the structure: the parallel inside the guard is a *discovered* parallel from a loop within the kernel body (MFEM_FOREACH_THREAD's `for (i = threadIdx.x; i < N; i += blockDim.x)`), not the kernel's thread parallel. Since loops inside a kernel get serialized in the end anyway, conditioning the whole loop once is better than making every thread re-test the guard — moving the guard inward is strictly worse code for no benefit.

And the measurement I quoted was confounded by my own build state: the '32 → 16 surviving wrappers' was measured before #2844 landed. With #2844 alone, lor_batched is already at 16; with #2844 plus this interchange it is also 16. The interchange contributes nothing once the async fix is in, so there is no case for it even setting the design point aside.

The real remaining shape, dumped from main with #2844:

```
scf.parallel (%grid_x, %grid_y) {
  scf.if (k < N) {
    scf.for %k = blockIdx.x to N step gridDim.x {     // CuKernel3D grid-stride
      scf.parallel (%i) = 0 to 288 {                  // discovered from MFEM_FOREACH_THREAD
        scf.if ... { scf.parallel (%j) = 0 to 144 { ... } }
```

so the fix is to serialize those discovered inner parallels rather than chase them into the launch geometry, and let the block dims come from the kernel's own launch shape.

PR #2846, opened 2026-08-11
convert-parallel-to-gpu: launch what the kernel gives, serialize the rest

wsmoses:
The launch decider required a parallel directly under the grid parallel for its thread shape and declined the kernel otherwise. Declining is not neutral: a wrapper no pattern converts keeps its wrapper, has its parallels serialized away by later passes, and reaches the LLVM lowering — which has no pattern for a `gpu_wrapper` at all. That is the `Unhandled unrealized conversion cast` raising MFEM's lor_batched.cpp as CUDA.

Per review, the ordering should be: decide and create the launch parameters first, free to take bounds from the outer parallel axes, and serialize what is left afterwards. So the decider now takes grid axes from the outer parallel, thread axes from a parallel directly beneath it when there is one, and one thread per block when there is not. Nothing is moved to manufacture a block parallel — in particular the guard still wraps the whole loop, which is cheaper than re-testing it per thread and is what the kernel already said.

MFEM's `CuKernel3D` is the shape that needs it:

```cuda
for (int k = blockIdx.x; k < N; k += gridDim.x) { body(k); }   // forall.hpp
```

the parallelism the raising finds (from `MFEM_FOREACH_THREAD`) sits under that loop, not under the grid parallel.

**Result:** lor_batched.cpp's surviving wrappers go 16 → 0 and the file's full pipeline passes. test_mass, tmop 302, mesh.cpp and the multikernel repro are unchanged. Two lit tests: the CuKernel3D shape, and the existing pin-bounds test updated — it used to assert the wrapper was left alone, which is precisely the behaviour this replaces.

**Known follow-up:** these kernels launch one thread per block for now. Choosing their thread axes from the parallels inside (288 and 144 wide in lor_batched) is the next step; one thread is slow, where failing to compile was not an option at all.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1712,10 +1712,14 @@ struct ParallelToGPULaunch : public OpRewritePattern<enzymexla::GPUWrapperOp> {
         getDirectlyNestedSingleParallel(wrapper.getBody());
     if (!gridPop)
       return failure();
+    // A kernel whose thread-level work is not a parallel directly under the
+    // grid parallel -- MFEM's CuKernel3D walks its grid-stride loop and does
+    // its work in loops within it -- still has to become a launch. Decide the
+    // launch's shape here and let what is left inside be serialized: leaving
+    // the wrapper unconverted only means it reaches the LLVM lowering, which
+    // has no pattern for it.
     scf::ParallelOp blockPop = getDirectlyNestedSingleParallel(
         gridPop.getBody(), /* allowAllocas */ true);
-    if (!blockPop)
-      return failure();
 
     // Only mutate once the match is certain: an op created before a bail
     // marks the IR changed on every scan, so a wrapper this pattern cannot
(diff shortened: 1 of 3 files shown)


wsmoses:
Closing — you're right, this treats the symptom, and the data says the fallback is actively harmful.

`SplitParallelOp` declines any kernel whose parallel contains a nested parallel:

```cpp
bool child = false;
pop->walk([&](scf::ParallelOp p) { if (pop != p) child = true; });
if (child) { LLVM_DEBUG(DBGS() << "only single parallel ops\n"); return failure(); }
```

which is exactly these 32 kernels: the raising proved their `MFEM_FOREACH_THREAD` loops parallel, so the kernel parallel is never split into grid×block and the launch decider finds no block parallel to use. The kernels are penalised for the raising having succeeded.

Comparing `gpu_wrapper` operands (grid xyz, block xyz) between the groups:

| group | dominant shape | count |
|---|---|---|
| no nested parallel (converts today) | `(dyn, dyn, 1, dyn, dyn, dyn)` | 107 |
| nested parallel (survivors) | `(dyn, dyn, 1, dyn, dyn, dyn)` | 28 |
| nested parallel | `(dyn, dyn, 1, 1, 1, 1)` | 4 |

28 of the 32 carry dynamic block dimensions, so one-thread-per-block throws away a thread shape the wrapper already records — worse than I claimed in the description.

Redoing it per your ordering: serialize the discovered inner parallels first so the kernel parallel is single, let SplitParallelOp split it into real grid and block dims, and the launch then has its true thread axes with the guard untouched.

PR #2847, opened 2026-08-11
convert-parallel-to-gpu: serialize the parallels found inside a kernel

wsmoses:
Replaces #2846, which treated the symptom — thank you for pushing on where the nested parallels were supposed to come from.

`SplitParallelOp` refuses to split a kernel whose parallel still contains another parallel:

```cpp
bool child = false;
pop->walk([&](scf::ParallelOp p) { if (pop != p) child = true; });
if (child) { LLVM_DEBUG(DBGS() << "only single parallel ops\n"); return failure(); }
```

which is precisely the kernel the raising did best on: `MFEM_FOREACH_THREAD` is `for (i = threadIdx.x; i < N; i += blockDim.x)`, so proving it parallel leaves a parallel inside the kernel's own. The kernel then never gets its grid/block split, the launch is never built, and the wrapper reaches the LLVM lowering — the `Unhandled unrealized conversion cast` from lor_batched.cpp. Proving the inner loop parallel cost the kernel its launch.

Serialize those parallels before the split (where a kernel-internal loop ends up regardless) and the split takes grid and block dims from the launch geometry the wrapper records. The safety line: a split's own block parallel sits *directly* in its grid parallel's body, while the raising's sit deeper under the kernel's guards and loops — only the deeper ones are serialized.

**Result:** lor_batched.cpp's 32 unconverted kernels all become launches, and with the thread shapes they were launched with:

| threads | count |
|---|---|
| dynamic (`%91 × %92 × %10`) | 15 |
| 8×8×4 | 2 |
| 7×7×4 | 2 |
| 6×6×6 | 2 |
| 5×5×5 | 2 |

versus 1×1×1 for all of them under #2846. lor_batched.cpp's full pipeline passes; test_mass, tmop 302, mesh.cpp and the multikernel repro are unchanged; lit test included.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1602,6 +1602,48 @@ struct RemovePolygeistGPUWrapperOp : public OpRewritePattern<OpType> {
   }
 };
 
+/// A parallel the raising found inside the kernel's own parallel -- what
+/// MFEM_FOREACH_THREAD's 'for (i = threadIdx.x; i < N; i += blockDim.x)'
+/// becomes when it is proven parallel -- is not where the launch gets its
+/// thread shape from: that comes from the launch geometry the wrapper
+/// records. Worse, split-parallel refuses to split a kernel that still
+/// holds one, so proving the inner loop parallel costs the kernel its grid
+/// and block dimensions entirely. Serialize them, which is where a loop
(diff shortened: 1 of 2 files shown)


wsmoses:
Superseded by #2848 — your version is right. Serializing in this pass put the serialization *before* the launch was decided, which is the inversion you flagged earlier; narrowing the predicate to direct children leaves the loops alone for parallel-serialization to handle afterwards, and gets the identical result (same 32 launches, same thread shapes) from 9 lines.

PR #2857, opened 2026-08-11
Bind kernels to clang's device stub instead of a synthetic one

wsmoses:
## Problem

Kernel registration is keyed by **host pointer**, and every kernel was registered under a synthetic stub the pass mints itself:

```cpp
stub = LLVM::LLVMFuncOp::create(
    rewriter, ctorloc, getFuncStubName(moduleName, f.getName()), ...);
```

No user code holds that address. It resolved only because the symbolic lowering paths (launch, occupancy, `ConvertGPUKernelAddressOp`) rewrote kernel references to the same synthetic symbol, so both sides agreed.

Those paths need the kernel as a `SymbolRefAttr`. An address that reaches the runtime *opaquely* never becomes one: it stays an `!llvm.ptr` holding clang's original device stub, wrapped in `__reactant$get_device_from_host`, which is resolved to the identity. That stub was never passed to `__cudaRegisterFunction`, so the query returned 400 (invalid resource handle).

Optimized builds hid this, since inlining drags the address back to the call site where the plugin's static rewrite fires. Reduced reproducer -- three queries of the *same* kernel at `-O0`, all passing under vanilla clang:

```
direct     err=0   ptx=120     // cudaFuncGetAttributes(&k) at the call site
inlined    err=400 ptx=-1      // same call inside a helper
noinline   err=400 ptx=-1      // same, __attribute__((noinline))
```

## Fix

Resolve a kernel to its host stub in one place, `getHostStubName`, and use it both as the registration key and as the target of all three symbolic lowerings, so the two cannot disagree:

- clang's device stub when the module has one -- the address user code can take, which is what clang itself registers;
- otherwise a synthetic stub, for kernels outlined from parallel regions, which have no stub of their own.

The synthetic function is now only emitted when actually needed. This closes both standing TODOs at that site, and means a captured kernel pointer called through `cudaLaunchKernel` resolves, which it previously could not.

The `__device_stub__` guard matters: without it a kernel named `main_kernel` maps to `main` and would bind the host `main` as a kernel stub. The negative test pins this.

## Testing

- New lit test covering both the device-stub and outlined-region cases, including that no synthetic stub is emitted in the former and that `main` is not bound in the latter.
- Full lit suite: 1156/1157 pass. The one failure is an untracked local WIP file exercising `convert-parallel-to-gpu1`, a pass this diff does not touch.
- End-to-end CUDA: all three queries in the reproducer return `err=0 ptx=120`, matching vanilla, with launches and results unchanged at `-O0`/`-O2`.

Two known gaps remain, both out of scope here: kernels that are *only* captured and never launched are dropped before reaching a `gpu.module` at `-O0` (no device code emitted at all), which is a plugin-side issue; and a separate `-O2` miscompile in the CUB scan path where `d_temp_storage` is lost in host-side pointer arithmetic.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -2896,6 +2896,9 @@ gdgo->erase();
     symbolTable.getSymbolTable(getOperation());
     getOperation()->walk([&](GPUErrorOp err) {
       std::string sm;
+      // The host symbol whose address the program passes around; outlined
+      // kernels reach here without one, and registration binds it.
+      StringRef hostSymbol;
       if (auto attr =
               dyn_cast_or_null<ArrayAttr>(err->getAttr("passthrough"))) {
         for (auto a : attr) {
@@ -2908,6 +2911,8 @@ gdgo->erase();
               continue;
             if (s0.getValue() == "target-cpu")
               sm = s1.getValue();
+            if (s0.getValue() == "polygeist.host_symbol")
+              hostSymbol = s1.getValue();
             if (backend == "rocm") {
               if (sm.find("sm_") != std::string::npos) {
                 llvm::errs() << "Error: Found NVIDIA architecture while "
@@ -2934,6 +2939,9 @@ gdgo->erase();
             symbolTable.lookupNearestSymbolFrom(launch, launch.getKernel()));
         if (!gfunc)
           return;
+        if (!hostSymbol.empty() && !gfunc->hasAttr("polygeist.host_symbol"))
+          gfunc->setAttr("polygeist.host_symbol",
+                         StringAttr::get(gfunc->getContext(), hostSymbol));
         auto gmod = cast<gpu::GPUModuleOp>(gfunc->getParentOp());
         if (!gmod.getTargetsAttr()) {
           Attribute target;
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -1642,29 +1642,26 @@ static std::string getFuncStubName(StringRef moduleName, StringRef name) {
 // it when it exists; kernels outlined from parallel regions have no stub of
 // their own and fall back to a synthetic one.
 static std::string getHostStubName(Operation *symbolTableOp,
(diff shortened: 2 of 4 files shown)


wsmoses:
Extended per the design discussion on #2896: the host-symbol association is now recorded as an attribute rather than derived by name surgery. Launch recognition stamps `polygeist.host_symbol` (the `reactant$`-stripped stub name) on the kernels it captures, and the registration and kernel-address lowerings prefer the attribute over the naming convention, so the address the program actually passes around is the registered handle even for renamed kernels. Lit test covers the renamed-kernel case.

With this, the identity lowering of `__reactant$get_device_from_host` becomes semantically correct on the runtime MWE from #2896's verification: `cudaFuncGetAttributes` through an opaque pointer now resolves to the raised kernel registered under the original stub and reports the source's `__launch_bounds__` (matching vanilla clang's answer for the same query), and an indirect `cudaLaunchKernel` through that pointer launches the same kernel the query described. If the full MFEM suite (Scan class in particular) validates without #2896's mapping chain, that PR can be dropped in favor of this one.

wsmoses:
Follow-up per review: recognition no longer assumes the `reactant$` prefix. The plugin now records `polygeist.host_symbol` at the source of the association (EnzymeAD/Reactant#84 — the `__cudaRegisterFunction` consumption site and both stub-separation declaration sites), the importer carries it as a passthrough pair, and recognition just copies it onto the kernels it creates. Verified end-to-end (attr present in the import, correct registration binding and runtime query); direct-launch, query-only, launch-bounds, and MFEM TMOP TUs regression-checked.

wsmoses:
Full MFEM validation of this branch's final state (main + this PR + #2859 + #2864, **without** #2896): all 433 objects compile, GPU unit suite **74/74** on a freshly linked binary — including the Inclusive/Exclusive Scan suites, the class #2896 originally fixed — with correct numerics (linf 8.9e-16 quad / 5.6e-17 hex) and the established benchmark profile (derivative globals 0.93–0.98x vs vanilla clang, quad fwd local ~1.07x).

The last commit here (`Decide each kernel's registration stub before conversion starts`) came out of this campaign: dialect conversion moves kernels out of their modules as patterns fire, so re-resolving the kernel mid-conversion to read its attribute could disagree with the registration's earlier decision, leaving a dangling reference to a synthetic stub that was never created (vector.cpp/test_scan.cpp). The (module, kernel) → stub decisions are now computed once over the intact module and shared by every consumer, with first-claimant-wins when several outlined kernels stand for the same host symbol.

PR #2861, opened 2026-08-12
Store a by-value launch argument instead of reading it as an address

wsmoses:
## Problem

Packing kernel arguments for a launch, a `byval` argument is taken to be a pointer to the argument's copy and put into the argument array as-is:

```cpp
bool isByVal = argAttrss && cast<DictionaryAttr>(argAttrss[en.index()])
                               .getNamed(LLVM::LLVMDialect::getByValAttrName());
if (isByVal) fieldPtr = en.value();
```

That holds while the parameter really is a pointer. When the parameter has been scalarised, the `byval` attribute outlives it and the argument arrives **by value** — and the operand's own bits get used as an address. The cast that follows does not even verify:

```
error: 'llvm.bitcast' op can only cast pointers from and to pointers
note: see current operation: %7 = "llvm.bitcast"(%arg0) : (i32) -> !llvm.ptr
```

The added test reproduces it on `main` in 18 lines. Note the kernel parameter is `i32` there and not a pointer: `gpu.launch_func` verifies operand types against the kernel signature, so the stale part is the attribute, not the operand.

## Fix

Key the decision on the operand rather than the attribute alone. A value operand takes the same temporary-and-store path as any other argument, which is what `byval` means in the first place — the slot ends up pointing at the argument's bytes either way.

## Why this matters beyond the verifier error

Found while chasing a CUB `DeviceScan` failure. With the attribute trusted blindly, three of the launch's `byval` slots were block arguments of type `i32`, so the launch received addresses fabricated from argument values. Fixing the cast alone (`inttoptr`) makes the IR valid and the program segfault instead, when `cudaLaunchKernel` dereferences the fabricated pointer. With this change the same case launches cleanly.

## Testing

- New lit test, which errors on `main` and emits `llvm.alloca` + `llvm.store` with the fix.
- Full lit suite: 1156/1157. The single failure is an untracked local WIP file exercising a different pass, unrelated.
- End to end on the CUB reproducer: the illegal memory access and the segfault are both gone and the scan launches with `err=0`. Its results are still wrong, which is a separate defect I am still chasing.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -2188,9 +2188,16 @@ Value ConvertLaunchFuncOpToGpuRuntimeCallPattern::generateParamsArray(
   auto argAttrss =
       dyn_cast_or_null<ArrayAttr>(launchOp->getAttr("reactant.arg_attrs"));
   for (const auto &en : llvm::enumerate(arguments)) {
-    bool isByVal =
-        argAttrss && cast<DictionaryAttr>(argAttrss[en.index()])
-                         .getNamed(LLVM::LLVMDialect::getByValAttrName());
+    // A byval argument reaches the kernel argument array as a pointer to the
+    // argument's copy, which is exactly what the array slot must hold. When
+    // the parameter has been scalarised the attribute outlives it and the
+    // argument arrives by value; a value needs the same temporary and store as
+    // any other, since reinterpreting it would hand the launch an address made
+    // out of the argument's own bits.
+    bool isByVal = argAttrss &&
+                   cast<DictionaryAttr>(argAttrss[en.index()])
+                       .getNamed(LLVM::LLVMDialect::getByValAttrName()) &&
+                   isa<LLVM::LLVMPointerType>(en.value().getType());
     Value fieldPtr;
     if (isByVal) {
       fieldPtr = en.value();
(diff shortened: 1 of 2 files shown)


wsmoses:
Closing this — you were right that it is a side effect of #2863 rather than a defect of its own.

`reactant.arg_attrs` is built in `GPULaunchRecognition.cpp:699` by walking the kernel function's `getArgAttrs()` in order, one entry per parameter, and `gpu.launch_func`'s kernel operands are 1:1 with parameters as well. So `argAttrss[i]` and `getKernelOperands()[i]` describe the same parameter by construction.

Before #2863, `arguments` came from `adaptor.getOperands().take_back(numKernelOperands)` — not a slice of the kernel operands at all, but the tail of the whole operand list, which on an async launch ends with the stream. That is a rotation: first real argument dropped, stream appended, count unchanged. `argAttrss[i]` was then describing operand *i-1*, which is how a slot marked `byval` came to be handed a non-pointer and `fieldPtr = en.value()` came to treat a value as an address.

The justification I wrote here — that a scalarised parameter leaves the attribute behind — was a hypothesis to explain that mismatch, not something observed. To check it I instrumented `generateParamsArray` to report every slot carrying a `byval` attr and ran it over all 273 MFEM CUDA translation units plus the CUB-scan, byval-struct and forall reproducers: no slot with a `byval` attr occurs at all, so the branch this guards is never taken and the added conjunct cannot be doing anything.

Leaving the comment in the tree would be worse than not having it, since it asserts a mechanism I cannot demonstrate.

PR #2862, opened 2026-08-12
ConvertPolygeistToLLVM: keep the stream out of the kernel's parameter array

### LINKED ISSUE #2854: thrust/CUB kernels fail at runtime since the jax bump in #2698
| benchmark | exit | failure |
|---|---|---|
| XSBench | 139 | SIGSEGV inside `cuLaunchKernel` |
| RSBench | 139 | SIGSEGV inside `cuLaunchKernel` (at `BLOCKSIZE=256`) |
| LULESH  | 134 | `cudaErrorMisalignedAddress` (`cudaErrorIllegalAddress` on some builds) |
| LBM     | 0   | runs |

Using
```bash
make -B && ./XSBench -m event                                    # → 139
make -B && ./lulesh -s 45                                        # → 134
BLOCKSIZE=256 make -B && ./rsbench -m event -l 10200000          # → 139
```
RSBench needs `BLOCKSIZE=256`. At the Makefile default of 32 it needs 318,750 blocks and dies earlier with `cudaErrorInvalidConfiguration` from #2789.

The two failure modes differ in when they surface, but both are consistent with a bad thrust/CUB launch:

- XSBench,RSBench die on the host side, inside the driver. From RSBench:
  ```
  #8  cuLaunchKernel () from libcuda.so.1
  #10 cudaLaunchKernel ()
  #11 run_event_based_simulation(...)
  #12 main ()
  ```
- LULESH does not fail at the launch. It is reported at the next synchronizeing call:
  ```
  terminate called after throwing an instance of thrust::system::system_error
    what():  trivial_device_copy D->H failed: cudaErrorIllegalAddress
  ```

    The reported error is not stable across builds, another build of the same source reported `cudaErrorMisalignedAddress` from `parallel_for: failed to synchronize`.

**Extra info: The invalid `gpu.launch` IR that #2698 also introduced, it is already fixed by #2844**

**What #2698 changes**
```
workspace.bzl                    JAX_COMMIT     269d3f7b → cec06d11
workspace.bzl                    ENZYME_COMMIT  1b8c5bfa → 54e6d5c5
patches/jax.patch                fixed up for the new JAX
patches/xla_stream_pool.patch    deleted, "now included in XLA properly"
patches/llvm.patch               deleted, "now upstreamed"
```
and Passes
`ConvertParallelToGPU.cpp`, `ConvertPolygeistToLLVM.cpp`, `GPULaunchRecognition.cpp`, `LowerKernel.cpp`, `ParallelLower.cpp`

xys-syx:
Partially fix #2854 

After the fix, RSBench and XSBench will not segfault in the runtime.
**The root cause**:
`generateParamsArray` took the launch's arguments with `adaptor.getOperands().take_back(getNumKernelOperands())`, but `gpu.launch_func` declares `asyncObject` after `kernelOperands`. On a launch carrying a stream that window sits one operand too far right: the first kernel argument falls out and the stream lands in the last slot, which the driver dereferences as the address of the last parameter. The array is still the right length, so nothing downstream notices.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -2160,9 +2160,8 @@ Value ConvertLaunchFuncOpToGpuRuntimeCallPattern::generateParamsArray(
     gpu::LaunchFuncOp launchOp, OpAdaptor adaptor, OpBuilder &builder,
     Block *allocaBlock) const {
   auto loc = launchOp.getLoc();
-  auto numKernelOperands = launchOp.getNumKernelOperands();
   SmallVector<Value, 4> arguments =
-      adaptor.getOperands().take_back(numKernelOperands);
+      llvm::to_vector<4>(adaptor.getKernelOperands());
   auto numArguments = arguments.size();
   SmallVector<Type, 4> argumentTypes;
   argumentTypes.reserve(numArguments);
--- a/test/lit_tests/lowering/cuda_async_codegen.mlir
+++ b/test/lit_tests/lowering/cuda_async_codegen.mlir
@@ -31,69 +31,70 @@ module attributes {gpu.container_module} {
 }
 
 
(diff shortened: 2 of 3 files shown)


wsmoses:
superceeded by conflicting fix to main

PR #2864, opened 2026-08-12
Treat an address-taken host stub as capturing its kernel

wsmoses:
## Problem

The capture scan in `GPULaunchRecognition` only walks uses of the **device-side** function. But the address user code takes is that of clang's **host stub**, and when it escapes — handed to a call rather than reaching a runtime query directly — there is no call site at which to rewrite it to the device symbol.

The kernel then looks like an ordinary launch: it is lowered to a parallel region inside a stub nothing calls, never reaches a `gpu.module`, and so no device code is emitted for it and nothing registers it. A later `cudaFuncGetAttributes` on that address returns 400 (invalid resource handle).

Optimized builds hide this, since inlining brings the address back to a call site where the existing rewrite fires. It shows up at `-O0`, and in CUB's `PtxVersion()`, which queries `EmptyKernel<void>` through a helper.

## Fix

Also treat an address-taken host stub as a capture of its kernel, using the same `reactant$` naming relation the registration path relies on.

## Testing

- New lit test: a kernel whose host stub's address escapes into a call is promoted to a `gpu.func`. Nothing is promoted without the change.
- Full lit suite: 1156/1157, the remaining failure being an untracked local WIP file exercising a different pass.
- End to end at `-O0`, queries that returned 400 now return the same attributes as vanilla clang, for a plain kernel, a captured-only kernel, and a `linkonce_odr` template kernel.

## Draft, because of the ordering

Promoting kernels that were previously never promoted exercises code that nothing reached before, and it uncovered three latent bugs. This wants to land **after** all of them:

- #2858 — promoting a kernel with no argument attributes dereferences an empty optional. Hit directly by this change; the test here gives its kernel an attributed argument to stay independent, but real `-O0` input does not.
- #2861 — a scalarised `byval` argument is read as an address.
- #2863 — kernel operands sliced from the back pick up the stream.

Landing this first would turn a quiet mis-registration into build failures on CUB-using code.

### DIFF
--- a/src/enzyme_ad/jax/Passes/GPULaunchRecognition.cpp
+++ b/src/enzyme_ad/jax/Passes/GPULaunchRecognition.cpp
@@ -477,6 +477,39 @@ enum __device_builtin__ cudaMemcpyKind
           }
         }
       }
+      // The scan above only sees uses of the device-side function. An address
+      // user code takes is that of clang's host stub, and when it escapes --
+      // handed to a call rather than straight to a runtime query -- nothing
+      // rewrites it to the device symbol, so the kernel would never reach a
+      // gpu.module and never be registered. An address-taken host stub is a
+      // capture of its kernel.
+      if (!captured) {
+        StringRef hostStubName;
+        if (auto attr = dyn_cast_or_null<ArrayAttr>(
+                launch.first.getPassthroughAttr())) {
+          for (auto a : attr) {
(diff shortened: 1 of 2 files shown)


wsmoses:
Rebased onto main and switched the host-stub lookup to the recorded `polygeist.host_symbol` passthrough pair (EnzymeAD/Reactant#84) — no `reactant$` name surgery; lit test updated to carry the attribute as the plugin emits it.

PR #2865, opened 2026-08-12
Stop reading through a truncation when it changes the number

wsmoses:
Building an affine index walks back through the arithmetic that produced it, and treated a truncation as one more cast to see past. A widening leaves the number alone, but a truncation is a *different number* wherever the bits it drops were set, so seeing past one puts those bits back into the index.

CUDA hands a launch its `dim3` coerced into an `i64` holding x and y side by side, and x comes back out as a truncation of it:

```mlir
%2 = arith.shli %hi_i64, %c32 overflow<nuw> : i64
%3 = arith.ori  %2, %lo_i64 {isDisjoint} : i64   // the packed dim3
%4 = arith.trunci %3 : i64 to i32                // x
%5 = arith.index_cast %4 : i32 to index
```

Three places read through it, and all three stop:

- `isValidIndex`, which answered "is this a valid index?" by looking at what the truncation truncated;
- the operand legalization's decast loop, which stripped it to find something to use as a symbol;
- `AffineExprBuilder::buildExpr`, which had `arith::TruncIOp`/`LLVM::TruncOp` in the same `isa<>` list as the extensions and the index casts.

So an access at `x` became one at `y*2^32 + x`, and the truncated value's role was taken over by an `arith.index_cast %3 : i64 to index` of the whole packed word.

The predicate is the important one: `isValidIndex` saying yes is a promise the legalization has to make good on, and it reports a fatal error when it cannot. With it refusing too, such an operand is never offered, so the other two sites simply never meet one.

### Where it showed up

In MFEM's `quadrature_interpolator::Det2D<2,3>` the affected launch is the one the kernel's own extents come from. `%3` reached `gpu.launch_func` as a kernel operand in place of a block extent, so the kernel received `(Q1D<<32)|Q1D` where it wanted `Q1D`:

```
Invalid __global__ read of size 8 bytes
  by thread (0,0,0) in block (0,0,0)
  Access at 0x7b2a18400400 is out of bounds
  and is 103,079,214,080 bytes before the nearest allocation
```

103,079,214,080 bytes is 3·2³² doubles — the packed `Q1D = 3` sitting in the high half. The test aborts later, in an unrelated `cudaFree`.

### Test

`test/lit_tests/raising/trunc-packed-index.mlir`: the packed-dim3 shape, where the raised load must keep reading what the unraised store writes; and a truncation with nothing suspicious about it, which is left standing rather than dropped.

`test/lit_tests/raising/masked_affine_for.mlir` raises exactly as before and is unchanged.

### Verification

Full lit suite passes. Every one of MFEM's 273 CUDA translation units compiles clean, and the out-of-bounds read above is gone.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -603,13 +603,6 @@ AffineApplyNormalizer::AffineApplyNormalizer(AffineMap map,
         decast = idx.getIn();
         continue;
       }
-      // A widening leaves the number alone, but a truncation is a different
-      // number wherever the discarded bits were set, so dropping one puts
-      // those bits back. CUDA packs a dim3's x and y into a single i64 and
-      // truncates to recover x; using the packed value in its place makes
-      // the operand y*2^32 + x. Nothing here can tell the two apart, so keep
-      // the operand as it stands -- mid-chain is not a place to stop, since
-      // the value there has the wrong type to stand in for it.
       if (decast.getDefiningOp<TruncIOp>()) {
         decast = t;
         break;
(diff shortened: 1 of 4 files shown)


xys-syx:
Hello, I have independently verified on LULESH (sm_120, CUDA 12.9), which is a different reproducer from the MFEM one:
- The bad constant is gone: -549755813888 (= -128 × 2³², the block size sitting in the high half) appeared twice in the raised IR on main and zero times with this PR.
- main_kernel_486, the shared-memory min-reduction that was faulting, now runs. The failure moves to the next kernel, main_kernel_487

PR #2866, opened 2026-08-12
Only lift a barrier out of a branch every thread decides alike

wsmoses:
`BarrierHoist` moves a barrier out of an `scf.if`/`affine.if` whenever nothing side-effecting follows it in the body, and symmetrically before the branch when nothing precedes it. Lifting one out makes **every** thread run it where before only the threads taking the branch did, so it needs one of two things to hold:

- the branch selects no subset of the threads, so all of them ran it already; or
- it moves past the branch, nothing beyond holds a barrier for the threads that skipped to run instead, and none is left behind inside for the ones that took it.

The upward move is only sanctioned by the first.

### What went wrong

MFEM's `SmemPAMassApply3D_Element` has eight `MFEM_SYNC_THREAD` under `if (k >= N) { return; }`. Only the trailing one can move — stores follow the rest, so `below` is false for them — and lifting just that one gives:

```mlir
scf.if %17 { ... 7 gpu.barriers ... }
gpu.barrier      // lifted out
gpu.return
```

A thread with `k >= N` waits at the barrier out here; a thread with `k < N` waits at the first one still in there. Neither ever reaches the other's. With `TBATCH=2` the launch is `blockDim (4,4,2)` — exactly one warp — and `k = blockIdx.x*blockDim.z + threadIdx.z` varies within it, so a tail block splits its own warp. `cuda-gdb` on the live hang:

```
Grid 205  Status Active   SMs Mask 0x...2000
GridDim (14,1,1)  BlockDim (4,4,2)
```

One block resident, thirteen retired — a wedged block, not a runaway loop. Hoisting all eight would have been fine, and hoisting none would have been fine; hoisting exactly one is the defect.

`Det2D` escapes on geometry alone: `NBZ=1` makes `blockDim.z=1`, so `k` is decided alike by the whole block.

### Deciding what the threads settle alike

The barrier's operands name the thread indices it synchronises over, and so the parallel whose iterations are the threads. Walking the condition from there: anything defined outside that parallel holds the same value in every thread, whatever computed it. Inside it, a block argument is something like a serial loop's index — which is not uniform, since that loop's own bounds may be thread-dependent — so the walk stops rather than assuming.

### Test

`test/lit_tests/barrier-hoist-divergent.mlir`, three cases:

- `@lone` — a single barrier under a thread-dependent branch. Lifting it is a legalization: every thread then runs it, at one instruction, with nothing after. **Must still hoist.**
- `@two` — a second barrier pinned inside by a store between them. **Must not hoist.**
- `@serial_iv` — the condition reads an `scf.for` index whose bound is a thread index. **Must not hoist**, and would have been mistaken for uniform by a walk that stops at block arguments.

### Result

`H1 Assembly Levels` no longer hangs: **0.82 s**, against 1.10 s for plain clang, where it previously spun indefinitely. Full lit suite passes (1163/1164; the one failure is an untracked local file that fails without this change too).

The test does not yet **pass** — 104 of 121 assertions do, 17 fail on values. That is a separate defect underneath this one, still being chased; this PR is the deadlock only.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Ops.cpp
@@ -1534,13 +1534,14 @@ llvm::cl::opt<bool> BarrierOpt("barrier-opt", llvm::cl::init(true),
                                llvm::cl::desc("Optimize barriers"));
 
 // Whether every thread runs `ifOp`'s body, i.e. the condition selects no subset
-// of them. `threadIVs` are the indices the barrier synchronises over, which
-// name the parallel whose iterations are the threads.
+// of them. `threadIVs` are the indices the barrier synchronises over, which name
+// the parallel whose iterations are the threads.
 static bool selectsAllThreads(Operation *ifOp, ValueRange threadIVs) {
+  // Synchronising over no index at all: there is no dimension for the
+  // condition to vary along, so no subset for it to select.
   if (threadIVs.empty())
-    return false;
-  Operation *par =
-      cast<BlockArgument>(threadIVs.front()).getOwner()->getParentOp();
+    return true;
+  Operation *par = cast<BlockArgument>(threadIVs.front()).getOwner()->getParentOp();
 
   SmallVector<Value> worklist;
   if (auto scfIf = dyn_cast<scf::IfOp>(ifOp))
--- a/src/enzyme_ad/jax/Dialect/Ops.cpp
+++ b/src/enzyme_ad/jax/Dialect/Ops.cpp
@@ -1533,18 +1533,27 @@ using namespace mlir::enzyme;
 llvm::cl::opt<bool> BarrierOpt("barrier-opt", llvm::cl::init(true),
                                llvm::cl::desc("Optimize barriers"));
 
-// Whether every thread in the block decides `ifOp`'s condition the same way,
-// i.e. none of `threadIVs` reaches it.
-static bool isBlockUniform(Operation *ifOp, ValueRange threadIVs) {
-  SmallVector<Value> conds;
+// Whether every thread runs `ifOp`'s body, i.e. the condition selects no subset
+// of them. `threadIVs` are the indices the barrier synchronises over, which
+// name the parallel whose iterations are the threads.
+static bool selectsAllThreads(Operation *ifOp, ValueRange threadIVs) {
+  Operation *par = nullptr;
+  for (Value iv : threadIVs)
+    if (auto ba = dyn_cast<BlockArgument>(iv)) {
+      par = ba.getOwner()->getParentOp();
+      break;
+    }
+  if (!par)
+    return false;
+
+  SmallVector<Value> worklist;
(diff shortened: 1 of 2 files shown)


wsmoses:
**Correction — the rationale above is wrong, and this PR is aimed at the wrong place.**

`enzymexla.barrier` is `__syncthreads` (PPoPP'23 §3.1: "all threads in a block have finished executing all instructions prior to the function call, before any threads executes any instruction after the call"), with the operands naming the thread indices synchronised over, expressed through memory effects rather than as an opaque barrier.

§3.2.2 gives the model's only legalization for a barrier nested in control flow: bracket the construct with barriers, split the parallel loop around it, then *interchange the construct with the parallel for*. Interchanging `parallel { if (c) {...} }` into `if (c) { parallel {...} }` is valid only where `c` is invariant with respect to the parallel IVs.

So a barrier under a thread-dependent condition has no legalization, and by the syncthreads equivalence it is UB. The input in `test/lit_tests/barrier-hoist-divergent.mlir` is therefore already invalid IR, and `BarrierHoist` is not the defect — it is a pass that turns latent UB into an observable deadlock. Refusing the hoist would only hide IR that should not have been built.

The real defect is upstream, where that IR is created: MFEM's

```cpp
if (k >= N) { return; }
body(k);   // contains 8 __syncthreads
```

becomes `scf.if (k < N) { ...8 barriers... }` by `enzyme-lift-cf-to-scf` or earlier. That conversion loses the one fact that makes the CUDA legal: a thread that `return`s has **exited**, and exited threads are excluded from `__syncthreads`, whereas an `scf.if` means divergence. The model has no way to say "this thread is done", so a kernel with an early return above a barrier cannot be represented faithfully.

Leaving this open only as a record of the analysis; the fix belongs in lowering a tail-position guard as an early `gpu.return` so the skipped threads exit before any barrier, which is what plain clang emits. Do not merge as-is.

wsmoses:
Body and code replaced: the rule is now `selectsAllThreads(branch) || (moving past it && no barrier beyond && none left inside)`, per review. Both earlier framings above are withdrawn — the first was too broad (it would have blocked `@lone`, which is a legal legalization), and the second wrongly concluded the input IR was invalid and this the wrong fix site.

wsmoses:
Review fixes pushed (8ee1c30a1).

`selectsAllThreads`: returns false on empty `threadIVs`, and takes the parallel from `cast<BlockArgument>(threadIVs.front())` — every index a barrier synchronises over is one, so there is no case where the old scan found none and it fell through to yes.

`anyBarrierAfter`: was only reading the operations beside the branch. It now continues into the parent, stops at the thread parallel where a thread's execution ends, and counts a barrier anywhere in a parent that may run more than once, since such a region puts its own earlier operations after this one. Same traversal shape as `getEffectsAfter`.

Fourth test case `@loop_wraparound` covers exactly that: the branch is last in an `scf.for` body with a barrier at the top of the loop, so what follows a thread is the next iteration. The old code hoisted it.

Lit 1163/1164. Re-running MFEM against this revision.

PR #2869, opened 2026-08-12
Keep kernels whose grid and block parallels were fused

wsmoses:
## The bug

`SplitParallelOp` (`ConvertParallelToGPU.cpp`) gated its exact-shape fallback on

```cpp
pop.getUpperBound().size() == 6 && pop.getUpperBound() == wrapper.getOperands()
```

A kernel **with no barrier** has its grid and block `scf.parallel` ops fused into one by earlier passes, and dimensions of extent one are dropped, so the count is 5 or 4 rather than 6 and the fallback never fires. Every candidate block size then fails too, and the `curRegion == 0` path erased the wrapper and replaced the launch result with a `CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES` constant.

The consequence is a silent miscompile, not a diagnosed failure:

- the kernel disappears from the device module entirely,
- the launch becomes a no-op, so `cudaGetLastError()` still reports success,
- the output buffer keeps whatever it held before.

MFEM's CUDA build hit this on every kernel without a `MFEM_SYNC_THREAD`: all nine `EAMassAssemble1D<*>` instantiations, all three `PADiffusionSetup{2D<2>,2D<3>,3D}`, and two convection assembly kernels. The diffusion ones compute the geometric factors shared by every assembly level, which is why one dropped kernel failed `PARTIAL`, `ELEMENT` and `FULL` alike. 13 of the 17 failing `H1 Assembly Levels` assertions come from this.

## The fix

`matchWrapperShape()` maps the parallel's dimensions onto the wrapper's six bounds instead of demanding equality:

- a wrapper bound of extent one may have no corresponding dimension (it was dropped), and `exactMatch()` reinstates it as `0..1 step 1`;
- a bound may be narrowed to `min(N, bound)`, which is what folding the `k >= N` guard into the loop leaves behind. The narrowed bound is the real iteration space, so the launch is built from it, not from the wrapper operand.

When nothing matches even then, the pass now emits an error saying the kernel would be dropped, instead of quietly producing a program that computes the wrong answer.

## Tests

Three cases added to `test/lit_tests/lowering/convert-parallel-to-gpu1-blocksize-fallback.mlir`:

- `@fused_unit_dim` — five bounds carrying the wrapper's six; the split still applies and reinstates grid.z.
- `@min_narrowed_bound` — a bound behind an `arith.minsi`; matched, and the launch uses the narrowed value.
- `@allfail` (existing) — still drops, now with `expected-error`, so the diagnostic is covered.

## Measured

MFEM `H1 Assembly Levels`, swapping only the affected objects into an otherwise vanilla-clang build:

| TU | before | after |
|---|---|---|
| `bilininteg_mass_ea.cpp` | 1 failed | **0** |
| `bilininteg_diffusion_kernels.cpp` | 12 failed | **0** |
| `bilininteg_convection_ea.cpp` | 4 failed | 4 failed |

Convection no longer drops any kernel, so its remaining 4 failures are a separate defect; I am still chasing it and it is not addressed here.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -172,6 +172,46 @@ template <int S = 3> SmallVector<Value, S> getUpperBounds(scf::ParallelOp pop) {
   return bounds;
 }
 
+/// Whether `ub` is the launch bound `bound`, possibly narrowed by a min against
+/// the kernel's own trip count -- folding the `k >= N` guard into the loop
+/// makes the bound `min(N, bound)`. The narrowed bound is the real iteration
+/// space, so the launch is built from it, not from `bound`.
+bool boundMatchesLaunch(Value ub, Value bound) {
+  if (ub == bound)
+    return true;
+  if (auto mn = ub.getDefiningOp<arith::MinSIOp>())
+    return mn.getLhs() == bound || mn.getRhs() == bound;
+  if (auto mn = ub.getDefiningOp<arith::MinUIOp>())
+    return mn.getLhs() == bound || mn.getRhs() == bound;
+  return false;
+}
+
+/// Match the dimensions of `pop` against the six grid and block bounds of
+/// `wrapper`, filling `dimOf` with the dimension of `pop` that carries each
+/// bound, or -1 for a bound of extent one that `pop` does not carry. Passes
+/// that fuse the grid and block parallels drop such trivial dimensions, so an
+/// exact-shape match must tolerate their absence.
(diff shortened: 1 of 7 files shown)


wsmoses:
Superseded — split into logically isolated PRs, each with its own tests, and rebased onto main (this branch was accidentally cut from `integration-all`, so it also carried #2857/#2859/#2867's commits):

- #2870 — do not take a grid dimension into an already full block. Independent, no known fallout, fixes the 4 Convection assertions.
- #2871 — say when a kernel is dropped instead of dropping it silently. Diagnostic only, no codegen change.
- #2872 — keep kernels whose grid and block parallels were fused. Fixes the 13 Mass/Diffusion assertions, but **draft**: it makes 33 `fem/tmop` TUs fail to build by exposing an unresolved-stub path, documented in the PR.

PR #2873, opened 2026-08-12
Give shared-memory globals the alignment the middle end assumes

wsmoses:
## The bug

`SharedLLVMAllocaToGlobal` / `SharedMemrefAllocaToGlobal` created the shared-memory globals with no explicit alignment. That is not a neutral choice in LLVM:

- the middle end assumes a definition without explicit alignment carries the **preferred** alignment — `DataLayout::getPreferredAlign` rounds anything larger than 128 bits up to **16** — and `opt -O3` stamps `align 16` on loads through it, which the load/store vectorizer then turns into 16-byte accesses;
- `NVPTXAsmPrinter` emits `.align` from `getAlign().value_or(getPrefTypeAlign(ETy))` — the *type's* alignment, **8** for double arrays — never consulting `getPreferredAlign`;
- ptxas honors `.align 8` and packs the globals, so a 24-byte array upstream leaves the next one at an address that is 8 mod 16.

Net effect: `ld.shared.v2.b64` at a misaligned address. compute-sanitizer on MFEM:

```
Invalid __shared__ read of size 16 bytes
  at ...SmemPAMassApplyTriangle<2, 2>...kernel+0x8b0
  Access at 0x4a8 is misaligned
```

The error is sticky, so it surfaces at whatever `cudaMemcpy`/`cudaDeviceSynchronize` comes next — four test cases (`PA Simplices`, `DG Mass Inverse`, `PA DG Diffusion`, `VecDivPA`) fail this way with no indication of the real culprit. Reproduced with stock `opt -O3` + `llc -march=nvptx64` alone on the extracted module: loads say `align 16`, the PTX says `.align 8`.

Vanilla clang never hits this because it always emits explicit alignment on `__shared__` globals. In our pipeline the alignment actually survives most of the way — the imported clang global carries `align 8`, `ParallelLower` copies it onto the alloca, the `gpu.launch` split copies it again — and was dropped only at this last step.

## The fix

Keep the alloca's alignment, and raise it to the preferred alignment (16 for anything over 128 bits) so the middle end's assumption is spelled out explicitly and the vectorized accesses stay legal — rather than clamping to 8 and losing the vectorization.

Arguably `NVPTXAsmPrinter` should use `DL.getPreferredAlign(GV)` and this is worth an upstream issue; but emitting globals whose alignment the rest of the compiler has to guess is fragile regardless, and this makes the IR self-contained.

## Test

`test/lit_tests/lowering/shared-alloca-alignment.mlir`: four allocas through `convert-parallel-to-gpu2` —

- `memref<2xi32, 5>` (64 bits) → `alignment = 4` (no bump below the threshold)
- `memref<3xf64, 5>` (192 bits) → `alignment = 16` (the bump)
- `memref<2x8xf64, 5>` with `alignment = 32` → `alignment = 32` (explicit alignment preserved, not clobbered)
- `llvm.alloca !llvm.array<4 x f64>` → `alignment = 16` (the LLVM-dialect path)

## Measured

After recompiling only `bilininteg_mass_kernels.cpp`, the sanitizer fault in `SmemPAMassApplyTriangle<2,2>` is gone and moves to `SmemPADiffusionApplyTriangle<2,2>` in the not-yet-recompiled diffusion TU — confirming the mechanism kernel by kernel. Shared globals in the rebuilt TU: 495 now `.align 16`, small ones unchanged (`.align 4`/`.align 8`). Full-library rebuild + suite run in progress; will report the four test cases in a follow-up comment.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -313,6 +313,7 @@ struct SharedLLVMAllocaToGlobal : public OpRewritePattern<LLVM::AllocaOp> {
         rewriter, loc, ao.getElemType(), /* isConstant */ false,
         LLVM::Linkage::Internal, name, mlir::Attribute(),
         /* alignment */ 0, /* addrSpace */ 3);
+    globalOp.setAlignmentAttr(ao.getAlignmentAttr());
     rewriter.setInsertionPoint(ao);
     auto aoo = LLVM::AddressOfOp::create(rewriter, loc, globalOp);
 
@@ -348,7 +349,7 @@ struct SharedMemrefAllocaToGlobal : public OpRewritePattern<memref::AllocaOp> {
     memref::GlobalOp::create(rewriter, loc, rewriter.getStringAttr(name),
                              /* sym_visibility */ mlir::StringAttr(),
                              mlir::TypeAttr::get(type), initial_value,
-                             mlir::UnitAttr(), /* alignment */ nullptr);
+                             mlir::UnitAttr(), ao.getAlignmentAttr());
     rewriter.setInsertionPoint(ao);
     auto getGlobalOp = memref::GetGlobalOp::create(rewriter, loc, type, name);
 
--- a/test/lit_tests/lowering/shared-alloca-alignment.mlir
+++ b/test/lit_tests/lowering/shared-alloca-alignment.mlir
@@ -1,0 +1,35 @@
+// RUN: enzymexlamlir-opt %s --pass-pipeline="builtin.module(convert-parallel-to-gpu2{backend=cuda})" | FileCheck %s
+
+module attributes {gpu.container_module} {
+  gpu.module @kern {
+    gpu.func @kern(%out: memref<?xf64, 1>) kernel {
+      %c0 = arith.constant 0 : index
+      %plain = memref.alloca() : memref<2xi32, 5>
+      %aligned = memref.alloca() {alignment = 8 : i64} : memref<3xf64, 5>
+      %over = memref.alloca() {alignment = 32 : i64} : memref<2x8xf64, 5>
+      %one = llvm.mlir.constant(1 : i64) : i64
+      %raw = llvm.alloca %one x !llvm.array<4 x f64> {alignment = 16 : i64} : (i64) -> !llvm.ptr<5>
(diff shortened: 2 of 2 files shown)


wsmoses:
Revised per review: no invented default. The globals now carry the alloca's alignment when it has one (which is the normal case — clang's align 8 survives the whole round trip), and the type's natural ABI alignment otherwise. The fallback exists because in LLVM an *empty* alignment field on a global definition does not mean natural alignment — the middle end may assume preferred alignment (16 for >128-bit objects, `DataLayout::getPreferredAlign`) while the NVPTX printer emits only the type's alignment, which is the mismatch that caused the miscompile. With an explicit 8 the middle end reasons from 8 and the bogus 16-byte vectorization never forms, matching vanilla clang's output. The earlier revision's bump-to-16 (to keep the vectorization) is withdrawn — that would be a separate performance decision, not part of this fix.

wsmoses:
Revised again per review: pure forwarding, no default, comments removed. The global gets exactly the alloca's alignment attr (`ao.getAlignmentAttr()`), nothing when the alloca has none.

Does it change what passes? **No.** Rebuilt the simplex-kernel TUs and reran: DG Mass Inverse, PA DG Diffusion, VecDivPA, Vector Diffusion Diagonal PA all pass; PA Simplices unchanged at 7/8 (its remaining failure is a different bug — a guarded load speculated by affine-cfg, tracked separately). Mechanically: shared globals now come out `.align 8`/`.align 4` forwarded from clang, and the rebuilt TU has zero `ld.shared.v2` — with an explicit alignment present the middle end reasons from it and the 16-byte-assuming vectorization never forms, same as vanilla clang's output.

wsmoses:
Done — the attribute is now transferred verbatim on both paths. The memref path passes `ao.getAlignmentAttr()` straight into the `memref.global` builder; the LLVM path's convenience builder only takes a `uint64_t`, so the global is created without alignment and `setAlignmentAttr(ao.getAlignmentAttr())` transfers the attr directly (null passes through as unset). The whole code diff is now:

```
+    globalOp.setAlignmentAttr(ao.getAlignmentAttr());
-                             mlir::UnitAttr(), /* alignment */ nullptr);
+                             mlir::UnitAttr(), ao.getAlignmentAttr());
```

Added a no-alignment `llvm.alloca` case to the lit test so the null transfer is covered on both paths. Re-verified end-to-end: mass-kernels PTX unchanged (`.align 8`/`.align 4`, zero `ld.shared.v2`).

PR #2874, opened 2026-08-12
Restate known_block_size as nvvm.maxntid on outlined kernels

wsmoses:
## The bug

Kernels outlined from a launch with constant block dimensions carry them as `known_block_size = array<i32: x, y, z>` on the `llvm.func` inside the `gpu.module`. The NVVM translation, however, only reads `nvvm.maxntid` (`NVVMToLLVMIRTranslation.cpp`), so no kernel we emit ever gets launch bounds — ptxas allocates registers with no thread budget in mind.

For most kernels that's invisible. For a kernel whose block is known to be large it is fatal: MFEM's `QuadratureInterpolator` determinant kernel at `p=6, q=8` launches an 8×8×8 block, the raised kernel's register usage doesn't fit 512 threads unbounded, and **every** launch fails:

```
CUDA error: (cudaGetLastError()) failed with error:
 --> too many resources requested for launch [code: 701]
 ... in function: void mfem::CuWrap3D(...) [DBODY = (lambda at fem/qinterp/det.hpp:286:50) &]
```

The existing `AddLaunchBounds` pattern was meant to cover this but (a) is disabled by default, and (b) sets `nvvm.maxntidx` as an integer attribute — the dialect wants `nvvm.maxntid` as an i32 array — so the translation ignores it even when enabled (its own TODO notes the attr name was never updated).

## The fix

In `convert-parallel-to-gpu2` (cuda backend), walk the GPU modules and restate `known_block_size` as `nvvm.maxntid` on kernel functions that don't already carry one. The translation then emits it as the function's launch bounds and ptxas caps register allocation to what the block size admits — exactly what `__launch_bounds__` does for any CUDA kernel, and safe here because each callsite gets its own outlined kernel, so the recorded block size is exact.

## Test

`test/lit_tests/lowering/kernel-maxntid.mlir`: three kernels through the pass —

- `known_block_size = [8, 8, 8]` → gains `nvvm.maxntid = [8, 8, 8]`
- no `known_block_size` → unchanged (no invented bounds)
- pre-existing `nvvm.maxntid = [2, 2, 2]` → preserved, not clobbered

End-to-end verification against MFEM's `QuadratureInterpolator` test is queued behind a full library rebuild (in flight together with #2872/#2873); will report in a follow-up comment.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -3615,8 +3615,15 @@ public:
     // Add a dialect specific kernel attribute in addition to GPU kernel
     // attribute. The former is necessary for further translation while the
     // latter is expected by gpu.launch_func.
-    if (gpuFuncOp.isKernel())
+    if (gpuFuncOp.isKernel()) {
       attributes.emplace_back(kernelAttributeName, rewriter.getUnitAttr());
+      if (kernelAttributeName == NVVM::NVVMDialect::getKernelFuncAttrName())
+        if (auto blockSize = gpuFuncOp.getKnownBlockSizeAttr())
+          attributes.emplace_back(
+              StringAttr::get(rewriter.getContext(),
+                              NVVM::NVVMDialect::getMaxntidAttrName()),
+              blockSize);
+    }
     auto llvmFuncOp = LLVM::LLVMFuncOp::create(
         rewriter, gpuFuncOp.getLoc(), gpuFuncOp.getName(), funcType,
         LLVM::Linkage::External, /*dsoLocal*/ false, /*cconv*/ LLVM::CConv::C,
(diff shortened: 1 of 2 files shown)


wsmoses:
Corrected: the walk originally ran in `convert-parallel-to-gpu2`, where kernels are still `gpu.func`s — it matched nothing in the real pipeline (my lit test used `llvm.func` directly, so it passed while the end-to-end PTX had zero `.maxntid`). Moved to `strip-gpu-info`, the last pass before serialization; on MFEM's mass-kernels module all 21 constant-block-size kernels now carry `nvvm.maxntid`, verified on the extracted pre-binary module.

One scope note: MFEM's failing `QuadratureInterpolator` case turns out to be the `Det3D<0,0>` *fallback* instantiation with runtime block dimensions — no `known_block_size` exists for it, so no launch-bounds attribute can cover it. That kernel fails because the raised code simply uses more registers than clang's; the fix here prevents the class of failures where the block size is known, but the runtime-dims case is the register-pressure problem itself and stays open.

wsmoses:
Relocated per review: the walk over already-lowered functions in `strip-gpu-info` was a convenience placement. The attribute is now emitted in `ConvertPolygeistToLLVM`'s `GPUFuncOpLowering` — the same place the kernel's `llvm.func` already receives `nvvm.kernel` — gated on the NVVM kernel-attribute name, so the rocm path is untouched. Lit test updated to drive `convert-polygeist-to-llvm{backend=cuda}` on `gpu.func`s directly. End-to-end: MFEM's mass-kernels TU now emits 21 `.maxntid` in PTX, one per constant-block-size kernel.

PR #2875, opened 2026-08-13
Do not lift a guarded load into the conditional feeding its index

wsmoses:
## The bug

`LiftMemrefRead` (affine-cfg) rewrites a load whose index uses a conditional's results by cloning the load into that conditional's branches, each branch loading with its own arm of the index. The rebuilt conditional stands at the **original conditional's position** — which is where the load now executes — and nothing checked that the load was reached from there unconditionally. Three unsoundnesses with one root:

- a load nested in a further `if` was hoisted out of it, evaluated on paths its guard excluded;
- a load behind intervening stores would read stale memory;
- a load inside a loop between the two would change trip count.

Minimal reproducer (26 lines): with `a1 = select(y<2, y, y-2)` and a guarded load,

```mlir
%g = arith.cmpi eq, (a1 | x), 0
scf.if %g {
  %v = memref.load %m[s * a1]
  ...
}
```

affine-cfg produced

```mlir
%r:2 = affine.if (y <= 1) {
  %v = affine.load %m[y * s]          // unconditional now
  affine.yield %y, %v
} else {
  %v = affine.load %m[(y - 2) * s]    // unconditional now
  affine.yield %y2, %v
}
scf.if ((%r#0 | x) == 0) { ... }      // guard only guards the *use*
```

## How it manifested

MFEM's `SmemPADiffusionApplyTetrahedron<2,2>`: the guard *is* the bounds check — `forward_map2d` has `(D1D-1)² = 1` element and the source reads it only under `a1 < D1D-1`. The lifted load evaluated `forward_map2d[1]` on excluded threads (compute-sanitizer: 4-byte read one element past a 4-byte allocation, 100 errors), and the poisoned value indexed further loads → sticky illegal-address abort in `PA Simplices`. Vanilla clang under the sanitizer: zero errors. Bisected by printing after every pass in the raising pipeline: guard intact after pass 34, gone after the second `affine-cfg` run.

## The fix

Before any mutation, require that the load sits in the conditional's own block with only memory-effect-free operations between them. Then reaching the conditional already implied reaching the load, and the lift changes nothing but where the address is computed. Both match paths (pure `toLift` and the `postOp` rebuild) get the check against the position the load will actually move to.

## Tests

`test/lit_tests/raising/lift-memref-read-guarded.mlir`:
- `@guarded` — the reproducer above; the select still becomes a value-only `affine.if` and the load **stays** under its `scf.if`;
- `@unguarded` — same shape minus the guard; the lift **still fires**, loads in both branches.

## Measured

`PA Simplices` goes from a sticky-crash (7/8 asserts, rc=134) to **all 20 assertions passing**. Full-suite run with this fix is in flight; several other value-failure cases (Hcurl/Hdiv family, PA Markers, L2 Assembly) are plausible victims of the same lift and will be reported in a follow-up comment.


wsmoses:
Reworked to your alternate fix: after the conditional is identified, take `outer = loadOp->getParentOp()` and refuse when (a) `outer` has multiple regions or a multi-block region — a parent that may run its contents conditionally — or (b) `outer` is not an ancestor of the conditional — the load sits under control flow (an if, a loop) the conditional is not under, so moving it beside the conditional would stop that control flow from governing it. Both checks run before any mutation; the `executesWithAndAt` walk from the previous revision is gone.

Verified: the three lit cases pass (@guarded refuses via (a), a new @looped case refuses via (b) — load inside an `scf.for` with the conditional outside — and @unguarded still lifts into both branches), and end-to-end MFEM `PA Simplices` still passes all 20 assertions after rebuilding the simplex TUs.

One property of this variant worth being aware of: it permits side-effecting ops between the conditional and the load (the previous revision refused them), so the load can move backward past an intervening store; fine if these index maps are effectively read-only at this point in the pipeline, which I believe is the intent.

wsmoses:
The CI lit failure was real: `raising_northsouth2.mlir` and `northsouth3.mlir` regressed (my local run had only exercised the new test file, not the suite — the full suite run caught it). Those tests lift a load whose index chain crosses a loop: there the pattern takes the impure-slice path and rebuilds the conditional at **postOp, inside the load's own loop** — the load never leaves the loop, which is legal and must keep working.

So the ancestor check now targets the position the load actually moves to: the conditional on the pure path, `postOp` on the impure path (the region/block check on the load's parent is unchanged and stays up front). northsouth2/3 pass again, the three cases in the new test still behave (guarded refused via the region check, looped refused via the ancestor check, unguarded lifts), the full lit suite is green (1163/1164; the one failure is a stale untracked file on my machine, not in the PR), and MFEM `PA Simplices` still passes all 20 assertions end-to-end.

wsmoses:
Generalized per review: the load's parent walk now steps outward through loops that provably execute at least one iteration (all constant bounds, positive step, lb < ub for every dim, via `LoopLikeOpInterface`) before the region/block and ancestor checks apply — inside such a loop, reaching the parent still means reaching the load, and since the lifted index cannot depend on the loop IV (the existing block-argument dominance checks refuse that), lifting the loop-invariant load out is sound.

New lit case `@looped_const`: same shape as `@looped` but with constant loop bounds — now lifts into both branches; `@looped` (runtime bound, not provably nonempty) still refuses. Full lit suite green, MFEM `PA Simplices` still 20/20 end-to-end.

wsmoses:
CI build failure was the layering check: the new includes (`mlir/Interfaces/LoopLikeInterface.h`, `mlir/Dialect/Utils/StaticValueUtils.h`) need their exporting modules declared. Added `@llvm-project//mlir:LoopLikeInterface` and `:DialectUtils` to `XLADerivatives`' deps; verified locally with `--features=layering_check` and the lit tests re-pass.

PR #2882, opened 2026-08-13
Improve scf.for to stablehlo.while raising

Pangoraw:
we can align and broadcast up to a permutation if there is a mismatch between an iter arg and its yielded value.


wsmoses (approved):

wsmoses:
looks good, but we should try to fix the red ci [which all appear unrelated]

Pangoraw:
Reactant integration is failing on `Error in testset for: binomial checkpointing autodiff matches no checkpointing:`, which i would expect without https://github.com/EnzymeAD/Reactant.jl/pull/3180.

wsmoses:
yeah fair, want to merge this and jll/get the reactant side fix in?

PR #2884, opened 2026-08-13
Serialize barrier-carrying parallels by distributing at the barrier

wsmoses:
Part 2 of 2 fixing the MFEM Hcurl/Hdiv wrong-values family (with #2883; both are needed).

## The bug

A parallel op left inside a `gpu.launch` after mapping (e.g. the per-element inner loops of a thread-per-element kernel) runs within a single thread. A barrier inside it — placed by the gpu1 normalization that merges sibling parallels — synchronizes that parallel's own iterations, not the launch's threads. But `ParallelToGPULaunch` blanket-converted **every** `enzymexla.barrier` in the launch to `gpu.barrier`, and `parallel-serialization` then turned the parallel into a plain `scf.for` with the `gpu.barrier` left inline. Result:

- thread-zero-guarded ops ran on the first iteration, before the remaining iterations computed the values they read (wrong values), and
- the barrier became a block-wide sync inside divergent control flow (deadlock hazard).

## The fix

1. `ParallelToGPULaunch`: only convert barriers **not** nested in a remaining `scf.parallel`; those belong to their parallel and keep their operands.
2. Pipeline: run `cpuify{method=distribute}` (the existing Polygeist barrier-aware machinery) before `parallel-serialization`, so barrier-carrying loops are distributed at their barriers: every iteration of the segment before the barrier completes before the segment after it starts. Applied to both the native and xla paths.

## Verified

- New lit test `serialize-inner-barrier.mlir`: gpu1 keeps `enzymexla.barrier` inside the unmapped inner parallel (and emits no gpu.barrier); gpu1+cpuify+serialization yields the fissioned loops with the guarded sum after the store loop.
- Full lit suite green (including the existing cpuify suite).
- The standalone `PAHcurlHdivMassApply2D` reproducer, wrong since it first compiled, now matches vanilla exactly (checksum 761.6123046875, all elements). Full MFEM suite re-run in progress; will report on the PR.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1914,7 +1914,15 @@ struct ParallelToGPULaunch : public OpRewritePattern<enzymexla::GPUWrapperOp> {
     }
     rewriter.eraseOp(yieldOp);
 
+    // A barrier nested in a parallel op that remains after mapping
+    // synchronizes that parallel's iterations, not the launch's threads; it
+    // is serialized together with its parallel by distributing the loop at
+    // the barrier, which needs the enzymexla barrier and its operands.
     launchBlock->walk([&](mlir::enzymexla::BarrierOp op) {
+      for (Operation *parent = op->getParentOp(); parent != launchOp;
+           parent = parent->getParentOp())
+        if (isa<scf::ParallelOp>(parent))
+          return;
       rewriter.setInsertionPoint(op);
       rewriter.replaceOpWithNewOp<gpu::BarrierOp>(op);
     });
--- a/src/enzyme_ad/jax/raise.cpp
+++ b/src/enzyme_ad/jax/raise.cpp
@@ -154,7 +154,9 @@ extern "C" std::string runLLVMToMLIRRoundTrip(std::string input,
       if (getenv("REACTANT_OMP")) {
         pass_pipeline += ",convert-scf-to-openmp,";
       } else {
-        pass_pipeline += ",parallel-serialization,";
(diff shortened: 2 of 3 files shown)


wsmoses:
Full MFEM campaign results (clean rebuild of all 433 TUs + per-case GPU suite, with this PR + #2880 + #2883 + #2885 applied):

- Build: 0 failures (the 33-TU tmop compile regression the first campaign hit was #2872's unresolved-kernel path, fixed by #2885).
- vs the last valid pre-fix campaign: **L2 Assembly Levels, PA Markers, PA Simplices now pass**; the Hcurl/Hdiv family dropped from ~47 failing assertions to ~18; PA Gradient 6→1.
- Remaining known classes are pre-existing (verified: merged-main is *worse* on the reduced Hcurl reproducer, 6 bad configs vs 4): the Hcurl/Hdiv output-section miscompile (isolated to a 2-min reproducer, being chased next), Scan ×2 (stub-binding class), Batched LinAlg 1, QuadratureInterpolator 1 (register pressure).

Also filed while chasing the remainder: #2886 — every runtime-grid `forall` kernel was falling into the one-thread-per-block fallback because the `dim3(x,1)` pack's high-half read never folded to 1.

wsmoses:
With #2894 (ParallelizeBlockOps declines parallels that will be serialized in a thread), the primary path no longer creates barriers inside serialized parallels, and the full MFEM CUDA suite gives identical results without this PR (same 8 residual failing suites as the with-#2884 baseline; details on #2894). Its pieces remain sound as defense-in-depth -- the barrier-walk scoping is correct hygiene and cpuify{method=distribute} protects fallback paths where a wrapper fails conversion and mapped-level parallels get serialized wholesale -- so keeping or closing is a judgement call rather than a correctness need.

PR #2885, opened 2026-08-13
Drop kernels that were never resolved to a device body

wsmoses:
Fixes a compile regression from #2872 on 33 MFEM tmop TUs (e.g. `fem/tmop/metrics/302.cpp`).

## The regression

A `gpu_wrapper` whose kernel launch was never resolved to a device body (the device instantiation is absent from the `.re_export`) holds only a call to an external `reactant$__device_stub__*` declaration, and carries the **host** function's target attributes (`target_cpu = "x86-64"`).

Before #2872 these wrappers failed every block-size split *and* the 6-bound equality, so they took the loud-error drop path and the TU compiled (the kernel is a genuinely never-launched instantiation). #2872's shape matching now succeeds on them — trivially so: the parallel's bounds and the wrapper's launch bounds were built from the same launch configuration, so they always match. The "kernel" is then emitted, the outlined `gpu.module` inherits the host chip, and compilation dies with:

```
ptxas fatal   : Value 'x86-64' is not defined for option 'gpu-name'
fatal error: error in backend: 64-bit code requested on a subtarget that doesn't support it!
```

Bisected across saved libRaise builds: main-before-#2872 compiles `302.cpp`, main+#2872 fails, all later changes irrelevant.

## The fix

In `SplitParallelOp`, before looking for any split: if the wrapper's `target_cpu` is neither `sm_*` nor `gfx*`, the kernel cannot be compiled for the GPU under any block size — emit the error and take the existing drop path (`CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES` at the launch site), same as the pre-#2872 outcome but with a diagnostic naming the real cause.

## Verified

- New lit test `drop-unresolved-kernel.mlir` (exact-shape-matching wrapper with host target and external stub call → error + constant, no launch).
- `302.cpp` compiles again (rc=0, three "not resolved to a device body" diagnostics); resolved-kernel path unaffected (Hcurl reproducer still exact-matches vanilla).
- Full lit suite green.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -526,6 +526,24 @@ struct SplitParallelOp : public OpRewritePattern<enzymexla::GPUWrapperOp> {
 
     auto loc = pop->getLoc();
 
+    // A wrapper still carrying a host target holds a kernel that was never
+    // resolved to a device body: its content is a call to an external device
+    // stub, and the outlined module would inherit the host chip. No block
+    // size makes that compilable, including an exact shape match.
+    if (auto arch =
+            dyn_cast_or_null<StringAttr>(wrapper->getAttr("target_cpu")))
+      if (!arch.getValue().starts_with("sm_") &&
+          !arch.getValue().starts_with("gfx")) {
+        wrapper.emitError()
+            << "kernel with host target '" << arch.getValue()
+            << "' was not resolved to a device body; the kernel is dropped";
+        rewriter.setInsertionPoint(wrapper);
+        auto err = arith::ConstantIndexOp::create(
+            rewriter, loc, CUDA_ERROR_LAUNCH_OUT_OF_RESOURCES);
+        rewriter.replaceOp(wrapper, err->getResults());
+        return success();
+      }
+
     int curRegion = 0;
     llvm::SmallSet<int, 6> emittedBlockSizes;
     std::vector<Attribute> descs;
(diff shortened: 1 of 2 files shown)


wsmoses:
This PR turns out to fix a hard compile regression on current main: 33 MFEM TMOP TUs (fem/tmop/**) fail to build through the raising plugin at HEAD. Those TUs contain template instantiations (e.g. TMOP_AssembleDiagPA_2D<24,24,0,0> via CuWrap2D) whose device bodies are never resolved, so launch recognition wraps the host device-stub; kernel formation then fails for every block size ("Failed to make kernel with exact dimension" / the #2871 diagnostic), and the surviving stub gpu.module is serialized with a target attribute built from the host function's attrs — ptxas gets `-arch x86-64` and the compile dies with "Value 'x86-64' is not defined for option 'gpu-name'".

Verified: bare main (35a49050d) fails diag2.cpp with exactly this signature (as do #2880/#2883-era commits — the earlier full-suite campaigns all carried this branch's changes as working-tree edits, which is why it went unnoticed); main + this PR compiles it cleanly, with the unresolved kernels dropped before serialization. A separate small PR adds a belt-and-braces guard so a non-`sm_*` chip can never reach the NVVM target attribute even when a stub does survive.

wsmoses:
Superseded at the root by EnzymeAD/Reactant#81 (uniform stub binding in runtime calls) + #2912 (recognition refuses bodyless declarations). Validated without this PR on a full clean MFEM rebuild: all 433 objects compile (including the 33 TMOP TUs this PR was fixing), GPU suite 74/74, benchmark identical to the prior baseline (hex deriv global 0.98x, quad deriv global 0.93x vs vanilla, correct numerics). Closing in favor of the root-cause pair.

PR #2892, opened 2026-08-14
Drop nneg when arith.extui's fold merges extension chains

wsmoses:
This is the actual root cause of the long-standing MFEM Hcurl/Hdiv wrong-values family (`Hcurl/Hdiv PA Coefficient` et al., always `trialHcurl=true` only): an **upstream MLIR miscompile in `arith::ExtUIOp::fold`**, patched here against the pinned LLVM until it lands upstream.

## The bug

`ExtUIOp::fold` merges `extui(extui(x))` by reassigning the outer op's source — keeping the outer op's `nneg` flag:

```cpp
if (auto lhs = getIn().getDefiningOp<ExtUIOp>()) {
  getInMutable().assign(lhs.getIn());
  return getResult();
}
```

`nneg` asserts the *source* is non-negative **as a signed value**. The merge changes the source, about which the outer flag says nothing. The killer instance, produced naturally when `polygeist-mem2reg` forwards a raised bool-byte store to its load:

```mlir
%w = arith.extui %b : i1 to i8          // bool byte
%r = arith.extui %w nneg : i8 to i32    // valid: the i8 is 0 or 1
```

folds to `arith.extui %b nneg : i1 to i32` — but signed i1 `1` **is −1**, so this is poison whenever the bool is **true**. After translation, LLVM's host -O3 legally folds the bool to false everywhere the merged value flows. In MFEM this hit the kernel-launch argument computations for `PAHcurlHdivMassApply2D`: every `trialHcurl`-derived loop bound and offset passed to the kernel took its `false` arm, scrambling `y` indices (and reading one element past a basis array — the sanitizer OOB that started this hunt). `trialHcurl=false` was correct by construction, which is why only half the family ever failed, and why the failure survived every pipeline-level bisection: the MLIR was correct at every stage; the poison only cashed out in LLVM host codegen.

## The fix

Keep `nneg` on the merged extension only if the **inner** extension carries it (its assertion is about the surviving source). Applied to the pinned LLVM via the existing `patch_cmds` mechanism; needs the same one-line fix upstream (`mlir/lib/Dialect/Arith/IR/ArithOps.cpp`).

## Verified

- New lit test `extui-nneg-merge.mlir`: the i1 chain merges **without** nneg; an all-nneg chain keeps it.
- The 50-line standalone reproducer (separate-TU kernel with two closure-pair selects): 4 wrong configs → 0.
- Real MFEM `Hcurl/Hdiv PA Coefficient` with the fixed TU: **32/43 → 42/43 assertions** (remaining 1 + an exit-time abort under investigation).
- Full lit suite green.

Diagnosis trail (for the curious): suite → single TU → 50-line separate-compilation reproducer (single-TU builds fold the bool and mask the bug) → intermediates instrumented on-device (`wy`, `massX` correct; store indices wrong; loop bounds taking the `th=0` arm) → `LD_PRELOAD` interception of `cudaLaunchKernel` showing a wrong host-computed argument → x86 disassembly showing `zext(th)` compiled to the literal `0` → `zext nneg i1` in the final IR → the fold.

### DIFF
--- a/test/lit_tests/extui-nneg-merge.mlir
+++ b/test/lit_tests/extui-nneg-merge.mlir
@@ -1,0 +1,29 @@
+// RUN: enzymexlamlir-opt --canonicalize %s | FileCheck %s
+
+// Merging extui(extui(x)) must not keep the outer op's nneg flag: the flag
+// asserts the SOURCE is non-negative as a signed value, and the merge changes
+// the source. extui nneg i8->i32 over extui i1->i8 is satisfied for every
+// bool byte, but the merged extui nneg i1->i32 is poison whenever the bool
+// is true -- LLVM then folds the bool to false wherever the merged value
+// flows, which is how MFEM's trialHcurl=true kernels lost their loop bounds.
+
+func.func @merge(%b: i1) -> i32 {
+  %w = arith.extui %b : i1 to i8
+  %r = arith.extui %w nneg : i8 to i32
+  return %r : i32
+}
+
+// CHECK-LABEL: func.func @merge(
+// CHECK-NOT: nneg
+// CHECK: arith.extui %{{.*}} : i1 to i32
+// CHECK-NOT: nneg
+
+// When the inner extension itself carries nneg, the merged one may keep it.
+func.func @keep(%v: i8) -> i64 {
+  %w = arith.extui %v nneg : i8 to i32
+  %r = arith.extui %w nneg : i32 to i64
+  return %r : i64
+}
+
+// CHECK-LABEL: func.func @keep(
+// CHECK: arith.extui %{{.*}} nneg : i8 to i64
--- a/workspace.bzl
+++ b/workspace.bzl
@@ -202,6 +202,14 @@ sed -i.bak0 "s/cupti_driver_cbid/cupti/g" xla/backends/profiler/gpu/cupti_tracer
 sed -i.bak0 "s/patch_cmds = \\[/patch_cmds = \\[\\\"find . -type f -name config.bzl -exec sed -i.bak0 's\\/HAVE_LINK_H=1\\/HAVE_LINK_H=0\\/g' {} +\\\",/g" third_party/llvm/workspace.bzl
 """,
     """
+# arith.extui's fold merges extui(extui(x)) by swapping the outer op's source
+# while keeping its nneg flag. The flag asserts the source is non-negative as
+# a signed value, about which the outer flag says nothing once the source
+# changes: extui nneg i8->i32 over extui i1->i8 becomes extui nneg i1->i32,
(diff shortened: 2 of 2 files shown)


wsmoses:
Full campaign with this PR (+ #2880/#2883/#2884/#2885/#2886): clean rebuild of all 433 TUs, **0 build failures**, per-case GPU suite:

| suite | before session | now |
|---|---|---|
| Hcurl/Hdiv PA Coefficient | 33 failed | **1 failed** (41/42) |
| Hcurl/Hdiv Mixed PA Coefficient | all failing | **1 failed** (48/49) |
| 3D Bilinear VectorFE PA | 4 failed | 2 failed |
| Hcurl/Hdiv diagonal PA | 1 failed | 1 failed |
| L2 Assembly / PA Gradient / PA Markers / H(div) Element Assembly / Vector Diffusion Diagonal / PA Simplices | failing | **all pass** |
| Inclusive+Exclusive Scan | 2 failed | 2 failed (pre-existing stub-binding class) |
| Batched Linear Algebra | 1 failed | 1 failed (pre-existing) |
| QuadratureInterpolator | 1 failed | 1 failed (pre-existing register-pressure class) |

58 of 66 suites fully green vs vanilla clang; the remaining 9 assertions are being chased separately.

wsmoses:
End-to-end benchmark now that the suite is near-green: `bench_dfem_linearized_simple -d cuda` runs correctly through the raising path for the first time (the long-standing illegal memory access in the primal path was this same poison bug). Best of 3 vs vanilla clang+ClangEnzyme, same machine/day (RTX 5090):

| metric | vanilla | Reactant | ratio |
|---|---|---|---|
| quad forward global | 3.114e-05 | 3.120e-05 | 1.00× |
| quad derivative global | 5.504e-05 | 5.197e-05 | **0.94×** |
| quad forward local | 1.887e-05 | 2.085e-05 | 1.10× |
| quad derivative local | 2.506e-05 | 2.502e-05 | 1.00× |
| hex forward global | 4.577e-05 | 4.772e-05 | 1.04× |
| hex derivative global | 8.043e-05 | 7.859e-05 | **0.98×** |
| hex forward local | 3.116e-05 | 3.127e-05 | 1.00× |
| hex derivative local | 3.726e-05 | 4.137e-05 | 1.11× |

`linf_error` ≤ 1e-15 everywhere. Overall parity; derivative-global is slightly ahead of vanilla, and the two ~10%-behind local paths are the next perf targets.

wsmoses:
Filed upstream as llvm/llvm-project#216315 (fix verified against upstream main: full `mlir/test/Dialect/Arith` passes with the new cases). Once it lands and the LLVM pin advances past it, the `patch_cmds` entry here can be dropped.

PR #2894, opened 2026-08-14
Decline ParallelizeBlockOps for parallels serialized in a thread

wsmoses:
Only the grid parallel (directly in the wrapper) and the block parallel (directly in the grid parallel) map onto the gpu launch. A parallel nested any deeper -- under an if, a loop, or a third parallel level -- runs serialized inside a single thread. These arise when the raiser auto-deduces per-thread loops with runtime trip counts as parallel (the MFEM Apply kernels' local-array loops under the divergent \`if (k >= N) return\` guard are all of this shape).

ParallelizeBlockOps used to fire on them too, moving the neighbouring sequential code into the parallel behind a first-iteration guard and a barrier. That is pure churn for a parallel that is about to be serialized -- the surrounding code already runs in the right order there -- and it is what manufactured the barrier-inside-a-serialized-loop shapes that #2884 has to distribute around (and, before #2883/#2884, the Hcurl value-bug family, e.g. the v1.cu reproducer printing 6 7 8 instead of 10 11 12).

With the pattern declined at those depths:
- the end-to-end reproducers (v1, mwe_hcurl, m_cut5) pass even with #2884's cpuify{method=distribute} removed; #2884 remains useful only as defense-in-depth for wrappers that fail conversion entirely,
- kernels whose bodies contain such inner parallels no longer fall back to \`block_size=1\` launches: the multi-parallel test now gets a proper 2x128 split instead of grid=256/block=1, since no barrier lands inside the residual parallels,
- the parallelize-multi-parallel test is updated from "each application owns its segment" to "nested siblings are declined and keep their original order".

Stacked on #2880 (base pb/parallelize-store-order) since the test it updates lives there.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1075,6 +1075,11 @@ struct ParallelizeBlockOps : public OpRewritePattern<scf::ParallelOp> {
       LLVM_DEBUG(DBGS() << "cannot parallelize around a reduction\n");
       return failure();
     }
+    auto parentPar = dyn_cast<scf::ParallelOp>(pop->getParentOp());
+    if (!parentPar || !isa<enzymexla::GPUWrapperOp>(parentPar->getParentOp())) {
+      LLVM_DEBUG(DBGS() << "parallel is serialized in a thread\n");
+      return failure();
+    }
     auto loc = pop->getLoc();
     Block *outerBlock = pop->getBlock();
     Block *innerBlock = pop.getBody();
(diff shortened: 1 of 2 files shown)


wsmoses:
End-to-end confirmation on the MFEM CUDA suite (clean rebuild of build_cuda_mlir + full gpu_unit_tests) with a library built as main (incl. #2880) + this PR + #2885, and **#2884 fully absent** (stock parallel-serialization, unconditional barrier walk):

- 0 compile failures; full lit suite 1171/1171; v1.cu / mwe_hcurl.cu (checksum 761.6123046875) / m_cut5.cu all exactly match vanilla clang.
- The failing-suite set is identical to the with-#2884 baseline (campaign8): the 8 known residual suites (Scan x2 stub class, Batched LinAlg, Hcurl/Hdiv PA Coeff family, 3D VectorFE, diagonal PA, QuadInterp), none barrier-related.
- Two of those improved from SIGABRT (rc=134) to clean assertion failures: Hcurl/Hdiv PA Coefficient 134 -> 2 and 3D Bilinear VectorFE PA 134 -> 4.

So with this guard, no barrier is ever manufactured inside a to-be-serialized parallel, and #2884's distribute-before-serialize becomes defense-in-depth for wrappers that fail conversion entirely rather than a correctness requirement.

PR #2895, opened 2026-08-15
Bound kernels with runtime block sizes at the architecture maximum

wsmoses:
A kernel outlined from a launch with runtime block sizes gets no launch bound: outlining only infers \`known_block_size\` (hence \`.maxntid\`) from constant dims, and \`AddLaunchBounds\` -- gated behind \`emitGPUKernelLaunchBounds\`, which the pipeline never enabled -- also bailed on non-constant dims. ptxas is then free to allocate registers as if the block were small, and a larger legal launch fails at runtime with \`too many resources requested for launch\`.

Concretely, MFEM's dynamic \`Det3D<0,0,*>\` quadrature-interpolator kernels compile to 134 registers on sm_120 (the vanilla clang build of the same source sits at 55); any launch above 489 threads is unlaunchable, and the suite launches (8,8,8)=512 -- QuadratureInterpolator dies with error 701 after 848 assertions.

The fix enables the launch-bounds pattern in the pipeline and gives it an unknown-dims branch: cap at the architecture maximum (1024), which is exactly the CUDA contract for a kernel without \`__launch_bounds__\` -- every legal launch must fit. Verified against the driver that a 3D (8,8,8) launch is accepted against \`.maxntid 1024, 1, 1\` (clang's own \`__launch_bounds__\` lowering shape). The constant-dims branch is modernized from the inert integer \`nvvm.maxntidx\` spelling to the \`nvvm.maxntid\` array attribute that convert-polygeist-to-llvm and the NVVM translation actually consume.

With the cap the Det3D kernel compiles to 64 registers and the full QuadratureInterpolator suite passes (1530/1530 assertions, previously SIGABRT). Full lit suite green (1172). The kernel does now spill (STACK:248) because ptxas still expands the code ~6x relative to clang before fitting the budget; that expansion is the register-pressure gap tracked separately, and the cap turns it from a hard launch failure into a perf question.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -263,56 +263,92 @@ struct AddLaunchBounds : public OpRewritePattern<gpu::LaunchFuncOp> {
     // gpu::LaunchFuncOp's)
     auto gpuFuncOp = launchOp->getParentOfType<ModuleOp>().lookupSymbol(
         launchOp.getKernel());
+
+    // Launch bounds the kernel was compiled with: clang encodes
+    // __launch_bounds__(maxThreads, minBlocks) as nvvm annotations, which the
+    // importer carries in the device function's passthrough and launch
+    // recognition copies onto the error op.
+    int32_t srcMaxntid = 0, srcMinctasm = 0;
+    if (auto err = launchOp->getParentOfType<enzymexla::GPUErrorOp>()) {
+      if (auto attr =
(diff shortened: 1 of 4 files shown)


wsmoses:
Amended: kernels that already carry known_block_size are now skipped -- the lowering derives nvvm.maxntid from that attribute, and setting it here as well produced a duplicate nvvm.maxntid entry in the function's attribute dictionary (surfaced as unparseable IR text when round-tripping the module).

wsmoses:
Superseded by #2918, which carries this commit plus the launch-bounds preservation (#2904) as one unit based directly on main — #2904 had been merged into this branch rather than main.

PR #2896, opened 2026-08-15
Translate flowing kernel pointers to their registered stubs

wsmoses:
CUB's device dispatch reads the kernel's occupancy before launching: \`cudaOccupancyMaxActiveBlocksPerMultiprocessorWithFlags(&occ, scan_kernel, ...)\` where \`scan_kernel\` is a function-pointer value selected at runtime from the tuning policy. Reactant's fixup rewrites the kernel-pointer argument of such runtime queries when it is a direct function reference, but a flowing value only gets wrapped in \`__reactant$get_device_from_host\` -- which this pass then lowered as identity. The query therefore received the original host stub, whose cuda-runtime registration the plugin removed, and failed with \`invalid resource handle\` (400); CUB bails before ever launching. This is what killed both Scan suites (Inclusive/Exclusive, SIGABRT) in the MFEM unit tests, reduced to a 25-line \`cub::DeviceScan::InclusiveSum\` reproducer.

The registration this pass emits (one \`__polygeist_*_device_stub\` per raised kernel, registered under the kernel's name) is exactly what the query needs -- the pointer just never reaches it. The fix records, on each registration stub, which host function its kernel was raised from (kernel name minus the \`_kernel\` suffix, or minus the \`reactant$\` prefix, when that symbol exists), and lowers \`__reactant$get_device_from_host\` as a compare/select chain over those pairs instead of identity: a known host stub translates to its registered stub, unknown pointers pass through unchanged. When there are no raised kernels or no callers the old identity erase is kept, so nothing changes for modules that never take a kernel's address.

Verified: the cub reproducer goes from \`invalid resource handle: 0 0 0 0 0 0 0 0\` to matching vanilla (\`1 3 6 10 15 21 28 36\`); both MFEM Scan suites pass end-to-end (40/40 assertions each, previously SIGABRT); full lit suite green with the new \`host-stub-translate.mlir\` covering the emitted chain and pass-through call.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp
@@ -2450,6 +2450,20 @@ ConvertGPUModuleOp::matchAndRewrite(gpu::GPUModuleOp kernelModule,
             rewriter.setInsertionPointToEnd(stub.addEntryBlock(rewriter));
             LLVM::ReturnOp::create(rewriter, ctorloc, ValueRange());
           }
+          // Remember which host function this kernel was raised from, so
+          // that a kernel pointer flowing into a cuda runtime query can be
+          // translated to this registered stub.
+          {
+            StringRef origName = f.getName();
+            if (origName.starts_with("reactant$"))
+              origName = origName.drop_front(strlen("reactant$"));
+            else if (origName.ends_with("_kernel"))
+              origName = origName.drop_back(strlen("_kernel"));
+            if (moduleOp.lookupSymbol<LLVM::LLVMFuncOp>(origName))
+              stub->setAttr(
+                  "polygeist.host_origin",
+                  FlatSymbolRefAttr::get(rewriter.getContext(), origName));
+          }
           auto aoo = LLVM::AddressOfOp::create(ctorBuilder, ctorloc, stub);
           auto bitcast = LLVM::AddrSpaceCastOp::create(ctorBuilder, ctorloc,
                                                        llvmPointerType, aoo);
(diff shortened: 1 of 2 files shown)


wsmoses:
Re-verified against current main (post #81/#2912/#2918) with a runtime MWE — this PR is still required, and the failure it prevents is now cleaner to state.

```cpp
__global__ void __launch_bounds__(64) kern(double *p) { p[threadIdx.x] = 2.0; }

__attribute__((noinline)) int query(const void *f) {   // pointer is a runtime value:
  cudaFuncAttributes a;                                 // the plugin wraps it in
  cudaFuncGetAttributes(&a, f);                         // __reactant$get_device_from_host
  return a.maxThreadsPerBlock;
}
int main() { ...; kern<<<1, 32>>>(d); ...; printf("%d\n", query((const void*)kern)); }
```

Three builds of the same file:

| stack | result |
|---|---|
| bare main (identity lowering of `get_device_from_host`) | `cudaFuncGetAttributes` fails with **invalid resource handle** — the original host address is not a registered handle in the raised binary |
| main + #2857 (identity, clang stub registered) | query "succeeds" but returns **64 — the original clang kernel's bound**, not the raised kernel's |
| main + this PR's mapping (+ stub family) | returns **32 — the raised kernel's actual bound** (its exact-dims maxntid) |

The middle row is the subtle one: with identity the query resolves against the *original* kernel's registration while the *raised* kernel is what launches. Any occupancy-driven code (CUB's scan: query attributes/occupancy, derive a launch config, launch) then configures launches from the wrong kernel's limits — the Scan-suite SIGABRT class this PR fixed. The host_origin map is what redirects a flowing original-host pointer to the stub registered for the kernel that will actually run.

wsmoses:
Superseded by the attribute-based registration binding (EnzymeAD/Reactant#84 + #2857): with kernels registered under the host symbol the program actually passes around, the identity lowering of `__reactant$get_device_from_host` is semantically correct and the runtime translation map is no longer needed. Validated without this PR on a full clean MFEM rebuild: 74/74 GPU suite — Inclusive/Exclusive Scan included — correct numerics, benchmark at the established profile.

PR #2898, opened 2026-08-15
Only drop an if constraint subsumed by a parent that implies it

wsmoses:
AffineIfSimplification removes a constraint from an \`affine.if\`'s set when the same constraint appears in an enclosing \`affine.if\` with the same operands. The comparison only looked at the constraint expression, not whether each side is an equality or an inequality: nested inside a guard with \`s0 - 1 >= 0\`, an inner \`affine.if\` on \`s0 - 1 == 0\` had its only constraint dropped as redundant, the set became vacuously true, and the conditional was replaced by its (empty) then-region.

End-to-end this deleted the entire back-substitution remainder of MFEM's batched \`LUSolve\` device kernel (clang rotates the \`for (j = m-1; j >= 0; j--)\` loop into a peeled first iteration guarded by \`m >= 1\` plus an \`if (m != 1)\` remainder -- exactly the eq-inside-ineq nesting): every batched solve returned its right-hand side unchanged. The "Batched Linear Algebra" suite failure, and one "Hcurl/Hdiv diagonal PA" failure with the same shape in the curl-curl kernels, trace here; found by object-level delta-debugging to \`linalg/batched/native.cpp.o\`, a standalone LUFactor/LUSolve driver (factors bit-identical, solve a no-op), and instrumenting the pass to log which constraint removal fired (\`CANREMOVE cst=s0 - 1 myEq=1 parent: ()[s0] : (s0 - 1 >= 0)\`).

The fix requires implication: an equality parent subsumes either form; an inequality parent only subsumes the same inequality.

Verified: the LUSolve driver goes from solution == RHS to matching the host reference to 5.6e-17; "Batched Linear Algebra" passes 216/216 and "Hcurl/Hdiv diagonal PA" 152/152 end-to-end after rebuilding the affected TUs; full lit suite green with the new \`affine-if-eq-in-ineq.mlir\` covering the nesting directly.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -3218,10 +3218,16 @@ struct AffineIfSimplification : public OpRewritePattern<affine::AffineIfOp> {
         bool canRemove = false;
         for (auto paren = op->getParentOfType<affine::AffineIfOp>(); paren;
              paren = paren->getParentOfType<affine::AffineIfOp>()) {
-          for (auto cst2 : paren.getIntegerSet().getConstraints()) {
+          for (auto cst2 :
+               llvm::enumerate(paren.getIntegerSet().getConstraints())) {
             if (paren.getElseRegion().isAncestor(op->getParentRegion()))
               continue;
-            if (cst2 == cst.value() &&
+            // The parent constraint only subsumes this one if it implies it:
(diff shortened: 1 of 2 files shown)


wsmoses:
Added the within-set case: an inequality is dropped when the same expression appears as an equality in the same set (a >= 0 && a == 0 is a == 0), with a test (@g) covering it.

PR #2900, opened 2026-08-15
Cast booleans to index unsigned when legalizing affine operands

wsmoses:
When affine legalization needs an index-typed symbol from an integer value -- composing map operands in \`fully2ComposeAffineMapAndOperands\`, or materializing the integer-set operands of a converted conditional -- it emitted a signed \`arith.index_cast\`. The operand walker strips zero-extensions to reach the underlying value, so a boolean condition reaches the cast as a raw \`i1\`, and the signed cast sends \`true\` to \`-1\`: an \`affine.if\` built as \`s0 - 1 == 0\` over that symbol evaluates false exactly when the original condition held.

End-to-end this broke MFEM's \`PAHcurlL2Apply2D\` (MixedScalarCurlIntegrator, the "Hcurl/Hdiv Mixed PA Coefficient" HcurlH1_2D case): the kernel's \`H1 = (D1Dtest == D1D)\` flag reached the launch as \`-1\`, so \`select(H1 == 1, Bt, Bot)\` picked the L2 basis and the companion stride select picked \`D1D-1\`, for an H1 test space. Found by delta-debugging to \`bilininteg_hcurl_kernels.cpp.o\`, a one-hot operator-matrix probe showing Bot-scale products where Bt belonged, and an \`LD_PRELOAD\` dump of the 24 launch parameters showing \`a17 = 0xffffffffffffffff\` where the H1 flag (1) was expected.

The fix routes these casts through a helper that uses \`arith.index_castui\` for \`i1\` sources and keeps the signed cast otherwise.

Verified: the standalone kernel driver goes from 4/4 wrong configurations to 0; the "Hcurl/Hdiv Mixed PA Coefficient" suite passes 180/180 end-to-end after rebuilding the TU -- the last failing suite of the MFEM GPU unit-test campaign; full lit suite green with the new \`affine-i1-symbol-cast.mlir\` (without the fix its \`affine.if\` symbol is produced by a signed cast of the compare).

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -1013,6 +1013,14 @@ bool need(IntegerSet *map, SmallVectorImpl<Value> *operands, Region *scope) {
 // composeAffineMapAndOperands.  A caller passing no builder is asking a
 // question, since without one the normalizer cannot move an operand to make
 // it legal -- and the attribute keeps the answer from being dropped.
+// A boolean is the unsigned value 0 or 1; the signed cast would send true to
+// -1.
+static Value castToIndex(OpBuilder &b, Location loc, Value v) {
+  if (v.getType().isInteger(1))
+    return arith::IndexCastUIOp::create(b, loc, b.getIndexType(), v);
+  return arith::IndexCastOp::create(b, loc, b.getIndexType(), v);
+}
+
 [[nodiscard]] static bool fully2ComposeAffineMapAndOperands(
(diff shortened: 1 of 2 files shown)


wsmoses:
Final end-to-end confirmation: a clean rebuild of MFEM build_cuda_mlir plus the full gpu_unit_tests sweep with this fix stacked on #2895/#2896/#2897/#2898 completes with 0 compile failures and **74/74 suites passing** -- the first fully green run of the MFEM GPU unit-test campaign under the raising path (baseline before this PR series: 8 failing suites).

PR #2901, opened 2026-08-15
Add reverse-mode autodiff rule for `stablehlo.custom_call`

Janssena:
`stablehlo.custom_call` has no `ReverseAutoDiffOpInterface`, so any program containing one fails
reverse-mode AD with `could not compute the adjoint for this operation`. A custom call is opaque by
construction, so the only thing that can supply its adjoint is whoever built it.

This PR adds `AutoDiffCustomCallRev`, which lets a call site carry its own reverse rule as attributes
pointing at an ordinary `func.func`.

```mlir
%o, %lse = stablehlo.custom_call @foo(%q, %k, %v, %mask) {
    enzyme.reverse         = @foo_reverse,
    enzyme.active_operands = array<i64: 0, 1, 2>
} : (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>, tensor<8xi1>)
 -> (tensor<8xf32>, tensor<f32>)

func.func @foo_reverse(%q, %k, %v, %mask,   // every primal operand
                       %o, %lse,            // every primal result
                       %do, %dlse)          // the cotangent of every result
    -> (tensor<8xf32>, tensor<8xf32>, tensor<8xf32>)   // one per active operand, in order
```

## Motivation

In many cases, highly efficient and validated GPU kernels already exist (for example those written in `triton`) that users might prefer over XLA generated from tracing programs using JAX or Reactant.jl. Directly embedding these in the lowered MLIR results in a `stablehlo.custom_call`, which Enzyme cannot differentiate. In many cases, the user can also provide a kernel or python / julia for the *reverse*, which can appropriately be added to the MLIR, which is what this PR will allow the user to do.

## Why the rule lives on the call site

Not a registry keyed on `call_target_name`: one target name routinely dispatches many unrelated
kernels through its `backend_config`. The motivating case is a Metal-backed attention kernel where WIP functionality in the `jax-mps` PJRT-plugin defines a `mps.metal_kernel_lib` which dispatches `_fwd`, `_bwd_q`, `_bwd_kv` and `_bwd_b`: a forward and three backwards, all under one name, sharing no adjoint. Per-call-site attributes are also trivially emitted by any frontend that can build a custom call at all.

## Added properties

* `enzyme.reverse` — `FlatSymbolRefAttr` naming a `func.func` in the same module. It is an ordinary function: pure StableHLO, or containing custom calls of its own. This can be obtained from lowering python / julia code or supplying another `stablehlo.custom_call`.
* `enzyme.active_operands` — operand indices that can carry a gradient; anything omitted is treated as constant, which is what a boolean mask needs. Absent, it defaults to every float- or complex-typed operand. Listing a non-float operand is an error rather than a silent no-op, since an integer cotangent would otherwise accumulate through `arith.addi` and look like a gradient.
* `enzyme.nondiff` — gradient stop; takes precedence over `enzyme.reverse`.
* With none of these, the pre-existing error is emitted unchanged. The rule never silently differentiates a custom call to zero.

Results are deliberately *not* filtered the way operands are: every result contributes a cotangent argument, so the argument list stays positional. A result whose cotangent is genuinely zero gets a materialized zero rather than being dropped.

## Constraints

* `output_operand_aliasing` is refused: an aliased result may no longer hold the value its cotangent is computed from, which would make the cached primal result meaningless.
* Vector reverse mode (`width != 1`) is refused; the signature convention is defined for a single cotangent per result.
* The reverse function's signature is fully verified against the call site before the rule fires. A wrong convention here would produce a plausible-looking wrong gradient, so each mismatch is a diagnostic naming the offending position.

`has_side_effect` is fine and is not refused: the primal is cloned as-is and the caches are pushed around it.

## Implementation notes

* Reverse mode clones the primal, so no `AutoDiffOpInterface` (forward-mode) implementation is needed for the rule to be reachable. Forward mode continues to refuse.
* `cacheValues` caches every primal operand and result, since the convention hands all of them to the reverse function.
* The reverse `func.func` is referenced only from an attribute. MLIR's symbol-use scanning walks op attributes, so it survives symbol DCE with no keep-alive mechanism.

## Tests

Added 14 new lit tests in `test/lit_tests/diffrules/stablehlo/custom_call*.mlir`, all using a trivially known adjoint (a scaling, so the reverse is one `stablehlo.multiply`) and checkable with `enzymexlamlir-opt` alone.

| Test | Covers |
|---|---|
| `custom_call` | multi-result call, `i1` operand excluded, zero cotangent for a dead result |
| `custom_call_accumulate` | cotangents accumulate rather than overwrite, on both the result and the operand side |
| `custom_call_default_active` | the implicit float-operand filter, two returned cotangents |
| `custom_call_while` | call inside `stablehlo.while` — caches really pushed and popped per iteration |
| `custom_call_if` | call inside one branch of `stablehlo.if` |
| `custom_call_multi_site` | two sites sharing a `call_target_name` with different reverse functions |
| `custom_call_side_effect` | `has_side_effect`, and a reverse rule that itself contains chained custom calls |
| `custom_call_nondiff` / `_precedence` | gradient stop, and that it wins over `enzyme.reverse` |
| `custom_call_no_rule` | no attributes ⇒ the original error, unchanged |
| `custom_call_forward` | forward mode still refuses |
| `custom_call_aliasing` | `output_operand_aliasing` refused |
| `custom_call_bad_signature` | arity and result-type mismatches |
| `custom_call_invalid` | ten chunks: unknown symbol, out-of-range index, duplicate index, non-float active operand, both wrong attribute kinds, and the four remaining signature mismatches |

## Testing performed

Built and run on `linux/arm64` in the project's CI container (`ghcr.io/enzymead/reactant-docker-images`), bazel 7.7.0, clang, `-c opt`.

* All 14 `custom_call*` lit tests pass.
* Complete test suite `//test/lit_tests/...` 1180/1180 pass, i.e. no existing diffrule is disturbed by attaching `ReverseAutoDiffOpInterface` to `CustomCallOp`.
* End-to-end against a real frontend: used Reactant.jl to emit the attributes and trace a VJP into the module, and the resulting IR (five operands including an `i1` mask, two results, four active operands) was run through this rule. The resulting gradient matches FiniteDiff.jl and operates correctly: the rule is called with the five primal operands, both primal results, the output cotangent, and a materialized zero for the dead residual; the four returned cotangents accumulate into the right operands; the mask gets none (== Array{Bool}); and the permutations surrounding the call are differentiated by Enzyme itself.

## Known limitations according to Claude

* Split reverse mode (`DerivativeMode::ReverseModeGradient`) is untested. No lit test in this repository exercises it for any rule, so there was nothing to model a test on; `cacheValues` uses the same `initAndPushCache` API as the existing reduce/while rules.
* Vector mode (`width > 1`) is refused, and the refusal path itself has no test — again, no existing reverse-mode test in the repo uses a width above 1.
* Complex-typed operands are accepted by the default active-operand filter but untested.
* `enzyme.reverse` is a flat symbol reference, so the rule must live in the symbol table enclosing the call. A reverse function in a nested module will not resolve.
* Interaction with `while` checkpointing is untested.
* Every primal operand and result is cached unconditionally, because the convention passes all of them. Correct, but for a call inside a loop that is per-iteration memory. If it proves costly, a follow-up attribute could narrow what the rule asks to be cached.

### DIFF
--- a/custom_call_from_reactant.mlir
+++ b/custom_call_from_reactant.mlir
@@ -1,0 +1,23 @@
+module @"reactant_#3" attributes {mhlo.num_partitions = 1 : i64, mhlo.num_replicas = 1 : i64} {
+  func.func private @"##_fwd_reverse#281"(%arg0: tensor<1x1x2x2x2xf32>, %arg1: tensor<1x1x2x2x2xf32>, %arg2: tensor<1x1x2x2x2xf32>, %arg3: tensor<1x1x2x2xf32>, %arg4: tensor<1x1x2x2xi1>, %arg5: tensor<1x1x2x2x2xf32>, %arg6: tensor<1x1x2x2xf32>, %arg7: tensor<1x1x2x2x2xf32>, %arg8: tensor<1x1x2x2xf32>) -> (tensor<1x1x2x2x2xf32>, tensor<1x1x2x2x2xf32>, tensor<1x1x2x2x2xf32>, tensor<1x1x2x2xf32>) {
+    %0 = stablehlo.multiply %arg7, %arg7 : tensor<1x1x2x2x2xf32>
+    %1 = stablehlo.add %arg7, %arg5 : tensor<1x1x2x2x2xf32>
+    %2 = stablehlo.add %arg7, %arg7 : tensor<1x1x2x2x2xf32>
+    %3 = stablehlo.multiply %arg6, %arg6 : tensor<1x1x2x2xf32>
+    return %0, %1, %2, %3 : tensor<1x1x2x2x2xf32>, tensor<1x1x2x2x2xf32>, tensor<1x1x2x2x2xf32>, tensor<1x1x2x2xf32>
+  }
+  func.func @main(%arg0: tensor<2x2x2x1x1xf32>, %arg1: tensor<2x2x2x1x1xf32>, %arg2: tensor<2x2x2x1x1xf32>, %arg3: tensor<2x2x1x1xf32>) -> (tensor<2x2x2x1x1xf32>, tensor<2x2x2x1x1xf32>, tensor<2x2x2x1x1xf32>, tensor<2x2x2x1x1xf32>, tensor<2x2x1x1xf32>) {
+    %0 = stablehlo.transpose %arg0, dims = [4, 3, 2, 1, 0] : (tensor<2x2x2x1x1xf32>) -> tensor<1x1x2x2x2xf32>
+    %1 = stablehlo.transpose %arg1, dims = [4, 3, 2, 1, 0] : (tensor<2x2x2x1x1xf32>) -> tensor<1x1x2x2x2xf32>
+    %2 = stablehlo.transpose %arg2, dims = [4, 3, 2, 1, 0] : (tensor<2x2x2x1x1xf32>) -> tensor<1x1x2x2x2xf32>
+    %3 = stablehlo.transpose %arg3, dims = [3, 2, 1, 0] : (tensor<2x2x1x1xf32>) -> tensor<1x1x2x2xf32>
+    %c = stablehlo.constant dense<false> : tensor<1x1x2x2xi1>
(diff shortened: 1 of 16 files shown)


Pangoraw:
thanks for the pr. FYI, there is a PR that will also allow to do something similar in https://github.com/EnzymeAD/Enzyme/pull/2516.

Janssena:
> thanks for the pr. FYI, there is a PR that will also allow to do something similar in [EnzymeAD/Enzyme#2516](https://github.com/EnzymeAD/Enzyme/pull/2516).

Hey, interesting! Does it make more sense to have this in base Enzyme in order to add this functionality to a more generic way of passing reverse rules for custom call behaviour across languages?

Pangoraw:
> Does it make more sense to have this in base Enzyme in order to add this functionality to a more generic way of passing reverse rules for custom call behaviour across languages?

this is essentially a more general version of what is proposed in this pr at the function level, this should allow among other things Reactant users to generate custom rules for custom calls but also pure stablehlo computations.

Janssena:
> > Does it make more sense to have this in base Enzyme in order to add this functionality to a more generic way of passing reverse rules for custom call behaviour across languages?
> 
> this is essentially a more general version of what is proposed in this pr at the function level, this should allow among other things Reactant users to generate custom rules for custom calls but also pure stablehlo computations.

So what is the current state of that PR? I see it's been around for a while and this would be something I would be very interested in having. Is there some way I can help out?

Janssena:
I'll close this PR since a solution at the Enzyme level is preferable and in the works: https://github.com/EnzymeAD/Enzyme/pull/2516

PR #2904, opened 2026-08-16
Preserve the source's launch bounds through raising

wsmoses:
Reworked per review: the computed `minCtasTarget` knob is gone. With the serializer sink (#2907) addressing the register-pressure root cause, the right mechanism for occupancy targets is the source's own `__launch_bounds__(maxThreads, minBlocks)` — which clang already encodes as nvvm annotations that the importer carries onto the device function's passthrough and launch recognition already copies onto the error op. This PR simply consumes them: `nvvm.minctasm` and `nvvm.maxntid` are parsed where the target attributes are read and set on the kernel, with the preserved maxntid replacing the architecture-maximum fallback for runtime-block kernels (never overriding an exact `known_block_size`).

Verified end-to-end on a `__launch_bounds__(128, 4)` MWE: the raised PTX carries `.maxntid 128, 1, 1` and `.minnctapersm 4`, matching vanilla clang's device output, where previously the minctasm was silently dropped. Stacked on #2895; no pipeline options or env vars involved.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -263,56 +263,92 @@ struct AddLaunchBounds : public OpRewritePattern<gpu::LaunchFuncOp> {
     // gpu::LaunchFuncOp's)
     auto gpuFuncOp = launchOp->getParentOfType<ModuleOp>().lookupSymbol(
         launchOp.getKernel());
+
+    // Launch bounds the kernel was compiled with: clang encodes
+    // __launch_bounds__(maxThreads, minBlocks) as nvvm annotations, which the
+    // importer carries in the device function's passthrough and launch
+    // recognition copies onto the error op.
+    int32_t srcMaxntid = 0, srcMinctasm = 0;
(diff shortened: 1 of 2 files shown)


wsmoses:
Addressed the review direction: the branch now preserves the source's own launch bounds through raising. clang encodes `__launch_bounds__(maxThreads, minBlocks)` as nvvm annotations, which the importer already carries as passthrough pairs on the device function and launch recognition already copies onto the error op — they were just never consumed. The new commit parses `nvvm.minctasm`/`nvvm.maxntid` at the same place the target attributes are read and sets them on the kernel; verified end-to-end on a `__launch_bounds__(128, 4)` MWE, whose raised PTX now carries `.maxntid 128, 1, 1` and `.minnctapersm 4`, matching vanilla clang. Precedence: preserved minctasm > the computed `minCtasTarget` knob (which only fills when nothing was preserved); preserved maxntid replaces the architecture-maximum fallback for runtime-block kernels but never overrides an exact `known_block_size`. The stack (#2895 + this) is also rebased onto current main. Note MFEM's own sources pass no second launch_bounds argument, so the knob remains the only lever there — but source-annotated kernels no longer need it.

PR #2905, opened 2026-08-16
[ConvertParallelToGPU1] prevent emit a block that exceeds the capability on blockDim.z

xys-syx:
**Root Cause**:
An ordinary `k<<<dim3(2,2),128>>>()` when POLYGEIST_GPU_KERNEL_BLOCK_SIZE=512 lowers to `threads in (2,2,128)`: 
1. `ParallelLower` generate wrapper + two scf.parallel loops (grid parallel (2,2) around block parallel (128))
2. Upstream `canonicalize-parallel`  outer++inner -> flat (2,2,128) 128 is now at the last
3. `createSplitOp` scans backwards and inserts at the front, and the verification is a product not about cap
  ```
i=2    1128 = 128 <= 512   ->  [128]
  i=1    1282 = 256 <= 512   ->  [2, 128]
  i=0    256*2 = 512 <= 512   ->  [2, 2, 128]
```
4. position fill [0] -> x=2 [1] -> y=2 [2] -> z=128

The same source is legal at a requested 128 or 256 and not at 512, and both 512 and 1024 are in this pass's own `ALTERNATIVE_KERNEL_BLOCK_SIZES`.

**Fix**
A block whose dimensions already fit is emitted unchanged. One that does not is reordered, within two rules:
- the dimension the wrapper records as the launch's own `blockDim.x` takes x (trust the source code that map contiguous dimension onto `threadIdx.x`)
- the rest follow in descending order of extent

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -1871,6 +1871,11 @@ struct ParallelToGPULaunch : public OpRewritePattern<enzymexla::GPUWrapperOp> {
     if (!blockPop)
       return failure();
 
+    // The blockDim.x the wrapper records, read before the wrapper is replaced
+    Value originalBlockX;
+    if (wrapper->getNumOperands() == 6)
+      originalBlockX = wrapper->getOperand(3);
+
     // Only mutate once the match is certain: an op created before a bail
     // marks the IR changed on every scan, so a wrapper this pattern cannot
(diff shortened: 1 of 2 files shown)


xys-syx:
**Alternative way of fixing**

- Reorder every block, not only the ones that overflow
- Read the contiguous dimension out of the body: Walk the index expressions and take the induction variable whose coefficient is 1, instead of going by what the launch did. That would hold for a kernel written against the usual convention, and could improve one written badly. But it would be a large change: the coefficients have to be
recovered by walking the index expression rather than read off an affine map

PR #2907, opened 2026-08-16
Sink cheap address computations after the GPU serializer's LLVM pipeline

wsmoses:
This is the root-cause fix for the register-pressure gap discussed on #2904.

## Problem

The raising pipeline's affine composition produces identical closed-form index expressions at every access site. Every downstream stage then legally merges them — MLIR CSE first, and even when the MLIR-level IR is kept local, the serializer's own LLVM pipeline (EarlyCSE/GVN-PRE) re-hoists the chains into the kernel entry / loop preheaders. The result is a wall of long-lived scalar `i64` address values (e.g. hdiv setup: 74 entry defs vs clang's 18; 241 virtual `%rd` vs 152), which ptxas pays for with registers (80 vs 64) or spills (Det3D: up to 248 bytes). CodeGenPrepare only re-sinks GEP-shaped addressing that fits NVPTX addressing modes (`reg+imm`), so these scalar chains never come back down. Sinking at the MLIR level was measured and falsified: the serializer's O3 pipeline re-merges the clones even at `O = 1`.

## Fix

The only slot where sinking survives is *after* the serializer's optimization pipeline, immediately before instruction selection. MLIR's serialization infrastructure already exposes exactly that hook (`optimizedLlvmIRCallback`), but the upstream `gpu-module-to-binary` pass cannot pass callbacks. This adds `enzymexla-gpu-module-to-binary`, a drop-in replacement (same options, byte-identical output with `sink=0`) that installs a callback rematerializing cheap integer/address computations (integer binops, casts, GEPs) next to their users, per use block, erasing the hoisted original. Loop awareness: `sink=1` only sinks between blocks in the same loop; `sink=2` (default) also rematerializes into deeper loops, trading a recompute for liveness exactly where spills are worst. The machine layer (MachineCSE trivial-PRE) re-hoists only where its register-pressure heuristic deems it free, which is the right division of labor.

`REACTANT_LATE_SINK=<0|1|2>` overrides the mode in the clang-plugin round trip.

## Measurements (sm_120, CUDA 12.9)

QuadratureInterpolator det TU:

| kernel | base | sink=2 |
|---|---|---|
| Det3D dynamic (Lb1) spill bytes | 192 | 168 |
| Det3D dynamic (Lb0) spill bytes | 248 | 208 |
| CuWrap3D dynamic (Lb1) spill bytes | 160 | **72** |
| CuWrap3D dynamic (Lb0) spill bytes | 208 | **112** |
| Det3D<3,6> registers | 60 | **48** |
| Det3D<3,5> registers | 58 | **50** |
| Det3D<2,4> registers | 54 | **47** |

HcurlH1 mixed TU: PAHcurlMassApply2D 46→40 registers, PACurlCurlApply2D 40→38, PAHcurlL2ApplyTranspose2D 40→38; no kernel gained registers or spill.

Full MFEM gpu_unit_tests suite rebuild + run with the new default is in flight; I'll post the results (and benchmark deltas) here.

### DIFF
--- a/src/enzyme_ad/jax/BUILD
+++ b/src/enzyme_ad/jax/BUILD
@@ -1315,6 +1315,7 @@ cc_library(
         "@com_google_absl//absl/status:statusor",
         "@enzyme//:EnzymeMLIR",
         "@jax//jaxlib/gpu:triton_cc_proto",
+        "@llvm-project//llvm:Analysis",
         "@llvm-project//llvm:Core",
         "@llvm-project//llvm:ExecutionEngine",
         "@llvm-project//llvm:IPO",
--- a/src/enzyme_ad/jax/Passes/GpuModuleToBinarySink.cpp
+++ b/src/enzyme_ad/jax/Passes/GpuModuleToBinarySink.cpp
@@ -100,12 +100,12 @@ bool sinkCheapOpsInFunction(llvm::Function &F, int sinkMode) {
   return anyChange;
 }
 
-void sinkCheapOpsLate(llvm::Module &M, int sinkMode) {
+void sinkCheapOpsLate(llvm::Module &M, int sinkMode, bool dumpIR) {
   for (llvm::Function &F : M)
     if (sinkMode > 0 && !F.isDeclaration())
       sinkCheapOpsInFunction(F, sinkMode);
-  // Dump the IR as handed to instruction selection; lit tests key off this.
-  if (getenv("REACTANT_SINK_DEBUG"))
+  // The IR as handed to instruction selection; lit tests key off this.
+  if (dumpIR)
     M.print(llvm::errs(), nullptr);
 }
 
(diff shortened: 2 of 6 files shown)


wsmoses:
Validation update: the first full-suite run surfaced 33 failing MFEM TMOP TUs, which bisected to a pre-existing compile regression on bare main (unresolved-stub kernels serialized with a host x86-64 target — see the writeup on #2885, which fixes it), not to this PR: the failure reproduces identically with sink=0, with the upstream gpu-module-to-binary pass substituted via OVERRIDE_PASS_PIPELINE, and on main commits predating this branch, and disappears with #2885 applied while this PR's pass is active.

Full rebuild + 74-suite + benchmark is now running on a validation tree of main with the open CUDA PRs merged (#2885/#2857/#2864/#2859/#2896/#2895/#2904/#2898 + this one, sink=2 default); results to follow.

wsmoses:
Full validation complete on the merged tree (main + #2885/#2857/#2864/#2859/#2896/#2895/#2904/#2898 + this PR, `sink=2` default, minctasm off):

- **Build**: all 433 MFEM objects compile cleanly (the earlier TMOP failures were the bare-main regression fixed by #2885/#2908, not this PR).
- **Correctness**: full GPU unit suite **74/74 pass**, max linf_error 8.9e-16 (quad) / 5.6e-17 (hex) on the benchmark.
- **Benchmark** (bench_dfem_linearized_simple, best of 5, ratio vs vanilla clang):

| kernel | pre-sink baseline | with late sink |
|---|---|---|
| hex derivative local | 1.11x | **1.07x** |
| quad derivative global | 0.94x | **0.93x** |
| hex derivative global | 0.98x | 0.98x |
| quad fwd local | 1.10x | 1.11x |
| all others | ~1.00x | ~1.00x |

The sink recovers the hex-deriv-local occupancy loss from the root-cause side with `REACTANT_MINCTASM` unset; quad fwd local is the one residual where the #2904 knob still adds value (1.10x -> 1.04x in earlier measurement). No kernel regressed in registers, spill, or wall time.

PR #2908, opened 2026-08-17
Never build a GPU target attribute from a host chip

wsmoses:
Companion to #2885 (see the TMOP writeup there). When a launch is recognized through an unresolved host stub, the error op carries the host function's `target-cpu`/`target_features`; the target-attach walk at the end of `convert-parallel-to-gpu2` only guarded the *empty*-chip case, so `x86-64` (+x87 features) flowed into the `#nvvm.target` and serialization died in ptxas with `Value 'x86-64' is not defined for option 'gpu-name'`. `GPULaunchRecognition::initGPUModule` already refuses non-`sm_*` chips; this mirrors that guard here, falling back to the `sm_80`/`+ptx73` defaults.

With #2885 such kernels are dropped before serialization, so this is defense in depth for any stub that still reaches the walk; either alone turns the 33 failing MFEM TMOP TUs from a hard compile abort into a working build.

### DIFF
--- a/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
+++ b/src/enzyme_ad/jax/Passes/ConvertParallelToGPU.cpp
@@ -2859,6 +2859,15 @@ gdgo->erase();
                 dyn_cast_or_null<StringAttr>(err->getAttr("target_cpu")))
           sm = arch.getValue().str();
 
+      // A launch that was recognized through an unresolved host stub carries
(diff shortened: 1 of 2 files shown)


wsmoses:
Superseded by EnzymeAD/Reactant#81 + #2912: with stubs uniformly bound and bodyless declarations refused at recognition, no host-attributed target can reach serialization on the validated stack (full MFEM rebuild + 74/74 suite + benchmark, without this PR or #2885). The features-string sanitization concern is subsumed by not creating the placeholder at all. Closing.

PR #2909, opened 2026-08-17
Gate CompareAbs behind NoNan flags

Antipath1:
A second vibecoded fix for gating a pass that should be behind a NoNan flag. The problem is Nan != anything (including Nan) => true, and the opposite Nan == anything (including Nan) => false, which makes this pass wrong in two pathways, so we gate those behind the NoNan flag.


Pangoraw:
Can you rename the pattern to `no_nan_compare_abs`?

Antipath1:
> Can you rename the pattern to `no_nan_compare_abs`?

Done :)

Pangoraw (approved):

PR #2911, opened 2026-08-19
[XLAMath] throw lowering error only on tensor op

Pangoraw:
Closes https://github.com/EnzymeAD/Reactant.jl/issues/3196.

PR #2913, opened 2026-08-19
Fmul lowering for GPU

wsmoses:
should fix the fmuladd stuff [hopefully] @Pangoraw

### DIFF
--- a/.github/workflows/test-breeze.yml
+++ b/.github/workflows/test-breeze.yml
@@ -174,6 +174,10 @@ jobs:
                   PackageSpec(; name="Reactant", rev="${{ matrix.reactant_commit }}"),
               ])
           end
+
+          # Explicitely install Reactant_jll
+          Pkg.add("Reactant_jll")
+
           # Instantiate environment
           Pkg.instantiate()
           # Cleanup depot if possible
--- a/src/enzyme_ad/jax/Passes/LowerJIT.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerJIT.cpp
@@ -191,7 +191,6 @@ void buildLowerToNVVMPassPipeline(
 
   // Moved here to address https://github.com/EnzymeAD/Enzyme-JAX/issues/482
   pm.addPass(createArithToLLVMConversionPass());
-  pm.addPass(createConvertPolygeistToLLVM());
 
   // Host post-GPUModule-specific stuff
   buildHostPostPipeline(pm, options, toolkitPath, linkFiles);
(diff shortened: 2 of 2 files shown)


Pangoraw:
The interesting thing in https://github.com/EnzymeAD/Enzyme-JAX/pull/2911#issuecomment-5337653652 is that the error message is the one gated by the PR, not XLA refusing the op. So either the fix is not good, or Breeze CI is not running with the current's PR jll.

Pangoraw:
```
error: Failed to lower enzymexla math operation
  note: see current operation: %127 = "enzymexla.math.fmuladd"(%123, %14, %126) : (f32, f32, f32) -> f32

**Comment by wsmoses** [comment]
so I think there are two issues, both the lower xla math, and also the cpu/gpu backend needing to lower the instruction [currently onleexisting in the convert polygeist lowering pass]

**Comment by Pangoraw** [comment]
`Breeze.jl/benchmarking` does not have Reactant_jll in its deps so it does not pickup the local `libReactantExtra.so`. https://github.com/NumericalEarth/Breeze.jl/blob/main/benchmarking/Project.toml

**Comment by wsmoses** [comment]
that also may explain why we didn't see this earlier

**Comment by Pangoraw** [comment]
Ok by adding Reactant_jll to the env, we don't hit the xlamath problem, but a new one. Presumably added by the Polygeist lowering that we don't want in lower-jit?
```
error: Unhandled unrealized conversion cast %24 = "builtin.unrealized_conversion_cast"(%23) : (!gpu.async.token) -> !llvm.ptr
  note: see current operation: %24 = "builtin.unrealized_conversion_cast"(%23) : (!gpu.async.token) -> !llvm.ptr
  note: in entry
  note: used by llvm.call
/root/.julia/packages/Reactant/0G0Du/ext/ReactantCUDAExt.jl:1451:0

wsmoses:
Change it to just apply within the GPU module then, so it doesn't digest the caller code

We should eventually switch it over to fully use the polygrist lowering, but can do later

Pangoraw:
It's already run onto the GPU module only? 

https://github.com/EnzymeAD/Enzyme-JAX/blob/84cb53276021d2439e1b9c7ecc5c4cdde88e8938/src/enzyme_ad/jax/Passes/LowerJIT.cpp#L974-L980

I think we might want a `populateEnzymeXLAMathToLLVMConversionPatterns` that just does fmul for now, but then we are also missing tgamma in a similar fashion

Pangoraw:
Ok to merge and trigger a new jll as Breeze CI is green again?

wsmoses:
so that code checks to run it if there exists a gpu module, in contrast I meant run it only on the gpu modules [and not the host code, even if there is a gpu module]: https://github.com/EnzymeAD/Enzyme-JAX/pull/2919

Let's see if that resolves

PR #2919, opened 2026-08-20
GPU module convert polygeist to llvm pass

wsmoses:


Pangoraw:
```
LLVM ERROR: Can't add pass 'ConvertPolygeistToLLVM' restricted to 'builtin.module' on a PassManager intended to run on 'gpu.module', did you intend to nest

wsmoses:
Ah sigh, okay we should lift this restriction (or do polygrist in the jit pipeline properly).

In any case we can use your current conversion for the time being

Pangoraw:
i tagged O.0.406 with the current conversion, we actually need a similar lowering for the enzymexla.math.tgamma family of ops.

PR #2920, opened 2026-08-20
Rewrite ArithRaising to use a single walk across the module

Pangoraw:
the idea is to prevent walking the ir many times and instead apply rewrites through a simple `mlir::walkAndApplyPatterns` that walks the ir only once.
This speeds up running the pass, and in a follow-up we can put these patterns in tablegen.


src/enzyme_ad/jax/Passes/ArithRaising.cpp
```diff
@@ -40,695 +39,679 @@ using namespace mlir;
...
+struct ArithRaisingPass
+    : public enzyme::impl::ArithRaisingPassBase<ArithRaisingPass> {
+  using ArithRaisingPassBase::ArithRaisingPassBase;
+
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    auto *context = patterns.getContext();
+
+    // Ops with both a stablehlo and an mhlo raising.
+    // clang-format off
```
wsmoses, line 642: So I agree that multiple traversals per op isn't efficient, but we can do even better than a patternrewriter [which will keep inductively iterating], whereas here we can still do a single walk and just rewrite each op as it comes up

Pangoraw: `mlir::walkAndApplyPatterns ` does just that, it does not wait for a fixed point nor runs patterns on newly created ops.

wsmoses: Oh didn't know that api existed and didn't look sufficiently closely. Awesome!

PR #2922, opened 2026-08-20
[sparse_tensor] initial support for spmm

simeonschaub:
Per discussion in https://github.com/EnzymeAD/Reactant.jl/pull/3198#issuecomment-5339491689, introduce a custom `sparse.spmm` op that `stablehlo.dot_general` on a sparse CSR tensor is lowered to, which then gets translated into calls to either cuSPARSE or hipSPARSE. I'm not too familiar with this code base and most of this was written by Claude, but upon review this seemed like a sensible approach to me. I was also able to successfully test this on CUDA 13.1 at least


Pangoraw:
Nice! Could you use the same naming scheme as for other `enzymexla.*.*` ops lowering ? Something like `LowerEnzymeXLASparse.cpp`, etc..

PR #2927, opened 2026-08-24
PolygeistMem2Reg: fold piece stores into the slot value

wsmoses:
Piece stores of a slot previously landed in \`AliasingStoreOperations\`, blocking forwarding for any load after them. This is exactly the host side of the CUDA by-value kernel-argument bridge: the capture struct is built one field at a time and then loaded whole to cross into the GPU region, so the whole-struct load never forwarded and the struct survived to \`raise-affine-to-stablehlo\`, which has no struct support — a hard failure for every struct-capturing kernel (i.e. every MFEM \`DeviceTensor\` kernel) on the xla-gpu backend.

**Commit 1 — fold piece stores into the slot value.** The slot's value past a piece store is the value before it with that piece inserted (\`llvm.insertvalue\`), and a fresh alloca begins as \`undef\` so field-by-field initialization has a value to build on — the dual of the existing extractvalue support for piece loads out of whole stores. A piece store with no known value to fold into (an unknown write in between) conservatively resets the slot to unknown, as before.

**Commit 2 — what real MFEM TUs additionally need:**
- *Integer stores over several leaves.* Clang merges adjacent field stores and widens stores over padding (an \`i64\` at offset 8 covering \`{i32 field, padding}\`). \`coveredLeaves()\` enumerates the leaves a byte range covers (each must lie whole inside it; unclaimed bytes are padding), and the fold lands an \`lshr\`/\`trunc\` slice on each leaf (little-endian only, ≤8 leaves). The insert path can stop above a leaf at a same-extent wrapper (\`array<1 x i32>\`), so slices are cast to the type at the path.
- *Cross-block piece stores.* When the alloca is in an earlier block, the incoming value has no SSA name at the store; a synthetic whole-slot read created right before the store names it, and the existing deferred block-argument machinery forwards it like any load. Synthetic reads never count as progress and never enter the caller's erase loop — either one re-arms the outer fixpoint every round (observed: 5000+ rounds/min, effectively a hang).

The \`BranchOpInterface\` generalization that lets forwarding cross \`llvm.cond_br\`/\`llvm.invoke\` edges is now its own PR (#2934); this PR stands alone on cf terminators, and composes with #2934 for the CUDA launch shape (every launch sits behind an \`__cudaPopCallConfiguration\` invoke).

**Results** (union with #2934, on \`fem/integ/bilininteg_mass_pa.cpp\`, xla-gpu): surviving lambda-capture structs at raise time 10 → 5 (the remaining 5 are inside \`PAMassApply*\` bodies, not launch operands); zero \`failed to raise\` operand errors from by-value captures; pipeline-prefix replay minutes → ~1.3s. \`mwe_byval_static.cu\` (struct capture, static indexing) compiles and runs end-to-end on \`backend=xla-gpu\` with output identical to vanilla clang. CUDA backend verified against vanilla clang on the byval MWEs.

Tests: \`mem2reg_piece_stores.mlir\` (field-built slot read whole; piece store into an array member; piece store after an unknown write stays unforwarded; cross-block fold via added block args, cf terminators; merged/widened i64 stores), \`write_of_a_part\` in \`mem2reg_contained.mlir\` updated from documenting the bail to checking the fold.

### DIFF
--- a/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
+++ b/src/enzyme_ad/jax/Passes/PolygeistMem2Reg.cpp
@@ -2168,6 +2168,53 @@ static bool extractPath(Type from, uint64_t at, Type want, const DataLayout &dl,
   return false;
 }
 
+// The leaves of an aggregate that a range of bytes covers: every leaf the
+// range touches must lie whole inside it, and bytes of the range no leaf
+// claims are padding. Fails where a leaf straddles the range's edge.
+static bool coveredLeaves(Type from, uint64_t at, uint64_t size,
+                          const DataLayout &dl,
+                          SmallVectorImpl<std::pair<uint64_t, Type>> &leaves,
+                          uint64_t base = 0) {
+  auto fromSize = typeSize(from, dl);
+  if (!fromSize)
+    return false;
+  if (base + *fromSize <= at || base >= at + size)
+    return true;
+  if (auto ST = dyn_cast<LLVM::LLVMStructType>(from)) {
+    uint64_t byte = 0;
+    for (Type member : ST.getBody()) {
+      auto msize = typeSize(member, dl);
+      if (!msize)
+        return false;
+      if (!ST.isPacked())
+        byte = llvm::alignTo(byte, dl.getTypeABIAlignment(member));
+      if (!coveredLeaves(member, at, size, dl, leaves, base + byte))
+        return false;
+      byte += *msize;
+    }
+    return true;
+  }
+  if (auto AT = dyn_cast<LLVM::LLVMArrayType>(from)) {
+    auto esize = typeSize(AT.getElementType(), dl);
+    if (!esize || !*esize)
+      return false;
+    for (uint64_t i = 0; i < AT.getNumElements(); i++)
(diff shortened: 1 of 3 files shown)


wsmoses:
Pushed a third commit fixing a latent correctness/termination bug the always-inline MFEM experiment exposed: an i64 store of a packed dim3 pair lands on a same-extent `array<2 x i32>` slot; `extractPath` accepts any same-extent landing, but `castToType` had no spelling of an integer as a multi-element aggregate — its tail is `llvm_unreachable`, which a release build falls through, and the pass fixpoint respun forever (a fully-inlined `bilininteg_mass_kernels.cpp` never finished: >50 min, vs 32s on main).

The commit teaches `castToType` the missing direction (through the matching vector — the exact inverse of its existing leaf-collection path) and gates both the piece-store classification and the fold on `castCompatible`, an honest subset of what `castToType` handles, falling back to per-leaf slicing or leaving the store in place. Timing on the same TU: main 32s → this branch 90s (real added work, no longer unbounded); the BranchOpInterface PR #2934 alone measures 34s (innocent). Regression test: `dims_pair` in `mem2reg_piece_stores.mlir`.

PR #2936, opened 2026-08-24
Don't batch or unroll checkpoint segment loops

ftynse:
### Problem

Peak memory of a checkpointed reverse pass depends on whether the segments divide the trip count.

`LoopCheckpointing::segmentLength` returns the constant `nInner` when they divide evenly, and a `select` when they don't. A constant trip count makes the segment loop eligible for `GreedyWhileLoopBatchFission`, which hoists every induction-indexed user out of the loop and computes all iterations at once. That rebuilds the whole tape the checkpointing was paying recompute to avoid.

So the divisible case, which should be the cheapest, is the one that blows up.

### Fix

Mark the loops the checkpointing scaffold creates with `enzymexla.checkpoint_segment`, and skip them in the three passes that would materialize their iteration space: `GreedyWhileLoopBatchFission`, `WhileElementwiseReductionToReduce`, and `enzyme-hlo-unroll`.

Marking happens where Enzyme-JAX calls into `LoopCheckpointing`, by diffing the `stablehlo.while` ops before and after the call. The shared code builds the scaffold through hooks that are not told what they are building, so there is no per-loop place to mark from.

The mark does not survive on every loop of the scaffold. A rewrite that widens a loop to carry the checkpoint snapshots out cannot add results in place, so it builds a fresh op, and a fresh op inherits no discardable attributes. The guard therefore also fires on a loop that encloses a marked one, since fissioning the outer loop rebuilds the same tape.

### Evidence

Ablating `greedy_while_loop_batch_fission` from the pipeline, at every trip count where the segments divide. Measured before checkpointing moved into `LoopCheckpointing`, so periods are stated as segment lengths.

    iters  seg  mincut  peak GiB  peak GiB   time s   time s
                          normal  no-fission  normal  no-fission
    -----  ---  ------  --------  ----------  ------  ----------
       25    5     off     22.97       22.92   0.495       0.528
      100   10     off     36.31       36.15   1.888       2.019
      400   20     off     63.85       62.60   7.493       8.009
     1600   40     off    118.77      115.52  30.900      33.024
       25    5      on     22.54        9.54   0.409       0.426
      100   10      on     36.19       10.03   1.573       1.622
      400   20      on     63.48       11.02   6.254       6.507
     1600   40      on    118.07       12.99       -      26.642

Gradients bit-identical in every cell.

With mincut on, the divisible trip counts drop by 2.4x to 9.1x, and the gap grows with the trip count: 13 GiB at 25 iterations, 105 GiB at 1600. They land alongside the neighbouring non-divisible trip counts, which measure 9.79, 10.48, 11.86 and 14.68 GiB at 50, 200, 800 and 3200. The divisible line does not merely improve, it stops being a separate line.

With mincut off, peak memory moves by at most 2.7%, so this is a separate issue from that one.

Ablation is a pessimistic bound on the patch. It drops the pattern everywhere, which costs a flat 6.9% of gradient time on the mincut-off line where the pattern is doing useful work and is not the memory problem. Skipping only the scaffold loops should not pay that. Where the pattern does fire, the cost is 3-4%.

### DIFF
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -37,6 +37,7 @@
 #include "src/enzyme_ad/jax/Implementations/XLADerivatives.h"
 #include "src/enzyme_ad/jax/Utils.h"
 #include <cstdint>
+#include <utility>
 
 using namespace mlir;
 using namespace mlir::enzyme;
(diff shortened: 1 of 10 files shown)


wsmoses:
@Pangoraw

Pangoraw (approved):
sg

PR #2937, opened 2026-08-24
Migrate the build system to bzlmod

giordano:
Motivation is to try and make JAX/XLA upgrades less painful, by adopting their same build system.  Last case on point is #2843.  This is a multi-project change, it requires changes also in Reactant.jl and ReactantBuilder.  ***NOTE***: I did not look at the changes, really, I'm opening the PR mainly to save the progress so far.


mofeing:
This is def an improvement, but last time I checked I couldn't figure out how we could reference an external package of one of our deps (e.g. take the LLVM commit used by XLA). In principle, we shouldn't be doing that since that's what blzmod does, but there are many deps that are not yet ported to it.

PR #2941, opened 2026-08-24
simplify-affine-exprs: drop parallel bound operands another operand subsumes

wsmoses:
An `affine.parallel` upper bound `min(a, b)` where `b ≥ a` for every value of the dims and symbols is just `min(a)` — and a lower-bound `max` is the dual. The redundant operand usually states the same size through different arithmetic: after affine-cfg folds a launch guard into the bound, MFEM kernels iterate `min(N, ((N+255) floordiv 256)·256)` — the size against its own grid rounded up to whole blocks.

The subtlety is that the two operands usually arrive as *separate symbols* (affine-cfg deliberately leaves operands that are already valid symbols unexpanded), so no purely map-level reasoning can relate them. The pruner therefore composes the bound map through affine-cfg's own normalizer with a new `throughSymbols` flag (default off, so affine-cfg's output is unchanged): a valid symbol that is a cast of, or arithmetic over, other values is composed through as well, down to values whose definition has no affine form. isl is then asked whether the comparison can ever fail, as the emptiness of the set where it does; the IR keeps its original, unexpanded maps. The composition reads the arithmetic the way affine-cfg read it when it wrote the bounds — casts preserve the value, sums and products do not wrap, signed division is `floordiv` — which is the reading every raised bound and integer set is already committed to, so the pruner runs unconditionally rather than behind an option. Unrelated operands and inexpressible expressions are conservatively kept.

On MFEM's mass-integrator TU (with #2940 exposing the grid arithmetic), every guard-folded launch bound of the form `min(N, ceil(N/256)·256)` becomes plain `symbol(N)` — the ideal single-symbol form for eventual grid-value specialization.

Tests: single-symbol min, the two-operand form related only through defining arithmetic, an unrelated min that must stay, the lower-bound max dual, and a product of symbols that must stay. Full lit suite green (1193/1193 together with #2940).

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -1069,8 +1069,10 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
 // thing through different arithmetic -- a grid counted in blocks against the
 // size it was computed from -- so each symbol operand is expanded through its
 // defining arithmetic down to shared base values before asking isl whether
-// the comparison can ever fail. Signed division by a constant is read as a
-// floor division, the same reading the affine raising gives it.
+// the comparison can ever fail. The expansion reads the arithmetic as
+// arithmetic on non-negative mathematical integers, the reading the
+// assume-nonneg-arith option asserts: casts and extensions preserve the
+// value, sums and products do not wrap, and signed division rounds down.
 
 static Value lookThroughCasts(Value v) {
   while (true) {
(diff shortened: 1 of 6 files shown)


src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+    if (auto c = v.getDefiningOp<arith::ExtUIOp>()) {
+      v = c.getIn();
+      continue;
+    }
+    break;
+  }
+  return v;
+}
+
+static bool isExpandableArith(Operation *def) {
```
ftynse: IDK what "expandable" means in this context. Looks like the function checks for what is representable in affine maps.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+  return false;
+}
+
+static void collectBases(Value v, SetVector<Value> &bases, unsigned depth) {
+  v = lookThroughCasts(v);
+  APInt cst;
+  if (matchPattern(v, m_ConstantInt(&cst)))
+    return;
+  Operation *def = v.getDefiningOp();
+  if (depth && isExpandableArith(def)) {
```
ftynse: Prefer the `depth != 0` form to the implicit magic-cast.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+    collectBases(def->getOperand(0), bases, depth - 1);
+    collectBases(def->getOperand(1), bases, depth - 1);
+    return;
+  }
+  bases.insert(v);
+}
+
+static isl_aff *affForValue(Value v, const DenseMap<Value, unsigned> &basePos,
+                            isl_local_space *ls, isl_ctx *ctx, unsigned depth) {
```
ftynse: If you are using raw C isl structures, use the lifetime markers (`__isl_keep`, `__isl_take`) in signatures to indicate the intended memory behavior. Otherwise, there are usually C++ wrappers for ISL, and I suspect you may achieve the same result with FPL/Presburger.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+      isl_aff_free(rhs);
+      return nullptr;
+    }
+    if (isa<arith::AddIOp>(def))
+      return isl_aff_add(lhs, rhs);
+    if (isa<arith::SubIOp>(def))
+      return isl_aff_sub(lhs, rhs);
+    if (isa<arith::MulIOp>(def))
+      return isl_aff_mul(lhs, rhs);
+    return isl_aff_floor(isl_aff_div(lhs, rhs));
```
ftynse: I'd much rather have an `if` here and assume it's a floordiv. Also, double-check that division rounding is the same between arith and affine/isl (e.g. towards 0 or towards negative infinity for negative results).

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+  if (found == basePos.end())
+    return nullptr;
+  return isl_aff_var_on_domain(isl_local_space_copy(ls), isl_dim_param,
+                               found->second);
+}
+
+LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
+                                  affine::AffineParallelOp op) {
+  isl_ctx *ctx = islAnalysis.getCtx();
+  constexpr unsigned kExpandDepth = 8;
```
ftynse: Why this arbitrary undocumented limit? If you want to avoid stack overflow from recursion, just don't use recursive functions.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+  bool anyChanged = false;
+  for (int isUpper = 0; isUpper < 2; ++isUpper) {
```
ftynse: I dislike this kind of "smartness" in the code, makes it harder to grok if you read thousands of lines per day. Just factor out the function and call it twice, for lower and for upper bounds.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+    SmallVector<int32_t> groups;
+    for (auto g : groupsAttr)
+      groups.push_back(g.getZExtValue());
+    if (llvm::all_of(groups, [](int32_t g) { return g <= 1; }))
+      continue;
+
+    // One isl parameter per base value a symbol operand expands to; the map's
+    // dims stay opaque set dimensions.
+    SetVector<Value> bases;
+    for (unsigned i = 0; i < map.getNumSymbols(); i++)
```
ftynse: never evaluate .size() and co in the upper bound, this is the most basic LLVM nitpick.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+      collectBases(operands[map.getNumDims() + i], bases, kExpandDepth);
+    DenseMap<Value, unsigned> basePos;
```
ftynse: I highly suspect a linear lookup in a vector of ~10 elements is more efficeint than building and querying a hashmap.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+    for (unsigned i = 0; i < map.getNumDims(); i++) {
+      isl_id *id = isl_id_alloc(ctx, "dim", (void *)(size_t)(i + 1));
+      space = isl_space_set_dim_id(space, isl_dim_set, i, id);
+    }
+    for (unsigned i = 0; i < bases.size(); i++) {
+      isl_id *id = isl_id_alloc(ctx, "sym", (void *)(size_t)(i + 1));
+      space = isl_space_set_dim_id(space, isl_dim_param, i, id);
+    }
+    isl_local_space *ls = isl_local_space_from_space(isl_space_copy(space));
+    space = isl_space_free(space);
```
ftynse: Why make a copy of space only to delete the origin right after? Also, do we really need a _local_ space.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+
+    SmallVector<isl_aff *> symAffs;
+    for (unsigned i = 0; i < map.getNumSymbols(); i++)
+      symAffs.push_back(affForValue(operands[map.getNumDims() + i], basePos,
+                                    ls, ctx, kExpandDepth));
+
+    // Convert each bound expression, splicing the expanded operand in for
+    // each symbol.
+    std::function<isl_aff *(AffineExpr)> getAff =
```
ftynse: Don't we already have this somewhere?

Even if we don't have this particular feature, we may at least have the construct-AffineExpr-from-arith and we can reuse this instead of separately walking arith ops.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+                                     dim.getPosition());
+      if (auto sym = dyn_cast<AffineSymbolExpr>(expr)) {
+        isl_aff *aff = symAffs[sym.getPosition()];
+        return aff ? isl_aff_copy(aff) : nullptr;
+      }
+      return nullptr;
+    };
+
+    SmallVector<AffineExpr> newExprs;
+    SmallVector<int32_t> newGroups;
```
ftynse: Random `int32_t` when everything else is using unsigned...

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1063,6 +1063,253 @@ LogicalResult handleAffineIfOp(IslAnalysis &islAnalysis, AffineIfOp ifOp) {
...
+    ls = isl_local_space_free(ls);
+    if (!changed)
+      continue;
+    anyChanged = true;
+    auto newMap = AffineMap::get(map.getNumDims(), map.getNumSymbols(),
+                                 newExprs, map.getContext());
+    Builder b(op.getContext());
+    if (isUpper) {
+      op.setUpperBoundsMapAttr(AffineMapAttr::get(newMap));
+      op.setUpperBoundsGroupsAttr(b.getI32TensorAttr(newGroups));
```
ftynse: Is it really a I32TensorAttr?

ftynse (approved):
Overall comments:

1. Looking through index casts is not a noop unless the integer width matches the intended index width; looking through sext/zext_ui is not a noop at all if the highest bit is set. I'd hide this optimization under a flag akin to -ffast-math because it's not safe in general. It can be made safe and conservative by using integer range analysis + target info to prove that the transformation is valid for integers within that range.
2. It should be possible to do the same with Presburger lib aka FPL.
3. If we only care about the ((n + (C - 1)) celidiv C) * C kind of pattern, we can pattern-match that instead of the NP-hard ILP solving that is needed for the integer set emptiness check.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Renamed to `isAffineArithOp` (and the expansion is now `expandValue`).

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Gone: the expansion is now an iterative worklist with a memo, so there is no depth to limit.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Added `__isl_keep` on the remaining raw-isl signature. Kept isl rather than FPL so the pruner shares the analysis (`IslAnalysis`) and the converter (`AffineExprToIslAffConverter`) the rest of this pass already uses; the expression half no longer has its own isl builder at all, see below.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Rewrote as explicit `if`s with `llvm_unreachable` for the fallthrough. On the rounding: `arith.divsi` truncates toward zero while affine `floordiv` (and isl) round toward −∞, which coincide only for non-negative dividends, so the pruner now only runs under the new `assume-nonneg-arith` option that asserts exactly that reading (the raising pipeline sets it). The divisor additionally has to be a positive constant for the op to count as affine arithmetic at all.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Replaced the recursion with an iterative post-order worklist (`expandValue`): a value is pushed, its operands are pushed if not yet expanded, and it is finalized once both are, with a memo so shared subexpressions expand once.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Factored into `pruneBoundMap(ctx, map, operands, groups, isUpper)`; `pruneParallelBounds` calls it twice.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Switched to `llvm::seq` / `llvm::enumerate` and dropped the manual `.size()` loops.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: The position map is gone: a base value gets its symbol index when it is appended to `bases`, and the memo (`DenseMap<Value, AffineExpr>`) is what keeps a shared base from being appended twice — without it a value reached along two paths of the DAG would become two symbols and the redundancy check would lose that they are equal.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Removed the copy: the space now goes straight into `isl_local_space_from_space`, and the local space is what both the converter and `isl_aff_var_on_domain` take.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: The AffineExpr→isl half now reuses the file's `AffineExprToIslAffConverter` instead of a second lambda. For the other half (arith → AffineExpr) I did not find something reusable: AffineCFG's normalizer emits `affine.apply` ops rather than handing back an expression, and the affine-scope canonicalizers work on `affine.apply`/`affine.min` operands, not on arith trees behind casts. So `expandValue` turns the arith DAG into an `AffineExpr` over fresh symbols and the bound map result is substituted with `replaceSymbols`, after which the converter does the rest.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Now `unsigned` throughout; the only `int32_t` left is the vector handed to `getI32TensorAttr`.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Yes: `lowerBoundsGroups`/`upperBoundsGroups` are `I32ElementsAttr`, and `AffineParallelOp::build` creates them with `getI32TensorAttr`, so this matches upstream.

wsmoses:
Thanks for the review. Addressed in 9232bccf9 (one commit, rebased on the earlier i1 fix); replies inline for the line comments. On the three overall points:

**Casts.** Agreed that seeing through `index_cast`/`extsi`/`extui` and reading `divsi` as `floordiv` is unsound for arbitrary integer code. Rather than try to prove non-wrapping per value, the pruner now runs only under a new pass option, `assume-nonneg-arith` (default off), documented as reading the integer arithmetic as arithmetic on non-negative mathematical integers: casts and extensions preserve the value, sums and products do not wrap, and signed division rounds down. `raise.cpp` enables it in the raising pipeline, where the operands are sizes and thread/block indices. The test checks that nothing is pruned without the option.

**isl vs. FPL.** Kept isl, so the pruner shares `IslAnalysis` and the file's `AffineExprToIslAffConverter` with the rest of this pass instead of introducing a second solver; the ad-hoc AffineExpr→isl lambda is gone. The sets involved are tiny (a handful of parameters, one affine inequality per check), so the cost is negligible either way.

**Pattern-matching `ceildiv` instead.** I did look at that first, but the shapes vary enough that a syntactic match kept missing: the grid extent usually arrives as a separate SSA value computed with `divsi` (not a `ceildiv` in the map), the lower-bound `max` is the dual, and the same expansion is what #2946 builds the loop domain from. The expansion is now an iterative worklist over add/sub/mul-by-constant/div-by-positive-constant with a memo, no recursion and no depth cap.

ftynse (approved):

wsmoses:
f93d895 replaces the pruner's private symbol expansion with affine-cfg's normalizer:

- `AffineApplyNormalizer` / `fully2ComposeAffineMapAndOperands` gain a `throughSymbols` flag, default off. Without it nothing changes (affine-cfg's own tests are untouched). With it, an operand that is already a valid symbol is composed through the casts and arithmetic defining it — the same cases the normalizer already expands for non-symbols (add/sub/disjoint-or, mul/shift by constant, div/rem with a symbol divisor, constants) — and stops at a symbol whose definition has no affine form, which is kept as the symbol it was. `legalCondition` and the operand loop use one predicate (`keptAsSymbol`), so the `need`/compose loop terminates exactly when nothing composable is left.
- A no-rewriter, question-form entry point `fully2ComposeAffineMapAndOperands(AffineMap *, SmallVectorImpl<Value> *, Region *scope, bool throughSymbols)` is exported: nothing in the IR moves, and `false` means an operand would have had to.
- `pruneBoundMap` composes the bound map through that entry with `throughSymbols=true` and asks isl per group; the map written back to the IR is the original one with the redundant results dropped. `lookThroughCasts`/`isAffineArithOp`/`expandValue` are gone. New `@opaque` test: `min(n, n*k)` stays.

One shared-mode change in affine-cfg's cast peeling, which now also applies to `throughSymbols=false`: `arith.extsi` of an `i1` is no longer peeled (`true` sign-extends to `-1`, so the extension is not value-preserving). Say so if you'd rather that stay as it was.

#2946 is rebased onto this and its domain builder composes through the same entry point.

wsmoses:
402ed1c drops `assume-nonneg-arith`. The pruning reads a bound map the way affine-cfg wrote it (signed division is `floordiv`, casts preserve the value, sums and products do not wrap), and every raised bound and integer set is already committed to that reading by the time this pass runs, so gating on an option asserting it only made this pass stricter than the one producing its input. The pruner runs unconditionally, `raise.cpp` is back to plain `simplify-affine-exprs`, and the `OFF` check in the test is gone. #2946 is rebased accordingly and its folds run unconditionally too.

PR #2943, opened 2026-08-25
xla_wrapper: carry scalars the runtime specializes executables over

wsmoses:
Consumer half of grid-value specialization (draft until the raiser produces it; runtime half: EnzymeAD/Reactant.jl#3220).

`enzymexla.xla_wrapper` gains `num_specialized`: how many trailing inputs are scalars the runtime jits in as compile-time constants, specializing and caching the executable per distinct value. The `convert-polygeist-to-llvm{backend=xla-gpu}` lowering splits those inputs into their own i64 array (index/narrower ints widened) and emits the six-argument `reactantXLAExecSpec` call; with `num_specialized` absent the lowering is byte-for-byte the old `reactantXLAExec` form.

This is the plumbing for dynamic launch extents on the raising path: after EnzymeAD/Enzyme-JAX#2940–#2942 and EnzymeAD/Reactant#85 every MFEM launch bound reduces to a single `symbol(N)`, so the raiser can emit each kernel with one trailing `tensor<1xi64>` extent argument, mark it specialized, and the runtime folds it to a constant per distinct N — validated end-to-end on the runtime side with a `dynamic_iota` module run at two extents from one module string.

Test: `lowering/xla_specialized.mlir` (split arrays + `reactantXLAExecSpec` signature). Full lit suite green (1195/1195); CUDA and existing xla-gpu paths unchanged (compat shim verified against a previously built binary).

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -261,7 +261,10 @@ def XLAWrapperOp : EnzymeXLA_Op<"xla_wrapper", [
     SymbolRefAttr:$fn,
     Variadic<AnyType>:$inputs,
     OptionalAttr<DictArrayAttr>:$arg_attrs,
-    OptionalAttr<DictArrayAttr>:$res_attrs
+    OptionalAttr<DictArrayAttr>:$res_attrs,
+    // How many trailing inputs are scalars the runtime specializes the
+    // executable over, injected as compile-time constants per distinct value.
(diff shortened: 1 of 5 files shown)


wsmoses:
Per review: a single path — every `xla_wrapper` lowers to `reactantXLAExec`, whose signature gains the trailing `(nconsts, consts)` pair (Reactant.jl#3220 renames its entry point to keep the name); unspecialized wrappers pass `0`/null with no consts alloca.

wsmoses:
@copilot fix the format

PR #2945, opened 2026-08-25
Flatten viewed shared-memory scratch so raising sees plain accesses

wsmoses:
Shared-memory scratch arrives as a static alloca viewed through a `memref2pointer`/`pointer2memref` round trip that changes shape and address space, which no access-based raising can see through. When the alloca is only read and written through such flat views (or direct affine accesses, which move over with their maps linearized), the chain is replaced with one flat static alloca the raising handles directly — scoped to the regions the pass raises.

Now just the flattening, independent of the other pieces split from the original PR: barrier no-op (#2959, merged), broadcast-store mask or-reduce (#2960), uniform buffer select (#2962).

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3698,10 +3698,17 @@ struct AffineToStableHLORaisingPass
       auto MT = alloca.getType();
       if (!MT.hasStaticShape())
         continue;
-      SmallVector<enzymexla::Pointer2MemrefOp> views;
-      SmallVector<enzymexla::Memref2PointerOp> casts;
+      // Views reach the scratch through chains of address-space casts and
+      // constant-offset geps; each view carries the byte offset its chain
+      // accumulated.
+      SmallVector<std::pair<enzymexla::Pointer2MemrefOp, int64_t>> views;
+      SmallVector<Operation *> chainOps;
       SmallVector<Operation *> directAccesses;
       bool viewedOnly = true;
+      if (!MT.getElementType().isIntOrFloat())
+        continue;
+      int64_t elemBytes = (MT.getElementType().getIntOrFloatBitWidth() + 7) / 8;
+      SmallVector<std::pair<Value, int64_t>> ptrWork;
       for (Operation *user : alloca->getUsers()) {
         auto m2p = dyn_cast<enzymexla::Memref2PointerOp>(user);
         if (!m2p) {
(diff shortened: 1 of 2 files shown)


wsmoses:
Pushed a third commit (a803c252a) extending this stack to scalar control flow, so kernels raise fully before the JIT:

- `scf.for` with uniform runtime bounds → `stablehlo.while` with a rank-0 counter (grid-stride reductions carry per-lane accumulators as batched iter args).
- Loops whose bounds vary **per lane** (thread-stride remainders, `scf.for` and `affine.for`) iterate a scalar counter to the max lane trip count under an active-lane mask; stores mask, iter args hold once a lane finishes.
- `expandAffineExpr` now consults recorded access maps, so a masked loop's lane-tensor IV keeps its lane map; accesses over such IVs force gather/scatter.
- `arith.select` of buffers expands per access like the `affine.if` form.
- Racy stores (value varying along axes the destination doesn't index) refine to lane 0, or under a mask to a masked-pick reduction (exact for one-hot guards / equal racing values).
- Release-build hardening: all raise-path lookups go through `lookupOrNull`/`count`, `alignMemoryAccess` reports validity via an ok-flag (probing a default `AffineValueMap`'s `getAffineMap()` reconstructs through a null context — found by valgrind), shape bails reject alignment instead of tripping compiled-out asserts.

**Result**: BilinearForm TU raises 10/10 kernels and passes the full raising pipeline in strict mode; mass-kernels TU raises 126/129 (247 whiles). The 3 stragglers contain genuine do-while loops (`scf.while` with body-first work) — next item. Lit suite 1199/1199.

wsmoses:
Fourth commit (bb2e41771): **both MFEM TUs now raise completely and pass the full raising pipeline in strict mode** — BilinearForm 10/10 kernels (129 whiles), mass-kernels 129/129 (255 whiles), lit 1199/1199.

The last pieces:
- **`scf.while` by rotation**: peel one before-region execution (producing the condition and carried values), then a `stablehlo.while` carrying (cond, args, buffers) whose body runs the do region followed by the before region again. Uniform conditions only.
- **Union-space masked-pick** for guarded racy stores on both store paths: a guard spanning axes the update lacks (or updates beyond the scatter grid) reduces in the union of the two spaces, keeping exactly the axes the store indexes.
- `llvm.mlir.undef`/poison/zero wrapper operands map to rank-0 tensor zeros (they satisfy `m_Constant` but carry no usable value attribute).

wsmoses:
Split the later commits out for review: this PR now carries only the shared-memory-patterns commit. The stack is #2944 ← #2945 ← #2948 (scatter mask or-reduce) ← #2949 (scalar control flow) ← #2950 (do-while + guarded racy writes; both TUs fully raise).

PR #2946, opened 2026-08-25
Fold comparisons using loop bounds

wsmoses:
An integer comparison the enclosing affine loops and conditionals already decide folds to a constant: the guard of a peeled grid-stride residual (`blockIdx + gridDim` against an extent sharing `gridDim`'s base), a thread-id test under a constant-extent axis, a test under the else of a single-inequality `affine.if`.

`LoopDomain` builds the isl domain of an operation's block: the enclosing bounds and integer sets as `getDomain` computes them, with every symbol composed through the arithmetic defining it (as `pruneBoundMap` composes bound operands) and equated with the composition, so the compared values compose into the same parameters. Values are mathematical integers, which is what isl compares, so `comparisonSet` takes signed predicates only; an unsigned predicate is rewritten to its signed form once both sides are proven non-negative on the domain, and is left alone otherwise. The domain is built per query, as the access simplifications build theirs, so the pattern runs under drivers with and without a listener.

`FoldCmpUsingLoopBounds` is a pattern of `populateAffineExprSimplificationPatterns`, so affine-cfg applies it in the driver that raises loops — `@raised` in the test has the comparison against the bound of a loop the raising made affine fold. simplify-affine-exprs is now that pattern set under the greedy driver and nothing else: the handler walk and the trailing `recreateExpr` walk were the same rewrites (`handleAffineIfOp` recreates its set itself).

Scoped to the comparison fold as discussed; the min/max fold (#3091), the decided-trip-count unroll (#3092) and the never-looping `scf.while` fold (#3093) follow on top of this one, and the bitwise-or split against a power of two is its own PR on main.

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -10,7 +10,10 @@
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/IR/IntegerSet.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "src/enzyme_ad/jax/Dialect/Ops.h"
+#include "src/enzyme_ad/jax/Utils.h"
+#include "llvm/ADT/ScopeExit.h"
 
 #include <isl/aff.h>
 #include <isl/aff_type.h>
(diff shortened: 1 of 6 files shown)


src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
 
+// The enclosing affine loop nest as an isl domain: one set dimension per
+// parallel/for induction variable, constrained by its bounds and by the
+// integer sets of the enclosing affine.ifs. Every value is composed through
+// the arithmetic defining it the way the bound pruning above composes bound
+// operands, and each value that composition stops at becomes an
+// unconstrained parameter. Bounds and constraints that do not convert are
+// dropped, over-approximating the domain, which is sound for proving a
+// condition constant in either direction.
+namespace {
```
ftynse: We should already have this when preparing for isl-style scheduling. No need to duplicate.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+struct AffineDomainCtx {
+  isl_ctx *ctx = nullptr;
+  isl_local_space *ls = nullptr;
```
ftynse: We can obtain the space and the domain from the `isl_set` below, don't duplicate. It is also wrong to assume a local space will stay the same, non-unit loop strides and modulo conditionals will introduce new local (existentially quantified) dimensions.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+  isl_set *domain = nullptr;
+  Region *scope = nullptr;
+  const DenseMap<Value, Value> *subst = nullptr;
+  DenseMap<Value, unsigned> ivPos, paramPos;
+  DenseMap<Value, AffineExpr> exprs;
```
ftynse: I have no idea what these are and they are not documented.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+            paramPos.try_emplace(v, paramPos.size()).first->second, mctx);
+      (i < map.getNumDims() ? dims : syms).push_back(e);
+    }
+    SmallVector<AffineExpr> result;
+    for (AffineExpr e : map.getResults())
+      result.push_back(e.replaceDimsAndSymbols(dims, syms));
+    return result;
+  }
+
+  AffineExpr expr(Value v) {
```
ftynse: No clue what this does and why. Not documented, doesn't make much sense. Looks like it is creating a map and immediately decomposes it, potentially calling itself recursively.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+
+  // `v` over the domain's dims and parameters; null unless `build` saw it.
+  isl_aff *aff(Value v) {
+    auto it = exprs.find(v);
+    if (it == exprs.end())
+      return nullptr;
+    return conv.getIslAff(it->second);
+  }
+
+  // Consumes set.
```
ftynse: This isn't proper documentation of the function. Use `__isl_take` to indicate consumption, as the rest of the isl and add proper documentation to what the function does.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+  bool emptyOnDomain(isl_set *set) {
+    set = isl_set_intersect(isl_set_copy(domain), set);
```
ftynse: It's a bad idea to reuse a variable.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+namespace {
+struct AffineDomainCtx {
+  isl_ctx *ctx = nullptr;
+  isl_local_space *ls = nullptr;
+  isl_set *domain = nullptr;
+  Region *scope = nullptr;
+  const DenseMap<Value, Value> *subst = nullptr;
+  DenseMap<Value, unsigned> ivPos, paramPos;
+  DenseMap<Value, AffineExpr> exprs;
+  AffineExprToIslAffConverter conv{{}, {}, nullptr, nullptr};
```
ftynse: It's a bad idea to share state, it is only needed for two functions.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+                               true, pos});
+        }
+      } else if (auto forOp = dyn_cast<affine::AffineForOp>(parent)) {
+        unsigned pos = addIV(forOp.getInductionVar());
+        boundRefs.push_back({forOp.getLowerBoundMap(),
+                             SmallVector<Value>(forOp.getLowerBoundOperands()),
+                             false, pos});
+        boundRefs.push_back({forOp.getUpperBoundMap(),
+                             SmallVector<Value>(forOp.getUpperBoundOperands()),
+                             true, pos});
```
ftynse: These should handle non-unit strides or error out.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+          continue;
+        isl_aff *zero =
+            isl_aff_val_on_domain(isl_local_space_copy(ls), isl_val_zero(ctx));
+        domain =
+            isl_set_intersect(domain, ir.isElse     ? isl_aff_lt_set(ea, zero)
+                                      : set.isEq(i) ? isl_aff_eq_set(ea, zero)
+                                                    : isl_aff_ge_set(ea, zero));
+      }
+    }
+    return true;
```
ftynse: use LogicalResult instead of boolean magic

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+    // set is a single inequality: e >= 0 fails exactly when e <= -1.
+    for (auto [ir, es] : llvm::zip(ifRefs, ifExprs)) {
+      IntegerSet set = ir.ifOp.getIntegerSet();
+      if (ir.isElse && (set.getNumConstraints() != 1 || set.isEq(0)))
+        continue;
+      for (auto [i, e] : llvm::enumerate(es)) {
+        isl_aff *ea = conv.getIslAff(e);
+        if (!ea)
+          continue;
+        isl_aff *zero =
```
ftynse: Why do we do this nonsense with single inequality? We should just directly convert an IntegerSet to an `isl_set` and intersect that with the existing domain.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+// loop-trip reasoning below become affine.
+static LogicalResult foldMinMaxUsingLoopBounds(IslAnalysis &islAnalysis,
+                                               Operation *op) {
+  bool isMax = isa<arith::MaxSIOp, arith::MaxUIOp>(op);
+  bool isUnsigned = isa<arith::MaxUIOp, arith::MinUIOp>(op);
+  Value a = op->getOperand(0), b = op->getOperand(1);
+  AffineDomainCtx c;
+  if (!c.build(islAnalysis, op, {a, b}))
+    return failure();
+  isl_aff *aA = c.aff(a), *bA = c.aff(b);
```
ftynse: These are terrible variable names.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+  if (!aA || !bA) {
+    isl_aff_free(aA);
+    isl_aff_free(bA);
```
ftynse: Use `ScopExit`.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+  bool bGeA = !aGeB && c.emptyOnDomain(isl_aff_lt_set(bA, aA));
+  if (aGeB) {
+    isl_aff_free(aA);
+    isl_aff_free(bA);
+  } else if (!bGeA) {
+    return failure();
+  }
+  Value chosen = (aGeB == isMax) ? a : b;
+  op->getResult(0).replaceAllUsesWith(chosen);
+  op->erase();
```
ftynse: Take an IRRewriter and use it, you'll thank me later when you don't have random segfaults because of uncontrolled used of `.erase` in patterns.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+// Collapse a min/max whose order the enclosing loop bounds already decide
+// (maxsi(tid + 2, 2) under tid >= 0), so the bound expressions feeding the
+// loop-trip reasoning below become affine.
+static LogicalResult foldMinMaxUsingLoopBounds(IslAnalysis &islAnalysis,
+                                               Operation *op) {
+  bool isMax = isa<arith::MaxSIOp, arith::MaxUIOp>(op);
+  bool isUnsigned = isa<arith::MaxUIOp, arith::MinUIOp>(op);
+  Value a = op->getOperand(0), b = op->getOperand(1);
+  AffineDomainCtx c;
+  if (!c.build(islAnalysis, op, {a, b}))
```
ftynse: So this will go an navigate the IR for every statement this function is invoked on. Quite inefficient. You should instead run an analysis to construct domains for the loop/conditional structure of some defined region and reuse it.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+  if (!c.build(islAnalysis, forOp, {lb, ub, step}))
+    return failure();
+  isl_aff *lbA = c.aff(lb), *ubA = c.aff(ub), *stA = c.aff(step);
+  auto freeAll = [&]() {
+    isl_aff_free(lbA);
+    isl_aff_free(ubA);
+    isl_aff_free(stA);
+  };
+  if (!lbA || !ubA || !stA) {
+    freeAll();
```
ftynse: If only we had ScopeExit

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+  cmp.getResult().replaceAllUsesWith(cst.getResult());
+  cmp.erase();
+  return success();
+}
+
+// The folds above, ordered so each exposes work for the next.
+static void foldUsingLoopBounds(IslAnalysis &ia, Operation *op) {
+  SmallVector<arith::CmpIOp> cmps;
+  op->walk([&](arith::CmpIOp cmp) { cmps.push_back(cmp); });
+  for (auto cmp : cmps)
```
ftynse: Almost always avoid auto.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1178,6 +1179,537 @@ LogicalResult pruneParallelBounds(IslAnalysis &islAnalysis,
...
+    (void)foldCmpUsingLoopBounds(ia, cmp);
```
ftynse: What's the point of returning error state from static functions if it is ignored?

ftynse (commented):
We should have most of this functionality already somewhere, strive to reuse. I'm quite confident that we have logic that constructs affine domains using Presburger, and it also has emptiness checks.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Reused now: the domain is the `getDomain` set already built for the scheduling/access-simplification path (`getIndexSetEx` → `FlatAffineValueConstraints` → `isl_basic_set_from_constraint_matrices`), with symbols additionally tied to their composition through arithmetic so a comparison on `blockIdx + gridDim` meets the bound on `blockIdx` in the same parameters.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Dropped the stored local space; every conversion takes `isl_local_space_from_space(isl_set_get_space(set))` of the domain as it currently is, so the divs that non-unit strides (`addAffineForOpDomain`) or mod conditions introduce live in the set and are never assumed away.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Replaced by `LoopDomains::Domain` with each field documented (set dims = induction variables, params = symbols and the values compositions stop at, the two position maps, and the affine scope compositions stay within).

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Now `LoopDomains::compose`, documented: it seeds `fully2ComposeAffineMapAndOperands` (the composition machinery the bound pruning above uses) with the identity map on the value; whatever the composition does not go through stays an operand, and each remaining operand is an induction variable (set dim), a substituted value (composed in turn, for scf.while before-arguments read at their inits) or a parameter.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done: `__isl_take`/`__isl_keep`/`__isl_give` on every isl argument and result, with proper doc comments.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Fixed (`inDomain`).

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: The converter is now a local of `getAffs`/`getDomain`; no shared mutable state.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Handled by the reused `FlatAffineValueConstraints::addAffineForOpDomain`: a non-unit stride with a constant lower bound gets the exact stride constraint (local `q` with `iv - lb - step*q = 0`); otherwise the stride is dropped, which over-approximates the domain and stays sound for proving a condition constant.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done, `LogicalResult`/`FailureOr` throughout.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done: the enclosing `affine.if`s go through `addAffineIfOpDomain` (`FlatAffineValueConstraints::create(IntegerSet, operands)`, complement for the else region), i.e. the whole set at once, as part of the reused `getDomain`.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Renamed (`lhsAff`/`rhsAff`, `lbAff`/`ubAff`/`stepAff`).

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done, `llvm::make_scope_exit` everywhere.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done: the folds take an `IRRewriter` and use `replaceOp`/`replaceOpWithNewOp`/`eraseOp`/`inlineBlockBefore`.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done: `LoopDomains` builds the domain once per block (`getDomain` on first use, cached; entries dropped when a loop body is inlined or erased) and every comparison/min-max/loop fold under that block reuses it, only converting its own operands.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Done.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: The driver now consumes them: the folds iterate to a fixed point, with each fold's result driving whether another sweep runs (which also replaced the hand-ordered cmp → minmax → loops → cmp sequence).

wsmoses:
37049c9f0 turns `foldUsingLoopBounds` into patterns, applied with the greedy driver after the handler walk:

- `LoopDomains` caches the isl domain per block and is the driver's `RewriterBase::Listener`: a cached domain is dropped when an op nested in a modified/inserted/erased op is touched, or when a value it is keyed on (dim or parameter) goes away. `FoldCmpUsingLoopBounds`, `FoldMinMaxUsingLoopBounds<arith::{Max,Min}{S,U}IOp>`, `UnrollDecidedFor<scf::ForOp | AffineForOp>` and `FoldNeverLoopingSCFWhile` take the domains by reference; `populateLoopBoundFoldPatterns` registers them.
- `populateForOpRaisingPattern` (affine-cfg's `ForOpRaising`) is in the same set, so a loop whose bound the fold makes affine raises to `affine.for` within the pass, and its body then sees the loop in its domain — the new `@raise` case has the comparison inside the raised loop fold to a constant.
- The driver runs top-down so a min/max folds under the conditional it sits in before `ForOpRaising` hoists the normalized bound out of the `affine.if` (bottom-up processed the loop first and lost the domain for `@ifguard`). Constant CSE is off so constants stay where they are and the existing simplify-affine-exprs checks hold; folding stays on (one `affine.if` set in `simplify-affine-exprs.mlir` loses an unused dim, and two unused loads in that test now have uses so the driver does not DCE them).
- The `affine.for` unroll replaces the IV with the lower bound directly when the map is a single dim/symbol, otherwise an `affine.apply`/`affine.max` (`@affine_single_trip`).

Full lit suite 1276/1276 locally.

wsmoses:
Reworked per the follow-up, and split:

- `ForOpRaising` is out of simplify-affine-exprs again (`populateForOpRaisingPattern` gone). The folds are now patterns of `populateAffineExprSimplificationPatterns`, so affine-cfg applies them in its raising driver; simplify-affine-exprs applies that same set after its handler walk.
- Because affine-cfg's driver has no listener, the listener-invalidated domain cache is gone: `LoopDomain` is built per query, like the access simplifications build theirs. The greedy application in simplify-affine-exprs (folding on, constant CSE off) drops an unused `affine.if` dim and canonicalizes a top-level dim to a symbol in two existing tests (`simplify-affine-exprs.mlir`, `affine_if_recreate.mlir`); the sets are the same.
- This PR is now only the comparison fold and `LoopDomain` (rebased onto main, one commit, `65b06d2b9`); `@raised` in the test shows the fold under affine-cfg on a loop the raising made affine. The rest is on top of this branch: #3091 (min/max), #3092 (decided-trip-count unroll of `scf.for`/`affine.for`), #3093 (never-looping `scf.while`).

Full lit suite on each branch: green apart from `affine_if_recreate.mlir`, whose expectation fix (above) is in this commit and passes locally.

wsmoses:
d9e1b5f, per the review:

- The pass body is the pattern set under the greedy driver and nothing else. The handler walk was the same six rewrites, and the trailing `recreateExpr` walk was already done by `handleAffineIfOp` (which recreates its set even without a domain, since #3089).
- `comparisonSet` takes signed predicates only (`llvm_unreachable` on an unsigned one); `FoldCmpUsingLoopBounds` rewrites an unsigned predicate to its signed form with `signedPredicate` (Utils.h) once both sides are proven non-negative on the domain, and gives up otherwise. #3093's while fold does the same.
- The bitwise-or split against a power of two moved out to #3095 (on main, since it needs no domain); `@tid` here compares the ids directly.
- `composedThrough` stays as is: a composition that stops short leaves the operand a `Value` that becomes a free isl parameter, so every set queried is a superset of the real one and any never/always answer stays sound — we only prove less. Failing would instead drop the whole `LoopDomain` when one bound symbol composes through something that cannot be a symbol, losing trivially decidable folds inside that loop. Happy to make it fail if you prefer the simpler reading.

#3091–#3093 rebased on this; full suite green on each.

wsmoses:
63ed2a8: `LoopDomain::compose` returns `FailureOr` and fails when `fully2ComposeAffineMapAndOperands` does not run to completion; `getAffs` propagates it, so a query over a value that does not compose through is not folded. In the constructor a symbol whose composition fails simply gets no tie (it stays its own parameter), so the domain itself is unaffected. #3092's `loopBounds` propagates the failure too; #3091–#3093 rebased, full suite green on each.

wsmoses:
930c153: `comparisonSet` returns `FailureOr<isl_set *>`, failing (and freeing its operands) on an unsigned predicate instead of `llvm_unreachable`; both callers here and in #3093 propagate the failure. Followers rebased, full suite green on each.

PR #2949, opened 2026-08-25
Raise scalar control flow: scf loops, per-lane trip counts, racy writes

wsmoses:
Targets `main` directly (rebased off the #2948 stack); one commit. The remaining unraised MFEM kernels were scalar control flow the batched raising had no story for:

- **`scf.for` with uniform runtime bounds** raises as a `stablehlo.while` whose counter is a rank-0 scalar of the loop's integer type; grid-stride reductions carry per-lane accumulators as batched iter args. Test: `raise_scf_loops.mlir` `@scfred`.
- **Per-lane trip counts** (thread-stride remainder loops, `scf.for` via raised bounds and `affine.for` via elementwise bound expansion): iterate a scalar counter to the maximum lane trip count with an active-lane mask `lb + k*step < ub`; stores mask, iter args keep their value once a lane finishes, and the carried select permutes back into the iter arg layout. Test: `@lanefor`.
- **`expandAffineExpr` consults recorded access maps**, so a masked loop's lane-tensor IV keeps its lane map rather than a fabricated scalar one; accesses over such IVs force the gather/scatter path (a slice start index cannot be a lane tensor).
- **Buffer-yielding branches that are written through** expand per access: an `affine.if` yielding one of two buffers (or an `arith.select` of memrefs) becomes a branch per load/store, so each access reaches a real buffer and the usual select/mask raising applies. A branch nothing writes through is left to #2962's whole-tensor select (`raise_buffer_select.mlir` is unchanged). Test: `@pingpong`.
- **Racy stores** (value varying along axes the destination does not index): lane 0 without a mask; a masked-pick reduction under one (exact for one-hot guards and equal racing values).
- **Release-build hardening**: `IRMapping::lookup` misses return null in NDEBUG and a default `AffineValueMap`'s `getAffineMap()` reconstructs through a null context — probing it was itself the crash (found with valgrind after gdb's ASLR-disable masked it). `alignMemoryAccess` now reports validity via an ok-flag, lookups go through `lookupOrNull`/`count`, shape bails reject alignment instead of tripping compiled-out asserts, failed while raises leave the abandoned op in place.

On this rebase (main alone vs. main + this PR, same pipeline): main fails `fem/bilinearform_ext.cpp` in strict mode (32 `Could not broadcast an init`, 24 `cannot raise op to stablehlo`, then the backend fatal); with this PR the TU compiles with no raise diagnostics. Full lit: 1298/1298.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -466,7 +466,10 @@ static LogicalResult affineMapToSlice(affine::AffineValueMap accessValueMap,
   return success();
 }
 
-static SmallVector<int64_t>
+// The tensor shape an access map reads: one extent per result, 1 for the
+// results that do not vary along a parallel axis. Fails on an access it
+// cannot size (an IV with no static range).
+static FailureOr<SmallVector<int64_t>>
 affineMapShape(affine::AffineValueMap accessValueMap, ParallelContext pc) {
   AffineMap map = accessValueMap.getAffineMap();
 
@@ -480,14 +483,18 @@ affineMapShape(affine::AffineValueMap accessValueMap, ParallelContext pc) {
     }
 
     Value iv = getIVForExpr(accessValueMap, E);
+    if (!iv)
+      return failure();
     if (affine::isAffineForInductionVar(iv) && !pc.isParallelIV(iv)) {
       shape.push_back(1);
       continue;
     }
+    if (!affine::isAffineInductionVar(iv))
+      return failure();
 
     auto range = getIVRange(iv);
     if (!range.has_value())
-      return {};
+      return failure();
 
     shape.push_back(range->getNumIters());
   }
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1172,8 +1246,106 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+                                   builder.getZeroAttr(UTy.getElementType())));
+        Value falseInit = stablehlo::ConstantOp::create(
+            builder, loc, boolTy,
+            SplatElementsAttr::get(boolTy, builder.getBoolAttr(false)));
+        auto reduce = stablehlo::ReduceOp::create(
+            builder, loc, ValueRange{update, mAligned},
+            ValueRange{zeroInit, falseInit},
+            builder.getDenseI64ArrayAttr(raceDims));
+        {
+          Block *rb = new Block();
```
Pangoraw: use builder API to create a block here

wsmoses:
Rebased atop main (a237f79ec) and split into independent PRs: #3101 fail-soft groundwork → #3102 scf.for while → #3103 masked while (stacked); #3104 racy stores and #3105 buffer branches (each directly on main). This PR stays as the combined reference.

PR #2950, opened 2026-08-25
Raise scf.while by rotation

wsmoses:
Retargeted at `main` for independent review; now a single commit with only the `scf.while` raising (the masked-pick union-space refinement and the undef-operand mapping that used to ride along stay with the scalar-control-flow stack, #2949).

A general `scf.while` raises by rotation: the before region runs once (peeled), producing the condition and the carried values; the `stablehlo.while` carries `(cond, args, buffers)`, and its body runs the do region followed by the before region again — raised a second time through a body-local mapping, so nothing about the before region has to be a duplicable "condition-only" block. Only a uniform (rank-0) condition is accepted; per-lane conditions are left to the masked-while machinery. Buffers the loop touches are collected from both regions and carried like every other raised loop's.

This is the shape `canonicalize-scf-for` cannot turn into `scf.for` (a strided condition on the *rotated* state, or state that the do region rewrites), and it is what MFEM's in-kernel Newton solves (batchitrans) and CSR row walks (sparsemat) come in as.

Lit: `raising/raise_scf_while.mlir` (`@dowhile_loop`, `@dowhile_store` — the do-region store lands in a carried buffer and the re-evaluated before region reads the carried state). Raising suite green apart from `affine_to_stablehlo_wrapint` / `memref_constant_store`, which fail on `main` before this change.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -1575,6 +1575,137 @@ static LogicalResult tryRaisingForOpToStableHLOWhile(
   return success();
 }
 
+// A general scf.while raises by rotation: its before region runs once
+// (peeled), producing the loop condition and carried values; the
+// stablehlo.while then carries (cond, args, buffers) and its body runs the
+// do region followed by the before region again. Only a uniform (rank-0)
+// condition is supported.
+static LogicalResult tryRaisingSCFWhileOpToStableHLO(
+    scf::WhileOp whileOp, IRMapping &parentMapping, OpBuilder &builder,
+    llvm::DenseMap<Value, affine::AffineValueMap> &maps, ParallelContext pc) {
+  IRMapping mapping = parentMapping;
+  auto wloc =
+      rewriteLocation(whileOp.getLoc(), pc.options.strip_llvm_debuginfo);
+  Block *before = whileOp.getBeforeBody();
+  Block *after = whileOp.getAfterBody();
+  auto condOp = cast<scf::ConditionOp>(before->getTerminator());
+  auto yieldOp = cast<scf::YieldOp>(after->getTerminator());
+
+  for (auto [arg, init] :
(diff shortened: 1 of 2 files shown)


wsmoses:
Added a robustness commit found probing `fem/quadinterpolator.cpp` (#2968): the runtime-dimension fallback kernels' per-lane `du[3]` scratch — currently modeled as one shared buffer — reaches the masked dynamic-update-slice with a lane-wide update (`tensor<9xf64>` into `tensor<4xf64>`), and the slice builder fatally aborts in result type inference. Lower-rank updates now normalize to explicit size-1 dims, and genuinely unfit ones report a clean per-kernel error instead (with this, that TU raises 138/170 kernels and the remaining 32 fail recoverably). The sound fix for those 32 is scratch *privatization* — batching lane-private scratch over the lane axes — which is follow-up feature work; an any-lane masked pick was prototyped and rejected as unsound for genuine per-lane read-back. Reproducer is the TU itself; a minimal lit case needs stack-only routing I could not isolate small yet.

wsmoses:
Added `raising/raise_masked_overhang.mlir` covering the two masked-store fixes that landed here while debugging quadinterpolator: `@overhang` (a 32-lane batch storing `s[t]` under a dynamic `t < n` guard into a 25-element buffer — update and mask slice to the buffer extent, since out-of-range lanes could only have stored out of bounds and thus carry a false mask) and `@dynbuf` (a masked single-element store into a dynamic-extent buffer raises instead of tripping the unfit-update report, which previously compared against kDynamic numerically).

wsmoses:
Heads-up for the stack: pb/shared-scratch-raise gained a commit (flatten accepts plain memref direct accesses + the raiser now errors on store-through-alias `pointer2memref(memref2pointer(...))` views instead of silently snapshotting — this was silently dropping every shared-scratch store in MFEM's block reductions), and pb/scalar-cf-raise got its module-global preprocessing scoped to raised regions (host loops were being transformed). Children of these will want a trivial rebase.

PR #2960, opened 2026-08-25
Or-reduce broadcast-store masks over axes the store does not index

wsmoses:
A single-lane broadcast store (`tid==0`) writes one representative value: when the stored value is invariant along a masked axis the destination does not index, the mask or-reduces over that axis and the store happens under the reduced predicate. Split out of #2945; independent of the other split PRs.

Rebased onto `main` (2026-09-06). Without it `main` refuses every lane-masked store whose destination the lane does not index (`masked affine.store is dependent on less dimensions than masked stored value`), including a guarded scalar write into a `memref<f64>`; mfem's kernels do this wherever one thread publishes a block-wide value.

Tests (`raise_broadcast_store.mlir`, checked line for line): the isolated lane-0 store, the full block pattern (scratch fill + broadcast + barrier), and a lane-masked store of a uniform value into a rank-0 buffer, whose or-reduced mask selects between the value and the buffer itself. A value that varies along the masked lane stays an error (`affine.store is dependent on less dims than stored value`): which lane wins is data-dependent.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3965,6 +3965,61 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
       Value mask = pc.mask;
       affine::AffineValueMap maskMap = maps.lookup(mask);
 
+      // A mask axis the destination does not index (a single-lane broadcast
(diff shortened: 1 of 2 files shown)


Pangoraw (commented):
I think we need to do the same thing in `emitStoreAsScatter`. I think it could benefit from extracting the update masking logic in a helper function to reuse in the scatter code as well.

EDIT: oh that's https://github.com/EnzymeAD/Enzyme-JAX/pull/2948 😄

PR #2965, opened 2026-08-25
MoveWhileToFor: accept a duplicated increment feeding the condition

wsmoses:
MFEM's `reducers.hpp` grid-stride phase (every `Vector` reduction) rotates into an `scf.while` whose exit is `andi(i < N, idx+1 != items_per_thread)` — a shape `MoveWhileToFor`'s conjunction support handles — but clang emits the `idx+1` increment twice, and the pattern's SSA-identity check between the compare operand and the `scf.condition`-forwarded value rejected it (CSE upstream would also mask it, but the pattern shouldn't depend on that). Accept a structurally identical add of the same operands.

With this, the loop converts to an `scf.for` over `items_per_thread` with the `i < N` residual as a carried guard, unblocking raising of the reduction kernels' serial phase under `xla-gpu`.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -1310,8 +1310,17 @@ struct WhileToForHelper {
             auto beforeYield = cast<scf::ConditionOp>(
                 loop.getBefore().front().getTerminator());
             auto inductValue = beforeYield.getArgs()[ba3.getArgNumber()];
+            // The condition may pass a duplicate of the increment (clang
+            // often emits the addi twice); accept a structurally identical
+            // add of the same operands.
             if (inductValue != steppingVal) {
-              continue;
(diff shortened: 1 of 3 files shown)


wsmoses:
CI fix: `canonicalizefor/while_logical_negation_conjunct.mlir` deliberately pinned that the duplicated increment blocks the conversion — it documents the NCMesh `DofFinalizable` hazard (the forwarded conjunct must never fold to false at exit). Verified the conversion is semantically correct on exactly that point: the forwarded `%ok` rides the shadow iter_arg, updated on the failing evaluation too (early exit freezes it; scan exhaustion carries the last element's flag), and is returned directly — never the constant false. Updated the test to assert the converted form line-by-line, keeping the hazard note.

The other two dbg failures (`parallel-lower-noreturn`, `diffrules_enzymexla_gpu_wrapper_fwd`) also appear on other branches' dbg lanes (e.g. #2943's run) and are unrelated to this change.

PR #2966, opened 2026-08-25
Fold a monotone live flag into an induction-variable predicate

wsmoses:
Follow-up to #2965's conversion of MFEM's grid-stride reduction loops: the resulting `scf.for` still carried an `i1` live flag (plus its counter), because the residual `i < N` early-exit condition was threaded as loop-carried state.

The flag is `live(k) = AND of f(counter(j)) < N over j < k` — and since the counter advances exactly once per live iteration and `f` is nondecreasing in it (all multiplicative factors provably non-negative, e.g. `index_castui` block sizes), that conjunction is just the previous iteration's test. The new `ForLiveFlagToIVPredicate` pattern (companion to `MoveWhileToFor`, matching its exact output shape: `ub = maxsi(X,1)+1`, last-iteration-only continuation test) rewrites the guard to `iv == lb || f(cnt0 + iv - lb - 1) < N`, after which the flag and counter iter_args fold away entirely — the loop body's guard becomes a pure, affine-analyzable function of the induction variable.

The cloned comparison strips overflow flags since the first iteration evaluates it one step out of range (masked by the `or`, but it must not be poison). Test carries the full converted output; the existing canonicalize-scf-for corpus is unchanged.

**Status (2026-09-02):** rebased onto current main (the only conflict was the pattern-list line). Not required for the MFEM CUDA sweep — the full 122-TU compile sweep is green without it (#2968) — so this is a general loop cleanup to consider on its own merits rather than a campaign dependency.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -1736,6 +1736,322 @@ struct WhileToForHelper {
   }
 };
 
+// A while MoveWhileToFor converted often carries an i1 "live" flag: true on
+// entry, re-computed each iteration as `live ? (f(counter) < N) : false`,
+// where the counter advances by one per live iteration. Every earlier test
+// held whenever the flag is still true, so with f nondecreasing in the
+// counter the conjunction is just the previous iteration's test: the flag
+// equals `iv == lb || f(counterAtEntry + trip - 1) < N`, with trip the
+// iteration's number, a pure function of the induction variable. Rewriting
+// it lets the flag and the counter fold away, leaving a guard later
+// analyses can reason about affinely.
+struct ForLiveFlagToIVPredicate : public OpRewritePattern<ForOp> {
+  using OpRewritePattern<ForOp>::OpRewritePattern;
+
+  // Whether `value` is a sum or product of non-negative constants and
+  // zero-extensions.
+  static bool isKnownNonNegative(Value value) {
+    SmallVector<Value> worklist{value};
+    DenseSet<Value> visited;
+    while (!worklist.empty()) {
+      Value current = worklist.pop_back_val();
+      if (!visited.insert(current).second)
+        continue;
+      APInt constant;
+      if (matchPattern(current, m_ConstantInt(&constant))) {
+        if (constant.isNegative())
+          return false;
+        continue;
+      }
+      if (current.getDefiningOp<ExtUIOp>())
+        continue;
+      Operation *definingOp = current.getDefiningOp();
+      if (!definingOp || !isa<AddIOp, MulIOp>(definingOp))
+        return false;
+      worklist.append(definingOp->operand_begin(), definingOp->operand_end());
+    }
+    return true;
+  }
+
+  // Whether `arg` reaches `value` through sums and products.
+  static bool usesArg(Value value, BlockArgument arg) {
+    SmallVector<Value> worklist{value};
+    DenseSet<Value> visited;
+    while (!worklist.empty()) {
+      Value current = worklist.pop_back_val();
+      if (current == arg)
+        return true;
+      if (!visited.insert(current).second)
+        continue;
+      Operation *definingOp = current.getDefiningOp();
+      if (definingOp && isa<AddIOp, MulIOp>(definingOp))
+        worklist.append(definingOp->operand_begin(), definingOp->operand_end());
+    }
+    return false;
+  }
+
+  // Whether `value` is nondecreasing in `arg`: `arg` itself, a constant or a
+  // value from outside the loop, a sum of such values, or a product of one
+  // with a non-negative factor from outside the loop.
+  static bool isMonotoneInArg(Value value, BlockArgument arg, ForOp loop) {
+    SmallVector<Value> worklist{value};
+    DenseSet<Value> visited;
+    while (!worklist.empty()) {
+      Value current = worklist.pop_back_val();
+      if (!visited.insert(current).second)
+        continue;
+      if (current == arg || matchPattern(current, m_Constant()) ||
+          loop.isDefinedOutsideOfLoop(current))
+        continue;
+      Operation *definingOp = current.getDefiningOp();
+      if (!definingOp)
+        return false;
+      if (isa<AddIOp>(definingOp)) {
+        worklist.append(definingOp->operand_begin(), definingOp->operand_end());
+        continue;
+      }
+      if (!isa<MulIOp>(definingOp))
+        return false;
+      Value lhs = definingOp->getOperand(0), rhs = definingOp->getOperand(1);
+      bool lhsVaries = usesArg(lhs, arg), rhsVaries = usesArg(rhs, arg);
+      if (lhsVaries == rhsVaries)
+        return false;
+      Value factor = lhsVaries ? rhs : lhs;
+      if (!loop.isDefinedOutsideOfLoop(factor) || !isKnownNonNegative(factor))
+        return false;
+      worklist.push_back(lhsVaries ? lhs : rhs);
+    }
+    return true;
+  }
+
+  // The predicate that holds of (rhs, lhs) when `pred` holds of (lhs, rhs).
+  static CmpIPredicate swappedPredicate(CmpIPredicate pred) {
+    switch (pred) {
+    case CmpIPredicate::slt:
+      return CmpIPredicate::sgt;
+    case CmpIPredicate::sle:
+      return CmpIPredicate::sge;
+    case CmpIPredicate::sgt:
+      return CmpIPredicate::slt;
+    case CmpIPredicate::sge:
+      return CmpIPredicate::sle;
+    case CmpIPredicate::ult:
+      return CmpIPredicate::ugt;
+    case CmpIPredicate::ule:
+      return CmpIPredicate::uge;
+    case CmpIPredicate::ugt:
+      return CmpIPredicate::ult;
+    case CmpIPredicate::uge:
+      return CmpIPredicate::ule;
+    case CmpIPredicate::eq:
+    case CmpIPredicate::ne:
+      return pred;
+    }
+    llvm_unreachable("unknown predicate");
+  }
+
+  LogicalResult matchAndRewrite(ForOp loop,
+                                PatternRewriter &rewriter) const override {
+    APInt step;
+    if (!matchPattern(loop.getStep(), m_ConstantInt(&step)) ||
+        !step.isStrictlyPositive())
+      return failure();
+    // The trip bound must be maxsi(X, 1) + 1 with the last iteration's
+    // continuation test `iv < X`: only the final iteration can fail it, so a
+    // frozen counter is never read by a live iteration.
+    auto ubAdd = loop.getUpperBound().getDefiningOp<AddIOp>();
+    if (!ubAdd)
+      return failure();
+    Value maxVal = nullptr, boundX = nullptr;
+    for (int i = 0; i < 2; ++i) {
+      APInt one;
+      if (matchPattern(ubAdd->getOperand(1 - i), m_ConstantInt(&one)) &&
+          one.isOne()) {
+        maxVal = ubAdd->getOperand(i);
+        break;
+      }
+    }
+    if (!maxVal)
+      return failure();
+    if (auto maxOp = maxVal.getDefiningOp<MaxSIOp>()) {
+      for (int i = 0; i < 2; ++i) {
+        APInt one;
+        if (matchPattern(maxOp->getOperand(1 - i), m_ConstantInt(&one)) &&
+            one.isOne()) {
+          boundX = maxOp->getOperand(i);
+          break;
+        }
+      }
+    }
+    if (!boundX)
+      return failure();
+
+    for (BlockArgument flagArg : loop.getRegionIterArgs()) {
+      if (!flagArg.getType().isInteger(1))
+        continue;
+      APInt initTrue;
+      if (!matchPattern(loop.getTiedLoopInit(flagArg)->get(),
+                        m_ConstantInt(&initTrue)) ||
+          !initTrue.isOne())
+        continue;
+      // The flag's only use is as the condition of the body if.
+      if (!flagArg.hasOneUse())
+        continue;
+      auto ifOp = dyn_cast<scf::IfOp>(*flagArg.user_begin());
+      if (!ifOp || ifOp.getCondition() != flagArg ||
+          ifOp->getParentOp() != loop)
+        continue;
+      // The flag is re-yielded from the if: then a comparison, else false.
+      auto flagYielded =
+          dyn_cast<OpResult>(loop.getTiedLoopYieldedValue(flagArg)->get());
+      if (!flagYielded || flagYielded.getOwner() != ifOp)
+        continue;
+      unsigned flagResult = flagYielded.getResultNumber();
+      auto cmp =
+          ifOp.thenYield().getOperand(flagResult).getDefiningOp<CmpIOp>();
+      APInt elseFalse;
+      if (!cmp ||
+          !matchPattern(ifOp.elseYield().getOperand(flagResult),
+                        m_ConstantInt(&elseFalse)) ||
+          !elseFalse.isZero())
+        continue;
+      if (cmp->getParentOp() != ifOp)
+        continue;
+      // The test reads `varying < bound`, written either way round.
+      Value varying = cmp.getLhs(), bound = cmp.getRhs();
+      CmpIPredicate pred = cmp.getPredicate();
+      if (loop.isDefinedOutsideOfLoop(varying)) {
+        std::swap(varying, bound);
+        pred = swappedPredicate(pred);
+      }
+      if (!loop.isDefinedOutsideOfLoop(bound))
+        continue;
+      if (pred != CmpIPredicate::slt && pred != CmpIPredicate::ult &&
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
   }
 };
 
+// A while MoveWhileToFor converted often carries an i1 "live" flag: true on
+// entry, re-computed each iteration as `live ? (f(counter) < N) : false`,
+// where the counter advances by one per live iteration. Every earlier test
+// held whenever the flag is still true, so with f nondecreasing in the
+// counter the conjunction is just the previous iteration's test: the flag
+// equals `iv == lb || f(cnt0 + (iv - lb) - 1) < N`, a pure function of the
```
ftynse: I suppose `cnt0` is `counter-at-iteration-0`, but it needs to be spelled out.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+// induction variable. Rewriting it lets the flag and the counter fold away,
+// leaving a guard later analyses can reason about affinely.
+struct ForLiveFlagToIVPredicate : public OpRewritePattern<ForOp> {
+  using OpRewritePattern<ForOp>::OpRewritePattern;
+
+  static bool isKnownNonNegative(Value v, unsigned depth = 0) {
+    if (depth > 6)
```
ftynse: Can we export this out to the top of the file / some common header for future tweaking?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+      return false;
+    APInt cst;
+    if (matchPattern(v, m_ConstantInt(&cst)))
+      return cst.isNonNegative();
+    if (auto op = v.getDefiningOp<arith::IndexCastUIOp>())
+      return true;
```
ftynse: index is not guaranteed to be non-negative

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+             isKnownNonNegative(op.getRhs(), depth + 1);
+    return false;
+  }
+
+  // Whether `v` is `arg` plus/times values defined outside the loop, with
+  // every multiplicative factor along the path to `arg` known non-negative
+  // (so v is nondecreasing in arg and clonable at the loop's entry block).
+  static bool isMonotoneInArg(Value v, BlockArgument arg, ForOp loop,
+                              unsigned depth = 0) {
+    if (depth > 8)
```
ftynse: there's no reason for this to be 8 and for the other depth to be 6

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+      return isKnownNonNegative(op.getLhs(), depth + 1) &&
+             isKnownNonNegative(op.getRhs(), depth + 1);
+    return false;
+  }
+
+  // Whether `v` is `arg` plus/times values defined outside the loop, with
+  // every multiplicative factor along the path to `arg` known non-negative
+  // (so v is nondecreasing in arg and clonable at the loop's entry block).
+  static bool isMonotoneInArg(Value v, BlockArgument arg, ForOp loop,
+                              unsigned depth = 0) {
```
ftynse: I see the need to stop looking for efficiency purposes, but this looks like a hack to avoid stack overflows from recursion. The coding standards in MLIR is to not use recursion for use-def chains.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+      return true;
+    Operation *def = v.getDefiningOp();
+    if (!def)
+      return false;
+    if (auto add = dyn_cast<AddIOp>(def)) {
+      for (int i = 0; i < 2; ++i) {
+        Value side = add->getOperand(i), other = add->getOperand(1 - i);
+        if (usesArg(side, arg, depth) && !usesArg(other, arg, depth))
+          return loop.isDefinedOutsideOfLoop(other) &&
+                 isMonotoneInArg(side, arg, loop, depth + 1);
```
ftynse: Can't they _both_ be defined outside of the loop?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+    if (auto mul = dyn_cast<MulIOp>(def)) {
+      for (int i = 0; i < 2; ++i) {
+        Value side = mul->getOperand(i), other = mul->getOperand(1 - i);
+        if (usesArg(side, arg, depth) && !usesArg(other, arg, depth))
+          return loop.isDefinedOutsideOfLoop(other) &&
+                 isKnownNonNegative(other) &&
+                 isMonotoneInArg(side, arg, loop, depth + 1);
+      }
+      return false;
+    }
```
ftynse, line 1823: Constants defined in the loop should also be monotonous,shouldn't they?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+    if (auto op = v.getDefiningOp<MulIOp>())
+      return isKnownNonNegative(op.getLhs(), depth + 1) &&
+             isKnownNonNegative(op.getRhs(), depth + 1);
+    return false;
+  }
+
+  // Whether `v` is `arg` plus/times values defined outside the loop, with
+  // every multiplicative factor along the path to `arg` known non-negative
+  // (so v is nondecreasing in arg and clonable at the loop's entry block).
+  static bool isMonotoneInArg(Value v, BlockArgument arg, ForOp loop,
```
ftynse: I wonder if we can just try converting the body of the function to an AffineExpr and using Presburger/isl for a monotonicity check of that expression (f(i+1) >= f(i) \forall i  OR f(i+i) <= f(i) \forall i)

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+    return usesArg(def->getOperand(0), arg, depth + 1) ||
+           usesArg(def->getOperand(1), arg, depth + 1);
+  }
+
+  LogicalResult matchAndRewrite(ForOp loop,
+                                PatternRewriter &rewriter) const override {
+    // Step must be the constant one the conversion emits.
+    APInt stepCst;
+    if (!matchPattern(loop.getStep(), m_ConstantInt(&stepCst)) ||
+        !stepCst.isOne())
```
ftynse: Does it have to be constant _one_ as opposed to just a constant? We could adapt the argument of `f` accordingly.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+    // frozen counter is never read by a live iteration.
+    auto ubAdd = loop.getUpperBound().getDefiningOp<AddIOp>();
+    if (!ubAdd)
+      return failure();
+    Value maxVal = nullptr, boundX = nullptr;
+    for (int i = 0; i < 2; ++i) {
+      APInt one;
+      if (matchPattern(ubAdd->getOperand(1 - i), m_ConstantInt(&one)) &&
+          one.isOne())
+        maxVal = ubAdd->getOperand(i);
```
ftynse, line 1870: Nit: break?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+            one.isOne())
+          boundX = maxOp->getOperand(i);
+      }
+    }
+    if (!boundX)
+      return failure();
+
+    auto yield = cast<scf::YieldOp>(loop.getBody()->getTerminator());
+
+    for (auto [argIdx, flagArg] : llvm::enumerate(loop.getRegionIterArgs())) {
```
ftynse: BlockArguments carry their position, no need to enumerate around

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+                        m_ConstantInt(&elseFalse)) ||
+          !elseFalse.isZero())
+        continue;
+      if (cmp->getParentOp() != ifOp)
+        continue;
+      auto pred = cmp.getPredicate();
+      if (pred != CmpIPredicate::slt && pred != CmpIPredicate::ult &&
+          pred != CmpIPredicate::sle && pred != CmpIPredicate::ule)
+        continue;
+      if (!loop.isDefinedOutsideOfLoop(cmp.getRhs()))
```
ftynse: We don't need to hardcode rhs here

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+      // forwarded (possibly through a second guard if) to the yield.
+      BlockArgument counter = nullptr;
+      unsigned counterIdx = 0;
+      for (auto [ci, cArg] : llvm::enumerate(loop.getRegionIterArgs())) {
+        if (ci == argIdx || cArg.getType() != cmp.getLhs().getType())
+          continue;
+        // Locate `cnt + 1` in the then region.
+        AddIOp inc = nullptr;
+        for (auto &use : cArg.getUses()) {
+          if (auto add = dyn_cast<AddIOp>(use.getOwner())) {
```
ftynse: Just do `getUsers` instead

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+            }
+          }
+        }
+        if (!inc)
+          continue;
+        // The yield for this slot must trace back to the increment: either
+        // directly the if result whose then-yield is the increment, or that
+        // value routed through one more if whose then-yield forwards it.
+        Value slot = yield.getOperand(ci);
+        SmallVector<Value, 4> worklist{slot};
```
ftynse:
```suggestion
        SmallVector<Value> worklist{slot};
```

remember to never do this

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
```diff
@@ -1736,6 +1736,265 @@ struct WhileToForHelper {
...
+        bool reaches = false;
+        for (unsigned steps = 0; !worklist.empty() && steps < 4; ++steps) {
```
ftynse: Yet another magic stopping condition, and yet another different value

ftynse (commented):
This is extremely specific and could use some generalization.

An alternative approach to this by reusing existing transformations (and probably more generalizable) is to peel off the last iteration of the loop and attempt to hoist `live` out as loop-invariant.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Spelled out: counterAtEntry, and the comment now says trip for the iteration's number.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: The depth limits are gone altogether (see below), so there is nothing left to hoist.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Right; index_castui is no longer taken as non-negative. Only non-negative constants and zero-extensions are, through sums and products. The test's block size becomes a constant, which is what the raised MFEM launches carry.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Both depths are gone.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Every use-def walk (non-negativity, the counter's reach into the compared expression, monotonicity, the clone of the expression at the previous counter, the yield's route through the ifs) is now a worklist with a visited set, with no depth or step bound.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Yes: a value from outside the loop, or a constant, is now trivially monotone, and a sum is monotone when both operands are, so both sides may be from outside.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Yes; a constant counts wherever it is defined now.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: The coefficients here are loop-invariant SSA values (block sizes, offsets), so an affine form would have them as symbols and the monotonicity question becomes the sign of each symbolic factor, which Presburger cannot answer without the same non-negativity facts. The structural check is that sign reasoning, so kept it, with the recursion removed.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Generalized: any positive constant step, with the counter reconstructed as counterAtEntry + (iv - lb) / step.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Done, in both searches.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Switched to the LoopLikeOpInterface accessors (getTiedLoopInit, getTiedLoopYieldedValue, getTiedLoopResult); no more enumerate.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Now takes the comparison either way round: the side defined outside the loop is the bound, and the predicate is swapped when it was on the left.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Done.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Done; no inline sizes.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
wsmoses: Gone; a worklist with a visited set.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
ftynse: It may be worth having a limit on the size of the visited set, something like 1000 ops so we are reasonably fast. The previous guidance on extracting it out into the top of the file or some config struct stands.

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
ftynse: But does this condition catch it?

src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
ftynse: It doesn't mean they can't be dimensions, all valid symbols are also valid dimensions. The non-negativity facts can be encoded into presburger sets, just not into affine expressions specifically, and the monotonocity check would already require a set. So do try this, reusing as much of the existing functionality as possible, and report back concisely on whether it decreased the amount of code we have to write here and whether it could generalize better.

ftynse:
Any thoughts about the following?
> An alternative approach to this by reusing existing transformations (and probably more generalizable) is to peel off the last iteration of the loop and attempt to hoist live out as loop-invariant.

PR #2967, opened 2026-08-25
Convert rotated halving do-whiles to log-space fors

wsmoses:
The last raising blocker in MFEM's `reducers.hpp` kernels: the block tree reduction `for (i = bs>>1; i > 0; i >>= 1)` arrives do-while rotated — body, shift, and nonzero test all in the before region with the shifted value forwarded as carried state. `WhileShiftToInduction` (the existing Polygeist-heritage support for these) only matches the canonical body-in-after shape with a bare `ugt iv, 0` guard, so it never fires on this form.

`DoWhileShiftToFor` converts the rotated shape directly to a linear loop: `scf.for k in [0, max(bitwidth - ctlz(start), 1))` with `i = start >> k` — a zero start still runs the body once, matching do-while semantics, and the forwarded exit value is the constant zero. Test carries the full converted output; the canonicalize-scf-for corpus is unchanged.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -3003,12 +3003,14 @@ struct DoWhileShiftToFor : public OpRewritePattern<WhileOp> {
 
   LogicalResult matchAndRewrite(WhileOp loop,
                                 PatternRewriter &rewriter) const override {
-    // The after region must be a pure passthrough.
+    // The after region must be a pure passthrough: every yielded value is an
+    // after block argument, whose position names the condition operand it
+    // forwards. The order need not match the before arguments.
     Block &after = loop.getAfter().front();
     auto afterYield = cast<scf::YieldOp>(after.getTerminator());
-    for (auto [i, o] : llvm::enumerate(afterYield.getOperands())) {
+    for (Value o : afterYield.getOperands()) {
       auto ba = dyn_cast<BlockArgument>(o);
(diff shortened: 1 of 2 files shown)


wsmoses:
Good catch — fixed in 189c70da6. The indices only lined up because the passthrough check demanded the identity mapping, and the result bookkeeping could even index out of bounds when the condition forwards more values than the before region carries. The pattern now follows the after yield to translate between the two spaces (`beforeIvIdx` → yield operand → `afterIvIdx`), carried slots yield the condition operand their after argument names, and each while result maps to the for result of a before slot that feeds it (bailing when none does). Passthroughs may now permute, and the identity restriction is gone. Also fixed en route: the scf.for builder only auto-creates the implicit yield when there are no iter args, so `getBody()->getTerminator()` on the carried-value path walked an empty block — now the implicit yield is erased only when present. Added `@treereduce_swap` (permuted condition/yield with a carried counter) to the test with full CHECK lines.

wsmoses:
Also fixed in 30bcc50de: the induction-variable cast was unconditional, but for an index-typed shift variable it produced an invalid index→index cast — after first taking `getIntOrFloatBitWidth` of a type that has none (and `arith.shrui`/`cmpi`/`math.ctlz` all accept index, so the shape is reachable). The cast is now emitted only for non-index types. Since index has no fixed width, the trip count is computed in i64 — `index_castui` zero-extends into it, so `64 - ctlz` is the value's bit length regardless of how narrow the index lowering is — and the exit value becomes a constant index zero. Added `@treereduce_index` with full CHECK lines.

PR #1589, opened 2025-11-07
AlwaysInliner: disable single block optimization if unreachable

Pangoraw:
we can hit this assertion otherwise

https://github.com/EnzymeAD/Enzyme-JAX/blob/51d94b351ff8aaf7c6f3aae1c949f3611c394c7d/src/enzyme_ad/jax/Passes/AlwaysInliner.h#L74

PR #2971, opened 2026-08-25
AlwaysInliner: tolerate a no-return callee's empty terminator

wsmoses:
Fixes the pre-existing `-c dbg` CI failure of `parallel-lower-noreturn.mlir`: inlining a callee that cannot return hands `handleTerminator` an `llvm.unreachable` with no operands while the call has results. Release builds skipped the mismatched loop and #2832's caller-side yield repair filled the results; the assert aborted debug builds before the repair could run. The assert now permits the empty-terminator case and the loop guards the bound.


Pangoraw:
I think this could be replaced by https://github.com/EnzymeAD/Enzyme-JAX/pull/1589 (which copies the upstream inliner)

wsmoses:
@Pangoraw I'm going to merge this for now and let you resolve wrt the other pr in a follow up

PR #2973, opened 2026-08-26
Convert allocas with equal-sized punned views, bitcasting their accesses

wsmoses:
Clang stores a zeroed double as `i64 0` and copies one with i64 loads, so a `double du[3]` alloca arrives with both f64 and i64 `pointer2memref` views. The existing `convertLLVMAllocaToMemrefAlloca` bailed on the mixed views ("Non matching type of multi p2m"), which stranded the raw `llvm.alloca` inside the loops that use it — on MFEM's quadinterpolator.cpp this accounted for 56 of the raising failures (the alloca itself plus every enclosing `affine.for`).

Per review, rather than a separate pre-pattern this teaches the existing conversion (shared by `ConvertLLVMAllocaToMemrefAlloca` and `SimplifyInPlaceAlloc`) to accept a view whose element type is a different but equal-sized scalar when its uses are plain affine/memref loads and stores: the view retypes to the canonical element type — preferring the allocation's own scalar type, so the punned views retype toward it — with an `arith.bitcast` at each access. The casts fold against constants and paired accesses, so the du shape reduces to a plain `memref.alloca() : memref<3xf64>` with typed stores, which then raises through the zero-splat scratch path.

The in-place variant's existing early bail (view element already matching the allocation) is left untouched, so its behavior only changes where conversion previously failed outright.

### DIFF
--- a/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
+++ b/src/enzyme_ad/jax/Passes/LLVMToAffineAccess.cpp
@@ -142,13 +142,55 @@ convertLLVMAllocaToMemrefAlloca(FromAlloc alloc, RewriterBase &rewriter,
   }
   if (p2ms.size() == 0)
     return failure();
-  for (int i = 1; i < p2ms.size(); i++) {
+  // The allocation's own scalar element type is the canonical view type, so
+  // equal-sized punned views retype toward it rather than the other way.
+  {
+    Type allocScalar;
+    if constexpr (!inPlace)
+      allocScalar = alloc.getElemType();
+    else
+      allocScalar = alloc.getType().getElementType();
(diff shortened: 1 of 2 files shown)


wsmoses:
Reworked as suggested: the standalone `RetypePunnedView` pattern is gone, and the equal-sized-view support now lives in `convertLLVMAllocaToMemrefAlloca` itself, next to the existing i8-view tolerance — so both the `llvm.alloca` pattern and `SimplifyInPlaceAlloc` get it. The canonical element type prefers the allocation's own scalar type so punned views retype toward it. Same test output as before, plus a new `@punned_memref` case covering non-affine `memref.load`/`memref.store` users at a dynamic index; the full `llvm-to-affine-access` lit sweep passes.

PR #2979, opened 2026-08-26
Distribute barriers under serialized parallel axes before raising

wsmoses:
The raiser drops `enzymexla.barrier` as a no-op on the grounds that whole-tensor updates over a batched thread axis are already ordered — but that only holds when the axis is actually batched. A parallel axis of **dynamic extent** raises as a sequential loop, and dropping the barrier then runs each thread's entire body (tree reduction included) before the next thread ever stores its partial. MFEM's block reductions (`Vector::Sum`/`Min`/`Max`/norms, `reducers.hpp`) returned a single lane's value this way.

This distributes loops around such barriers before raising, reusing the cpuify machinery (`SCFCPUifyPass`'s pattern set exported as a scoped `distributeAroundBarriers` helper; the mincut variant recomputes barrier-crossing values, which keeps the crossing caches away entirely for these kernels). Downstream needs three companions, included here: `memref.alloca_scope` ops the distribution introduces are spliced away (they only delimit stack lifetime, meaningless under value semantics); one-dimensional dynamic-extent allocas raise as dynamic zero broadcasts; and a barrier still found under a serialized axis is now a hard error instead of a silent drop.

Together with #2978, MFEM's `Vector Sum` GPU test passes exactly through the raised XLA path (per-block partials verified against numpy under a PJRT harness).

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3398,9 +3398,63 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
     // it, which is exactly what materializing its initial value where the
     // alloca sits gives. Reads before any write see zeros.
     auto MT = alloca.getType();
-    if (!MT.hasStaticShape() || !isXLACompatiblePrimitive(MT.getElementType()))
+    if (!isXLACompatiblePrimitive(MT.getElementType()))
       return op->emitError("cannot raise dynamic or non-primitive alloca")
              << *op;
+    if (!MT.hasStaticShape()) {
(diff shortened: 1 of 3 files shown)


wsmoses:
On hold per review — #2980 lands the minimal operand-based guard first (error on a barrier whose spanned induction variables are dynamically sized, instead of miscompiling), and we're surveying which MFEM TUs actually hit it. Early results: linalg/vector.cpp (runtime `block_size = min(256, 1<<log2(N))` in `reduce()`), fem/quadinterpolator.cpp (runtime-`q1d` fallback kernels, 864 sites), linalg/batched/batched.cpp. Also worth noting: this PR's ancestor-walking trigger was too broad — a dynamic *grid* (ubiquitous and harmless, barriers don't span blocks) forced distribution even with static block axes, which is what dragged quadinterpolator into the subindex machinery; if we resurrect this, the trigger should be operand-based like #2980's check.

PR #2980, opened 2026-08-26
Error on a barrier over a dynamically sized parallel axis

wsmoses:
The raiser drops `enzymexla.barrier` as a no-op on the grounds that whole-tensor updates over a batched thread axis are already ordered. That's only sound for the axes the barrier spans when they're batched: an induction variable of a **dynamically sized** loop raises serialized, and dropping the barrier then lets one lane run its whole body — tree reduction included — before the other lanes store their partials. MFEM's block reductions (`Vector::Sum`/norms, reducers.hpp) silently returned a single lane's value this way.

Per review on #2979, this is the minimal guard first: check the induction variables actually passed to the barrier, and error when one belongs to a dynamically sized `affine.parallel` (or to a sequential loop after peeling) instead of miscompiling. Static-axis barriers are untouched.

Survey of who actually hits this on MFEM (block sizes are usually compile-time constants, so the reach is narrower than feared, but non-empty): the `reduce()` family in linalg/vector.cpp computes `block_size = min(256, 1 << log2(N))` at runtime, and fem/quadinterpolator.cpp's runtime-dispatch fallback kernels launch `MFEM_FORALL_2D(e, NE, q1d, q1d, 1)` with runtime `q1d` — full per-TU list to follow on the tracking issue once the sweep completes. #2979 (distribution) remains the candidate fix for actually raising those kernels.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3419,6 +3419,26 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
   // thread axis completes for the entire axis before the next op runs, which
   // is exactly what the barrier guaranteed.
   if (isa<enzymexla::BarrierOp>(op)) {
+    // Raised execution is ordered over whole tensors, so a barrier over
+    // batched axes is a no-op. That is only sound for the axes the barrier
+    // spans when they are batched: an induction variable of a dynamically
+    // sized loop raises serialized, with no whole-tensor ordering to lean
+    // on, and dropping the barrier would let one lane run ahead of the
+    // others' stores.
+    for (Value iv : op->getOperands()) {
+      auto ba = dyn_cast<BlockArgument>(iv);
+      if (!ba)
+        continue;
+      Operation *owner = ba.getOwner()->getParentOp();
+      if (auto par = dyn_cast<affine::AffineParallelOp>(owner)) {
+        if (!par.getConstantRanges())
+          return op->emitError(
+              "barrier over a dynamically sized parallel axis");
+        continue;
+      }
+      if (isa<affine::AffineForOp, scf::ForOp, scf::WhileOp>(owner))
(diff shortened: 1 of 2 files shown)


wsmoses:
Survey done (all 114 kernel-bearing MFEM library TUs recompiled with only this guard, distribution disabled): **29 TUs contain barriers over dynamically sized axes** — the hypothesis that block sizes are basically constants holds for the template-specialized kernel families but not for the runtime-dispatch *fallback* kernels, which are pervasive:

```
864 fem/quadinterpolator        560 integ/lininteg_domain_vectorfe   496 integ/lininteg_domain
384 fem/batchitrans             192 qinterp/grad_transpose           160 integ/lininteg_domain_grad
160 integ/lininteg_boundary_flux 160 integ/lininteg_boundary          96 qinterp/eval_transpose
 80 qinterp/det                  80 integ/bilininteg_vecmass_pa       80 integ/bilininteg_diffusion_ea
 64 linalg/vector (reduce)       64 integ/bilininteg_vecdiv_pa        64 integ/bilininteg_mixedcurl_pa
 48 fem/restriction              48 fem/prestriction                  48 integ/bilininteg_mass_ea
 32 integ/nonlininteg_vecconvection_pa{,_grad}  32 integ/bilininteg_{vectorfemass,vecdiffusion,curlcurl}_pa
 32 integ/bilininteg_dgtrace_ea  24 integ/bilininteg_dgdiffusion_pa  16 integ/bilininteg_{trace_jump,convection}_ea
  8 linalg/batched/batched        8 integ/bilininteg_hdiv_ea
```

The sources are (a) `reduce()`'s computed `block_size = min(256, 1<<log2(N))` in vector.cpp, and (b) every `MFEM_FORALL_2D/3D(..., d1d, q1d, ...)` fallback where d1d/q1d are runtime dispatch values. Score with guard-only: 55/114 TUs raise (soundly — these 29 now fail cleanly instead of miscompiling); the distribution approach in #2979 recovered ~11 of them including vector.cpp, whose raised `Vector::Sum` verified numerically exact. Suggested path: land this guard now; rework #2979's trigger to be operand-based (its ancestor-walk over-fired on dynamic *grids*) as the opt-in fix for these 29.

PR #2986, opened 2026-08-26
DotGeneralReshape: keep the dot's own result element type

### LINKED ISSUE #2984: DotGeneralReshape emits invalid IR for dots whose result element type differs from their operands' (bf16 x bf16 -> f32)
## Summary

`DotGeneralReshape` builds the hoisted `dot_general` with the **operand** element type
instead of the op's own result type. For a dot whose accumulation type differs from its
inputs -- e.g. the `bf16 x bf16 -> f32` dots JAX emits with `preferred_element_type=f32` --
the rewrite produces a bf16 dot followed by a `bf16 -> f32` reshape, and the module no
longer verifies.

## Minimal reproducer

```mlir
// repro.mlir
module attributes {transform.with_named_sequence} {
  transform.named_sequence @__transform_main(%arg0: !transform.any_op) {
    %0 = transform.structured.match ops{["func.func"]} in %arg0 : (!transform.any_op) -> !transform.any_op
    transform.apply_patterns to %0 {
      transform.apply_patterns.enzyme_hlo.dot_general_reshape
    } : !transform.any_op
    transform.yield
  }
  func.func @mixed(%arg0: tensor<4x8xbf16>, %arg1: tensor<8x16xbf16>) -> tensor<1x4x16xf32> {
    %0 = stablehlo.reshape %arg0 : (tensor<4x8xbf16>) -> tensor<1x4x8xbf16>
    %1 = stablehlo.dot_general %0, %arg1, contracting_dims = [2] x [0] : (tensor<1x4x8xbf16>, tensor<8x16xbf16>) -> tensor<1x4x16xf32>
    return %1 : tensor<1x4x16xf32>
  }
}
```

```
enzymexlamlir-opt --transform-interpreter --enzyme-hlo-remove-transform repro.mlir
```

```
error: 'stablehlo.reshape' op requires compatible element types for all operands and results
  note: see current operation: %1 = "stablehlo.reshape"(%0) : (tensor<4x16xbf16>) -> tensor<1x4x16xf32>
```

(Reproduced through Reactant.jl against `Reactant_jll` v0.0.405, which is where I hit it.)

## Cause

`EnzymeHLOOpt.cpp`, `DotGeneralReshape::matchAndRewriteImpl`:

```cpp
auto newDotGeneral = stablehlo::DotGeneralOp::create(
    rewriter, op.getLoc(),
    RankedTensorType::get(
        newShape,
        cast<RankedTensorType>(newLhs.getType()).getElementType()),   // <-- operand type
    newLhs, newRhs, ...);
rewriter.replaceOpWithNewOp<stablehlo::ReshapeOp>(op, op.getType(), newDotGeneral);
```

`op.getType().getElementType()` is what the trailing reshape needs. Same-element-type dots
are unaffected, which is why this only shows up on mixed-precision graphs.

## Where it bites

Reactant.jl's `:before_enzyme` pipeline on a bf16 training graph with q-chunked attention:
144 `bf16 x bf16 -> f32` dots in the module and the pipeline fails verification every time.
It looked shape/ordering sensitive at first (one compile of a given shape failed while
another succeeded), which is just whether the pattern got a chance to fire.
Workaround: `CompileOptions(excluded_passes=["dot_general_reshape"])`.

PR to follow.

## Environment

Reactant.jl main (v0.2.283 source) with `Reactant_jll` v0.0.405, Julia 1.12.7, Linux x86_64.
CPU backend; also seen on CUDA (H100).

AsafManela:
Fixes #2984.

`DotGeneralReshape` hoists a dim-inserting reshape from a `dot_general`'s operand to its
result, but built the new dot with the **operand** element type. For a dot whose
accumulation type differs from its inputs -- the `bf16 x bf16 -> f32` dots JAX emits with
`preferred_element_type=f32` -- that gives a bf16 dot followed by a `bf16 -> f32` reshape,
and the module stops verifying:

```
error: 'stablehlo.reshape' op requires compatible element types for all operands and results
  note: see current operation: %1 = "stablehlo.reshape"(%0) : (tensor<4x16xbf16>) -> tensor<1x4x16xf32>
```

Take the element type from the op being replaced instead. Dots whose result and operand
types agree are unaffected, which is why this only ever showed up on mixed-precision graphs.

Lit test: `test/lit_tests/dotgeneralreshape_mixed_element_types.mlir`.

### Testing

The C++ is **not built or run here** -- no bazel in this environment (the reproducer and the
diagnosis come from the shipped `Reactant_jll` v0.0.405, where the pattern is the same
compiled code). The lit test is written against the current sources but has not been
executed; please let CI have it, and tell me if the CHECK lines need adjusting.

https://claude.ai/code/session_012k6KZhhCNSCvZFKzErHaHH

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -28940,12 +28940,13 @@ struct DotGeneralReshape final
       newShape.push_back(dim);
     }
 
+    // The result element type may differ from the operands' (e.g. a bf16 x bf16
+    // -> f32 accumulating dot), so take it from the op being replaced: the
+    // trailing reshape has to be element-type preserving.
     auto newDotGeneral = stablehlo::DotGeneralOp::create(
         rewriter, op.getLoc(),
-        RankedTensorType::get(
-            newShape,
-            cast<RankedTensorType>(newLhs.getType()).getElementType()),
-        newLhs, newRhs,
+        RankedTensorType::get(newShape, op.getType().getElementType()), newLhs,
+        newRhs,
         stablehlo::DotDimensionNumbersAttr::get(
             rewriter.getContext(), adjustedLhsBatchingDims,
             adjustedRhsBatchingDims, adjustedLhsContractingDims,
--- a/test/lit_tests/dotgeneralreshape_mixed_element_types.mlir
+++ b/test/lit_tests/dotgeneralreshape_mixed_element_types.mlir
@@ -1,0 +1,26 @@
+// RUN: enzymexlamlir-opt --transform-interpreter --enzyme-hlo-remove-transform %s | FileCheck %s
+
+// A dot_general whose result element type differs from its operands' (a bf16 x
+// bf16 -> f32 accumulating dot, what JAX emits with preferred_element_type=f32)
+// must keep that result type when the feeding reshape is hoisted past it,
+// otherwise the trailing reshape changes the element type and fails to verify.
+module attributes {transform.with_named_sequence} {
+  transform.named_sequence @__transform_main(%arg0: !transform.any_op) {
+    %0 = transform.structured.match ops{["func.func"]} in %arg0 : (!transform.any_op) -> !transform.any_op
+    transform.apply_patterns to %0 {
+      transform.apply_patterns.enzyme_hlo.dot_general_reshape
+    } : !transform.any_op
+    transform.yield
+  }
+  func.func @mixed(%arg0: tensor<4x8xbf16>, %arg1: tensor<8x16xbf16>) -> tensor<1x4x16xf32> {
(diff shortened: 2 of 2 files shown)


AsafManela:
Update: **this is now built and verified**, so the "not built or run here" caveat in the
description no longer applies.

I compiled the patch into a JLL and compared it against an otherwise identical build from
the same pinned Enzyme-JAX commit, differing only by this change (confirmed by inspecting
each binary rather than trusting the build labels -- my first attempt at this comparison
was wrong because both bundles turned out to be patched).

Running the reproducer from the issue through `dot_general_reshape`:

```
without this patch:  error: 'stablehlo.reshape' op requires compatible element types for all operands and results
with this patch:     OK -- the dot keeps its f32 result and the trailing reshape is element-type preserving
```

A seeded fp32/bf16 smoke test gives bit-identical numerical results on both builds, so the
change is behaviour-preserving outside the failing case.

The lit test in this PR has still not been executed -- I verified through the compiled
pass rather than through `enzymexlamlir-opt` -- so please let CI have it and tell me if
the CHECK lines need adjusting.

PR #2993, opened 2026-08-27
Raising: dead lanes must not scatter at all

wsmoses:
A store under a lane mask raises as gather-select-scatter, writing the *original* value back on masked-out lanes. But a dead lane's index expression is unconstrained — strip-mine bound clamps (`max(extent, tid + bd)`) collapse dead lanes onto live lanes' slots — and scatter applies duplicate indices in unspecified order, so the stale write-back can land after the live lane's update and clobber it.

Found via mfem's QuadratureInterpolator under the XLA backend (#2968): the dynamic 2D tensor-product eval kernel produced partially-zero, element-varying results; a PJRT harness on the raised module reproduced it deterministically, and redirecting dead lanes out of bounds fixes it bit-exactly.

The fix selects dead lanes' indices to `-1`; out-of-bounds scatter updates are dropped, so masked-out lanes write nothing. This also removes the gather of the original values.

This PR now carries only the raising change. The `enzyme-hlo-opt` half that resolves the masked index ahead of the single-point scatter-to-DUS rewrite (`ScatterMaskedIndexSimplify`) is split out into its own PR, and the bounds guard on that rewrite is #3022. Ordering matters: with this change a masked-out lane's correctness rests on the scatter dropping its out-of-bounds index, so the DUS rewrite must not clamp it. This should land with, or after, one of those two.

The previous form let `enzyme-hlo-opt` fold a full-grid masked store into a pure `select`; with the index-select form the scatter survives in `affine_to_stablehlo_masked_scatter.mlir`. Teaching the dense-scatter fold to look through the index select would recover that.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -1472,21 +1472,6 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
                                              broadcastDims);
 
   if (pc.mask) {
-    SmallVector<int64_t> collapsedDims(scatterDimsToOperandDims.begin(),
-                                       scatterDimsToOperandDims.end());
-    SmallVector<int64_t> sliceSizes(collapsedDims.size(), 1);
-    Value orig = stablehlo::GatherOp::create(
-        builder, loc, input, indices,
-        stablehlo::GatherDimensionNumbersAttr::get(
-            loc.getContext(),
-            /*offsetDims*/ {},
-            /*collapsedSliceDims*/ collapsedDims,
-            /*operandBatchingDims*/ {},
-            /*startIndicesBatchingDims*/ {},
-            /*startIndexMap*/ collapsedDims,
-            /*indexVectorDim*/ (int64_t)gridShape.size()),
-        sliceSizes);
-
     // Broadcast the mask from its IV-space to the update's grid shape. A
     // mask axis over an IV the store does not index or-reduces away first:
     // the update is already known invariant along it (a variant update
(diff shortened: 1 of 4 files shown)


wsmoses:
@avik-pal @Pangoraw can I get a proper review/in depth audit of this [unreviewed by me so far to be clear] vibed up PR.

In particular one bad consequence is that this breaks the scatter removal so we may need to update things

wsmoses:
Pushed a required companion fix: the single-point scatter→`dynamic_update_slice` rewrite in enzyme-hlo-opt was unsound against the new out-of-bounds dead-lane indices — scatter drops OOB updates but DUS *clamps* the index back in range, so a masked-off store corrupted slot zero (caught by mfem's PAIdentityInterp, isolated with a buffer-level runtime diff + PJRT harness oracle). A new always-on pattern resolves the mask before those rewrites: scatter at the live index and write the original value back when masked off (a no-op store), which then lowers to a sound DUS. Lit test included.

src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1178,8 +1163,24 @@ emitStoreAsScatter(Location loc, Value update, Value input, ValueRange sIndices,
...
+        builder, loc, idxTy,
+        SplatElementsAttr::get(
+            idxTy, builder.getIntegerAttr(idxTy.getElementType(), -1)));
+    auto idxMaskTy =
+        RankedTensorType::get(idxTy.getShape(), builder.getI1Type());
+    SmallVector<int64_t> idxMaskDims;
+    for (int64_t k = 0, e = (int64_t)gridShape.size(); k < e; ++k)
+      idxMaskDims.push_back(k);
+    Value idxMask = stablehlo::BroadcastInDimOp::create(
+        builder, loc, idxMaskTy, broadcastedMask, idxMaskDims);
```
Pangoraw, line 1379: i think this can just be a reshape (since there should be a single additional trailing 1 element dim)

Pangoraw (commented):
~~Possibly we need the same thing for gather? Maybe we can move that logic in `buildGatherScatterIndices`~~

PR #2996, opened 2026-08-27
Raising: privatize lane-local scratch

wsmoses:
A thread-private array inside the lane-batched parallel (mfem's `real_t grad[3]` accumulators in the dynamic 2D grad kernels) was modeled as one shared buffer. A store whose index map does not involve the lane axes looks uniform to the raising, so every lane read back lane zero's value — QuadratureInterpolator's tensor PhysDerivatives froze the qy axis (caught via a PJRT harness oracle against a numpy reference, minimized to a 40-line kernel).

Fix: a new preprocessing normalization gives such buffers one leading dimension per lane axis and indexes every access with the lane IVs (bounded to the batched extents, small-buffer-gated). Lit test included; the minimized kernel raises and computes bit-exact results.

Part of #2968.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3660,6 +3660,125 @@ struct AffineToStableHLORaisingPass
   // of peeling to a serial loop: the axis becomes constant-extent and the
   // body sits behind an `iv < extent` guard, which the masking machinery
   // already understands. Barriers over the axis then stay batched no-ops.
+  // A thread-private array lives inside the lane-batched parallel: every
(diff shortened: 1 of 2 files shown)


wsmoses:
Pushed two fixes found running the MFEM suite end-to-end:

- **The 65k-element size budget was a silent miscompile**: mfem's 3D register-blocked kernels keep per-lane register tensors of 24×24 lanes × 126 registers = 72k elements; the budget skipped privatization and the scratch raised as a single shared tensor, so every lane read the wrong values (`VecDivPA` 3D wrong by ~3e-2 while the statically-specialized instantiations — which never hit the budget — were exact). Budget raised to 4M elements. A follow-up should make a skipped privatization fail the strict raise loudly instead of miscompiling.
- The lit test now uses affine loops only, so it runs on this branch without the scf.for raising machinery (CI was red because of that).

PR #2998, opened 2026-08-28
Raising: canonicalize maps before permutation matching; while fallback for constant-bound loops

wsmoses:
Two fixes that let mfem's EA mass-assembly kernels (7³ per-thread loop nests around chained 8×8 reductions) raise (#2968):

- `memoryEquivalentPermutation` failed when an eliminated unit axis left an unused dim behind in the yielded value's map — the while raise then rejected a carried accumulator whose map was memory-equivalent up to dead dims. Canonicalize both maps (drop unused dims/operands) before matching.
- When lockstep and unrolling both fail on a constant-bound `affine.for` (the chained-accumulator nests defeat both), fall back to the sequential while raise even under `prefer_while_raising=false`.

With these, `bilininteg_mass_ea`, `hdiv_kernels`, `hcurl_kernels`, `elasticity_kernels`, `mixedvecgrad_pa`, and `convection_ea` all raise strict; the raising lit suite is unchanged.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -1265,8 +1265,18 @@ static LogicalResult tryRaisingForOpToStableHLOUnroll(
 //  it returns the permutation to apply to a in order to align to b.
 //
 static std::optional<SmallVector<int64_t>>
-memoryEquivalentPermutation(const affine::AffineValueMap &a,
-                            const affine::AffineValueMap &b) {
+memoryEquivalentPermutation(const affine::AffineValueMap &aIn,
+                            const affine::AffineValueMap &bIn) {
+  // Dims that appear in no result do not affect the laid-out memory (an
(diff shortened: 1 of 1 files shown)


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -3391,6 +3401,14 @@ tryRaisingOpToStableHLO(Operation *op, IRMapping &mapping, OpBuilder &builder,
             .succeeded()) {
       return success();
     }
+    // The preference skipped the while for constant bounds; as a last
+    // resort a sequential while still raises what lockstep and unrolling
+    // could not (accumulators chained through nested reductions).
+    if (!pc.options.preferWhileRaising && forOp.hasConstantBounds() &&
+        tryRaisingForOpToStableHLOWhile(forOp, mapping, builder, maps, pc)
```
Pangoraw, line 3408: i think this would never hit this, since unroll above first does a raise to while then unroll

Pangoraw (approved):
looks good

PR #3002, opened 2026-08-28
While raising: broadcast under-attributed yields to the carried layout

wsmoses:
mfem's `bilininteg_diffusion_kernels.cpp` accumulators chain through index-table gathers whose raised values lose their lane attribution (a `() -> ()` map with a tensor type); the while raising then rejected the loop because the yield's layout could not be permutation-matched to the carried argument. Broadcast the yield up to the carried layout via `alignMemoryAccess` before matching, and let `alignMemoryAccess` report incompatibility through an `ok` out-flag instead of asserting on accesses it cannot size.

The tridiagonal golden changes form: the align at the loop tail now emits a permuting `broadcast_in_dim` (`dims = [0, 3, 1, 2]`) instead of a transpose — equivalent IR.

Part of #2968.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -881,6 +881,27 @@ static Block *getRaisedEntryBlock(Operation *op) {
   return &op->getParentRegion()->front();
 }
 
+// A loop-carried value can be yielded with fewer attributed axes than the
+// carried argument (a uniform chain through an index-table gather loses its
+// lane attribution): broadcast the yield up to the carried layout before
+// matching the permutation.
+static bool
+broadcastYieldToCarried(Value &yielded, Value carried, OpBuilder &builder,
+                        llvm::DenseMap<Value, affine::AffineValueMap> &maps,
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -1489,6 +1529,11 @@ static LogicalResult tryRaisingForOpToStableHLOWhile(
       Value raisedYieldedIterArg = mapping.lookup(yieldedIterArgs);
       Value raisedIterArg = mapping.lookup(iterArg);
 
+      if (!maps.count(raisedYieldedIterArg) || !maps.count(raisedIterArg))
+        return failure();
+      if (!broadcastYieldToCarried(raisedYieldedIterArg, raisedIterArg, builder,
+                                   maps, pc))
+        return failure();
       auto perm = memoryEquivalentPermutation(maps.lookup(raisedYieldedIterArg),
                                               maps.lookup(raisedIterArg));
```
Pangoraw, line 1706: i would change the memoryEquivalentPermutation here to be more permissive then since it's only used for this case.

PR #3006, opened 2026-08-28
Raising: buffer normalizations for viewed geps and raw accesses

wsmoses:
mfem kernels that are not fully inlined reach the raising with buffers addressed through pointer plumbing tensor semantics cannot stand for. This ports the foundation of the normalization suite used to bring the MFEM CUDA suite through the raising path (#2968):

- **rebaseViewedGeps**: a typed view carved out of a buffer at an offset (gep feeding a `pointer2memref`) rebases its accesses onto the underlying buffer at base + offset.
- **convertRawGepAccesses**: data-dependent indexing (CSR-style loops over runtime offsets) can never become affine and stays as raw gep+load; the access still addresses whole elements, so it becomes a plain memref access through a flat view, which raising turns into a gather.
- **inlineAllocaScopes / linearizeRegionBlocks**: alloca scopes only delimit stack lifetime, which is meaningless under value semantics; splice them into the parent so scratch normalizations see the whole function.
- **dropDeadPointerChains**: sweep the pointer plumbing the rewrites strand.

Follow-up PRs stack on this: buffer-branch expansion, packed capture-struct forwarding, struct scratch splitting/flattening.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -22,6 +22,7 @@
 #include "mlir/Dialect/Affine/LoopUtils.h"
 #include "mlir/Dialect/Affine/Utils.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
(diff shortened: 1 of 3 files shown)


wsmoses:
Split per review into three independent main-based PRs, each with lit tests: #3028 (convert raw gep accesses + dead-chain sweep), #3029 (rebase viewed geps), #3030 (alloca scopes / execute regions — reworked per review to handle them in the raising descent instead of inlining, with new lit coverage). The two-round fixpoint that lets #3028 and #3029 expose each other's opportunities rides with whichever lands last. The branch pb/buffer-normalizations stays as the base of the stacked PRs (#3010, #3011, #3012, #3015, #3016, #3025) until the pieces land, then those rebase onto main.

PR #3007, opened 2026-08-28
Raising: expand buffer-yielding branches into their accesses

wsmoses:
Stacks on #3006.

mfem's coefficient ternary (`const_coeff ? c(0,0,0) : c(i,j,e)`) arrives at the raising as a branch yielding a pointer or a whole memref — as an `arith.select` of buffers, an `scf.if` yielding a pointer whose arms compute geps, or an `affine.if` yielding a memref. No tensor value can stand for the yielded buffer, so each access is pushed down into a clone of the branch: only scalars cross the yield and the branch raises as a select. Arms may contain loads (reads are idempotent and clone safely per pushed-down access); branches whose buffer results die are rebuilt without them when scalar yields keep the branch alive.

The existing `raise_buffer_select` golden changes from a whole-tensor select to a per-access select — equivalent, and the expansion also covers the cases the whole-buffer path had to reject (a store through the selected buffer).

Part of #2968.


giordano:
> Stacks on #3006.

Quoting myself from https://github.com/JuliaMath/SpecialFunctions.jl/pull/552#issuecomment-5369010663

> Claude loves to talk about stacking PRs, but not [actually stacking them](https://gh.io/stacks-overview)

***Edit***: switched to a real stack.

wsmoses:
Recreated targeting main: the stack lock prevents changing this PR's base, and the branch has been rebased onto main.

PR #3009, opened 2026-08-28
Raising: forward the packed capture struct; distribute geps over selects

wsmoses:
Stacks on #3007 (which stacks on #3006).

A not-quite-inlined device lambda packs its captures (DeviceTensors: pointer + sizes) into a stack struct and reads them back through typed views at constant offsets. Forward each load to the unique dominating store of its slot so the pointer-typed members never reach the raising as memory:

- constant-attribute **and** constant-valued-SSA gep indexing;
- wide integer stores covering several slots (two i32 sizes packed into one i64 write, extracted little-endian shift+trunc);
- runtime-indexed shape reads (`sizes[i]` walked in a loop, as a `memref.load` or an affine map like `s0 floordiv 4`) resolved as a select chain over the constant-offset stores of their element type;
- stranded dead pointer selects (left by the buffer-branch expansion) skipped — narrowly, since a general dead-user skip admits big register scratch whose forwarding is unsound.

Also **distributeGepOverSelect**: a side-selected view (`select` of byte offsets or of the pointers themselves feeding geps) distributes the gep and the memref view over the select so the buffer-branch expansion (#3007) can take over.

This is what makes mfem's `bilininteg_vecdiffusion_pa.cpp` and the batched LOR kernels raise fully (#2968).

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3596,6 +3596,402 @@ struct AffineToStableHLORaisingPass
           AffineToStableHLORaisingPass> {
   using AffineToStableHLORaisingBase::AffineToStableHLORaisingBase;
 
+  // A side-selected view arrives as a select of byte offsets (or of the
+  // pointers themselves) feeding geps: distribute the gep and the memref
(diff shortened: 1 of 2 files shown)


wsmoses:
Recreated targeting main (content rebased): the stack lock prevents changing this PR's base.

PR #3012, opened 2026-08-28
Raising: split struct scratch and flatten viewed scratch

wsmoses:
Stacks on #3006.

Kernel scratch that is not a flat typed array reaches the raising in two shapes it cannot handle directly:

- **splitStructScratch**: a `memref.alloca` of an LLVM struct accessed through byte-offset typed views (mfem's `DeviceTensor` locals) splits each field into its own flat scratch; wide integer moves covering whole fields decompose into the fields they touch.
- **flattenViewedScratch**: a flat alloca whose accesses go through pointer chains re-viewed at byte offsets (`m2p → gep → p2m`) rebases them onto the alloca with the offset folded into the index.

Part of #2968.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3912,6 +3912,647 @@ struct AffineToStableHLORaisingPass
     }
   }
 
+  // Residue of an integer/index value modulo m, when the defining arith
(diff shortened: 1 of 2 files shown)


wsmoses:
Closing: this stacked `splitStructScratch`/`staticResidue` and `flattenViewedScratch` on #3006, whose base is gone. The same two helpers are filed independently against main as #2982 (split struct scratch) and #2945 (flatten viewed scratch), each with its own test.

PR #3013, opened 2026-08-28
Raising: unroll invariant-yield do-whiles

wsmoses:
A rotated strided-copy loop (`k = tid; do { copy(k); } while (k < n)` with the increment folded by the range analysis) arrives as an `scf.while` whose after region yields only loop-invariant values or pass-throughs of the condition-forwarded ones: every iteration past the second would repeat the second's state exactly, so a third implies the original program never terminates. Unroll it to `body(init); if (cond) body(invariant)`, selecting the forwarded results accordingly.

This is one of the normalizations behind the `bilininteg_curlcurl_pa` family raising strict in the MFEM CUDA→xla-gpu campaign (#2968). Stacked on #3006 (base `pb/buffer-normalizations`) for the preprocessing-round scaffolding.

Includes a lit test of the rotated strided-copy shape raising end to end.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3596,6 +3596,109 @@ struct AffineToStableHLORaisingPass
           AffineToStableHLORaisingPass> {
   using AffineToStableHLORaisingBase::AffineToStableHLORaisingBase;
 
+  // A rotated strided-copy loop (`k = tid; do { copy(k); } while (k < n)`
+  // with the increment folded by the range analysis) arrives as an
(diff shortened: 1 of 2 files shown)


wsmoses:
Superseded by #3071, which moves this unroll into canonicalize-parallel (so it runs before the final llvm-to-affine-access and frees allocas carried through the loop, tmop assemble diag) and fixes a soundness hole: this PR accepts any pass-through of a forwarded condition operand as invariant, so its own `@dowhile` test (`%next = %k + 16`, `scf.condition(%cond) %next`, `scf.yield %j`) is a real strided loop that gets unrolled to two iterations. #3071 requires the forwarded value itself to be invariant and the exit test to be pure over the carried state.

PR #3016, opened 2026-08-29
Raising: expand a branch between buffers read back through a pointer

wsmoses:
Stacked on #3027 (`expandBufferBranches`); merge that first.

When an `affine.if` yields one of two buffers as a memref and the result is only read back through `enzymexla.memref2pointer` → `enzymexla.pointer2memref` (the shape MFEM's `Vector`/`DeviceTensor` wrappers produce after `Memref2Pointer2MemrefCast`, #3068, has done its part but the intermediate buffer is a different rank), the branch cannot be raised: the yielded memref has no affine access to follow. This extends `expandBufferBranches` so that a memref result whose users are either direct `affine.load`/`affine.store`s, or such accesses through an m2p/p2m chain, is expanded per access — each access is cloned into a fresh `affine.if` with the arm's buffer substituted, the dead chain links are erased, and the original branch disappears.

The pointer-yield case (a branch yielding `!llvm.ptr` that is the same base as, or a gep of, the other side) is now handled generally in `canonicalize-parallel` by #3072 and is no longer part of this PR.

Test: `test/lit_tests/raising/if_memref_yield.mlir`.

Part of #2968.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -4015,17 +4015,40 @@ struct AffineToStableHLORaisingPass
           continue;
         Value thenV = thenB->getTerminator()->getOperand(i);
         Value elseV = elseB->getTerminator()->getOperand(i);
-        for (OpOperand &use : llvm::make_early_inc_range(res.getUses())) {
-          Operation *user = use.getOwner();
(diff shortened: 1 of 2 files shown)


wsmoses:
Rebased: this branch now carries #3027's expand-buffer-branches commit as its prerequisite on top of pb/buffer-normalizations (the base ref), since the pointer-yield expansion edits that function and its test needs the raw-gep raising from #3006. Merge order: #3006 → #3027 → this.

PR #3022, opened 2026-08-29
EnzymeHLOOpt: scatter to DUS only when the index is provably in bounds

wsmoses:
A scatter DROPS an out-of-bounds update while dynamic-slice and dynamic-update-slice CLAMP the start, resurrecting the write at a clamped slot (this corrupted `pa_data` slot 0 in the MFEM campaign via masked dead-lane indices).

Per review, instead of refusing known-bad index shapes, the single-index scatter-to-DUS conversion now proves its bounds before firing: a conservative signed-range walk over the index chain (constants, reshapes, widening integer converts, `clamp`, `min`/`max`, `select`, `add`/`sub`/`neg`, `mul` by constant, and range-preserving `slice`/`dynamic_slice`/`reverse`/`transpose`, computed in widened APInts) must show `0 <= index <= dim - 1`, or the scatter stays. The iota path gains the matching `start >= 0` check beside its existing `limit` check.

Two consumers encoded their invariants rather than relying on the unsound conversion:
- `scatterindexsimplify.mlir`'s dynamic-index tests now clamp their indices, so they keep exercising the conversion machinery under sound preconditions.
- The LAPACK getrf lowering writes its pivot contract (`1 <= ipiv[i] <= n`) into the IR as a clamp on the 0-indexed pivots, so the pivot-permutation scatter provably converts to the same dynamic-update-slice form as before.

Lit coverage: masked negative select, arbitrary dynamic index (the case the old guard missed), clamp-proven, constant in-bounds and out-of-bounds, a widening `ui8` convert proving a type-based bound, and an iota with negative start.

**Update:** a scatter carrying the unit attribute `enzymexla.inbounds` converts without the range analysis: whoever builds the scatter may know its index lands in bounds when the IR cannot prove it. The LAPACK lowering now marks its pivot scatter that way (LAPACK guarantees every pivot is in [1, n]) instead of clamping the pivots to make it provable; the `getrf` goldens lose the clamp. Test `scatter_dus_guard.mlir` gains the asserted case.


Pangoraw (commented):
could it rely on `utils::getBoundsFromIR`? reusing the bounds machinery and maybe improving it if needed. both codes look similar

Pangoraw (approved):

PR #3014, opened 2026-08-29
Raising: atomic adds as combining scatters

wsmoses:
An `llvm.atomicrmw fadd/add` whose result is unused converts to `memref.atomic_rmw` through the raw-gep access normalization and view rebasing, and raises as a `stablehlo.scatter` whose update region adds: adds commute (up to rounding, exactly like the atomic), so application order is irrelevant even at colliding indices, and `unique_indices` stays `false`.

This is the normalization behind `derefmat_op`/`pderefmat_op` raising strict in the MFEM CUDA→xla-gpu campaign (#2968). Stacked on #3006 (base `pb/buffer-normalizations`) for the raw-gep access machinery.

Includes a lit test of a colliding-index atomic add raising to a combining scatter.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016zErYp7upmqr4NHfhod9UD

PR #3025, opened 2026-08-30
Raising: carry atomic adds through viewed geps and raw accesses

wsmoses:
The view-rebasing and raw-gep access conversion treat `memref.atomic_rmw` like the loads and stores they already retarget, so atomics on pointers into views reach the combining-scatter raising of #3014 (whose core now targets main directly). These hunks extend functions introduced by the buffer-normalization stack, so this stacks on #3006.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -3757,6 +3757,9 @@ struct AffineToStableHLORaisingPass
             if (st.getIndices().size() == 1 &&
                 st.getValueToStore() != p2m.getResult())
               continue;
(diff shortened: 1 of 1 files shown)


wsmoses:
Closing: these hunks extend `rebaseViewedGeps`/`convertRawGepAccesses` from #3006, which was closed once the general canonicalizations (#3054, #3056, #3060, #3062, #3063) covered its TUs, so there is nothing on main for them to attach to. On main, `llvm.atomicrmw` already travels the same route as loads and stores: `llvm-to-affine-access` converts it to `enzyme.atomic_rmw`/`affine_atomic_rmw` (LLVMToAffineAccess.cpp ~L2233), and the canonicalize-parallel gep patterns accept atomic users (CanonicalizeParallel.cpp ~L522). If a sweep shows an atomics-on-views TU red, it will be refiled in that form.

PR #3049, opened 2026-09-01
Only apply auto-batching to chains that can be fully hoisted

Pangoraw:


Pangoraw:
While beneficial for memory cases, it also regresses polybench (only gemm and syrk) so i am looking at that.

PR #3051, opened 2026-09-01
reuse JIT addresses for MPI FFI handler function bindings

mofeing:
instead of separetely registering MPI function addresses for lowering to jit_call vs FFI handlers, reuse the addresses registered for JIT.

two concerns:
- can we move `EnzymeJaXMapSymbol` (and `EnzymeJaXLookupSymbol`) to Integrations/? it's a public API that we use a lot in Reactant.
- this adds a runtime cost for function lookup on FFI handlers. I expect it to be rather small compared to the actual network communication, but I would like to hear your thoughts.


romanlee:
Looks good to me

PR #3061, opened 2026-09-01
Loop canonicalizations for a pointer walked across nested loops

wsmoses:
Three loop canonicalizations that between them make the raiser's own pointer-induction pass unnecessary for the mfem translation units that needed it (#3011).

The shape comes from `kernels::AddMultAtB` (`linalg/kernels.hpp:486`), where a raw pointer is declared outside two loops and walked in the inner one:

```cpp
   TC *c = Cdata;
   for (int i = 0; i < Bwidth; ++i)
      for (int j = 0; j < Awidth; ++j)
      {
         ...
         *c += val;
         c++;
      }
```

After rotation that reaches the raiser as a pointer the enclosing loop carries, with the inner loop keeping a dead copy of it:

```mlir
%81 = scf.for %arg8 = %c1_i64 to %80 iter_args(%arg9 = %78) -> (!llvm.ptr) : i64 {
  %86 = scf.for %arg10 = %c1_i64 to %85 iter_args(%arg11 = %6) -> (!llvm.ptr) : i64 {
    %89 = llvm.getelementptr %arg9[%88]   // rooted at the enclosing iter arg
    %99 = llvm.getelementptr %89[8]
    scf.yield %99
  }
  scf.yield %86
}
```

**`ForOpFinalValueOfDeadIterArg`.** An iter arg the body never reads accumulates nothing, so the result is the yielded value of the last iteration. When that value varies with the loop only through the induction variable, it can be computed outside at the last induction variable the loop reaches, selected against the init on whether the loop ran. The slice it recomputes must be pure, region-free and under eight ops, since recomputing a whole body outside it would not be a canonicalization. Nothing about this is pointer-specific; converted do-whiles hit it with an `ub.poison` init, which is why eight blocks in `canonicalizefor/doWhile.mlir` now expect simpler output. Each was checked against its source: `@negative_step` returns -5, `@execute_once` returns 1, `@multiple_iter_args_2` returns 10 and keeps its live arg, `@test_fully_dynamic` returns `max(lb+1, ub)`.

**A stride the loop does not vary.** #3060 wanted every step of the advance defined outside the loop. That is sufficient but not necessary: mfem computes the distance in the body from values the loop does not vary, as `(max(%68,1) - 1) * 8`. The test is now whether the loop varies the value, by walking its slice back to operands from outside; the slice is recomputed where the replacement goes, since the original definition sits further down the body.

Together with #3060's induction rewrite, pointer-carrying loops in `linalg/batched/native.cpp` before raising go from 5 `scf.for` and 3 `affine.for` to none.

Measured with the raiser's `rewritePointerInduction` disabled: `fem/hybridization_ext.cpp` and `linalg/batched/native.cpp` both compile, and the full 122-TU mfem sweep is `real_red=0`. So #3011 can be closed rather than merged.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeFor.cpp
@@ -292,6 +292,34 @@ static Value castValue(PatternRewriter &rewriter, Value value, Value target,
   }
 };
 
+// The ops needed to recompute `v` where the loop's own operands are available,
+// in block order. A value the loop does not vary can still be computed inside
+// it, so being defined outside is sufficient but not necessary.
+static bool collectLoopInvariantSlice(scf::ForOp forOp, Value v,
+                                      SmallVectorImpl<Operation *> &ops) {
+  SmallVector<Value> todo{v};
+  SmallPtrSet<Operation *, 8> seen;
+  while (!todo.empty()) {
+    Value cur = todo.pop_back_val();
+    if (forOp.isDefinedOutsideOfLoop(cur))
+      continue;
+    Operation *def = cur.getDefiningOp();
+    if (!def || def->getParentRegion() != &forOp.getRegion())
+      return false;
+    if (!isMemoryEffectFree(def) || def->getNumRegions())
+      return false;
+    if (!seen.insert(def).second)
+      continue;
+    if (ops.size() >= 8)
+      return false;
+    ops.push_back(def);
+    llvm::append_range(todo, def->getOperands());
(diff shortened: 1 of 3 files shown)


wsmoses:
Split into #3062 (a stride the loop does not vary may be computed inside it) and #3063 (a dead iter arg's result is its last yielded value), each on main with its own tests. The mfem result -- both translation units compiling with the raiser's pointer induction disabled, and the 122-TU sweep at real_red=0 -- needs the pair, and is recorded in both.

PR #3071, opened 2026-09-02
canonicalize-parallel: unroll whiles that cannot iterate a third time

wsmoses:
Part of #2968.

A `scf.while` cannot iterate a third time in two cases, and then unrolls (under the forward-progress guarantee) to

```
body(init); if (cond) { after; body(next) }
```

1. The exit test is a pure function of carried state whose next-iteration values are loop-invariant: a third iteration would start from exactly the second's state and take the same exit decision, so the loop would never terminate. Only the state the exit test reads needs invariant yields; an accumulator can be carried alongside (`bilininteg_curlcurl_pa.cpp`'s `scf.while (%k, %acc)` sums with `%k` yielded as a constant and `%acc` passed through).
2. The bits known of the second iteration's state decide the exit test. They are computed with `llvm::KnownBits` from the initial operands through the first iteration's condition operands and yields, and through the `arith` ops, `arith.select`s and `scf.if`s on the way (an `scf.if` with an unknown condition intersects its two yields).

MFEM's strided copy loops (`k = tid; do { copy(k); k += stride; } while (k < n)`) reach us the first way once the range analysis folds the increment to a constant, e.g. in `bilininteg_curlcurl_pa.cpp`:

```
scf.while (%arg23 = %102) { ... %113 = arith.cmpi ult, %arg23, %80; scf.condition(%113) } do { scf.yield %c4_i32 }
```

The second way covers clang's done-flag loops, e.g. `tmop_pa_h3d.cpp` (assemble diag):

```
scf.while (%arg13 = %true, %arg14 = %59, %arg15 = %58) { ... scf.condition(%arg13) } do { scf.yield %false, %61, %62 }
```

the rotated single-trip loop around `bilininteg_batchitrans.cpp`'s team solve, whose trip flag is the result of an `scf.if` on the previous flag and is yielded `false`, and the reduction over a leading dimension of one or two in `bilininteg_vectorfemass_pa.cpp`:

```
scf.while (%acc = %cst, %k = %c0_i32) { ... %next = arith.addi %k, %c1; %bits = arith.ori %k, %odd; %again = arith.cmpi eq, %bits, %c0; scf.condition(%again) %sum, %next } do { scf.yield }
```

where the second iteration's `%k` is 1, so `%k | %odd` has its low bit set whatever `%odd` is.

The `tmop_pa_h3d.cpp` loop carries an `llvm.alloca` (`%58`, a nested `mfem::future::tensor` struct) through the loop as an iter arg, which keeps `ConvertLLVMAllocaToMemrefAlloca` from converting it; that TU then fails raising on the surviving `llvm.alloca`. Running the unroll as a `canonicalize-parallel` pattern, i.e. before the final `llvm-to-affine-access`, lets the alloca convert normally (with #3064's gep fold for the 8-byte offset). Replaying that TU's imported module through the full pipeline with this branch, the `llvm.alloca` failure is gone and raising proceeds to the next (independent) blocker, a lane-private `memref<9xf64>` scratch (#2996's scope); the union's `flattenLLVMArrayAllocas` helper is not needed for it.

This supersedes #3013's raiser-local `unrollInvariantYieldWhiles`, with one soundness correction: #3013 accepted any pass-through of a value forwarded by `scf.condition` as invariant, and its own `@dowhile` test (`%next = %k + 16`, `scf.condition(%cond) %next`, `scf.yield %j`) is a genuine multi-iteration loop that it unrolled to two iterations. Here after-block arguments resolve through the forwarded condition operands, and the exit condition must be pure over loop-external values and the before-block arguments (no memory reads, since the body may write what the test reads). The negative cases in the test cover the varying pass-through against an unknown bound, a condition that loads memory, a yield computed in the after region from the carried state, an exit test that reads the accumulator, a trip flag yielded `true`, and a masked (`andi`) counter whose second value is not decided.

Test: `test/lit_tests/canonicalize_parallel_while_unroll.mlir`.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
@@ -19,6 +19,7 @@
 #include "Enzyme/MLIR/Dialect/Ops.h"
 #include "Enzyme/MLIR/Interfaces/AutoDiffOpInterface.h"
 #include "mlir/Analysis/DataLayoutAnalysis.h"
+#include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
@@ -32,6 +33,7 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "src/enzyme_ad/jax/Dialect/Ops.h"
 #include "src/enzyme_ad/jax/Passes/Passes.h"
+#include "llvm/Support/KnownBits.h"
 
 namespace mlir {
 namespace enzyme {
@@ -41,6 +43,7 @@ namespace enzyme {
 } // namespace mlir
 
 using namespace mlir;
+using llvm::KnownBits;
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
   }
 };
 
+// A loop that cannot iterate a third time unrolls to
+// body(init); if (cond) { after; body(next) }. That holds when the exit test
```
ftynse: There's no "body" in the while loop, only "before" and "after" regions.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+// is a pure function of carried state whose next-iteration values are
+// loop-invariant: the third iteration would start from exactly the second's
+// state and take the same exit decision, so the loop would never end, and
+// under the forward-progress guarantee it runs at most twice. It also holds
```
ftynse: this explanation makes little sense; the "before" block runs at most twice, the "after" block runs at most once in this case

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+// when the bits known of the second iteration's state decide the exit test.
+// MFEM's strided copy loops (`k = tid; do { copy(k); k += stride; } while (k
+// < n)`) arrive the first way once the range analysis folds the increment to
+// a constant; a rotated single-trip loop around a team solve arrives the
+// second with its trip flag yielded as a constant, and a reduction over a
+// leading dimension of one or two (`do { .. } while ((k++ | odd) == 0)`) with
+// its counter's low bit set.
```
ftynse: We don't care about these history lessons. This specific one is also incorrect or incomplete as it's unclear why `stride` being constant anyhow change the behavior unless we have both stride and `n` as constants.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+struct UnrollInvariantYieldWhile : public OpRewritePattern<scf::WhileOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(scf::WhileOp w,
```
ftynse: Using single-letter names for long-lived variables is a terrible idea

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+                                PatternRewriter &rewriter) const override {
+    Block &before = w.getBefore().front();
+    Block &after = w.getAfter().front();
+    auto condOp = cast<scf::ConditionOp>(before.getTerminator());
+    auto yield = cast<scf::YieldOp>(after.getTerminator());
+
+    // The carried state a value is a pure function of, together with values
+    // from outside the loop; nullopt when it is not such a function.
+    auto stateOf =
```
ftynse: Why this is not a `getBackwardSlice`?  Also `stateOf` is a terrible name, use names with terminology already used in MLIR.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+        Operation *d = v.getDefiningOp();
+        if (!w->isAncestor(d))
+          continue;
+        if (d->getNumRegions() || !isMemoryEffectFree(d))
+          return std::nullopt;
+        stack.append(d->operand_begin(), d->operand_end());
+      }
+      return state;
+    };
+    auto invariant = [&](Value v) {
```
ftynse: Nit: isInvariant

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+      auto state = stateOf(v);
```
ftynse: If we keep `stateOf`, it may as well stop at the first value added to the vector. We don't actually care about the value, just the fact that it exists. `stateOf` is only ever used in combination with `invariant`.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+    DenseMap<std::pair<Value, unsigned>, Known> memo;
+    auto widthOf = [](Type t) -> unsigned {
+      if (auto it = dyn_cast<IntegerType>(t))
+        return it.getWidth();
+      if (isa<IndexType>(t))
+        return IndexType::kInternalStorageBitWidth;
+      return 0;
+    };
+    std::function<Known(Value, unsigned)> knownIn =
+        [&](Value v, unsigned iter) -> Known {
```
ftynse: This needs some concise documentation

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+      if (!w->isAncestor(ba ? ba.getOwner()->getParentOp() : d))
+        iter = 0;
+      auto key = std::make_pair(v, iter);
+      auto it = memo.find(key);
+      if (it != memo.end())
+        return it->second;
+      Known res = KnownBits(width);
+      APInt c;
+      auto operand = [&](unsigned i) {
+        return *knownIn(d->getOperand(i), iter);
```
ftynse: Recursion on use-def chains is forbidden

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+        Known kt = knownIn(t, iter), kf = knownIn(f, iter);
+        if (!kt || !kf)
+          return std::nullopt;
+        return kt->intersectWith(*kf);
+      };
+      if (ba) {
+        if (ba.getOwner() == &before) {
+          if (iter == 1)
+            res = knownIn(w.getInits()[ba.getArgNumber()], 0);
+          else if (iter == 2)
```
ftynse: This must assert on all other values of `iter`. It seems we can enter this with `iter==0` and nothing happens, weird

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -748,6 +750,245 @@ struct FlattenAggregateAlloca : public OpRewritePattern<memref::AllocaOp> {
...
+    Location loc = w.getLoc();
+    IRMapping m1;
+    for (auto [ba, init] : llvm::zip(before.getArguments(), w.getInits()))
+      m1.map(ba, init);
+    for (Operation &op : before.without_terminator())
+      rewriter.clone(op, m1);
+    Value cond1 = m1.lookupOrDefault(condOp.getCondition());
+    SmallVector<Value> fwd1;
+    for (Value v : condOp.getArgs())
+      fwd1.push_back(m1.lookupOrDefault(v));
```
ftynse: OrDefault will put nulls in the vector, this doesn't sound right

ftynse (commented):
No idea why this belongs to `CanonicalizeParallel.cpp`, it's an scf.while transformation.

We should see if it can be obtained as side effect of other opts: can we rotate a while loop so there's only the before or only the after part? can we convert such a loop to `scf.for` and then use the available optimizations to remove zero-iteration and inline single-iteration `scf.for`? Can we reuse integer range analysis and constant propagation passes (though `KnownBits` may be more powerful, it needs a justification)?

Briefly document design decisions but avoid multi-paragraph essays nobody will read.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Rewritten in terms of the before and after regions: before runs at most twice, after at most once.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Reworded: the loop unrolls to before(inits); if (cond) { after; before(yields) }, and the comment now says why the second evaluation of the test is decisive.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Dropped.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Renamed throughout (whileOp, conditionOp, yieldOp, firstEvaluation/secondEvaluation, definingOp, blockArg, ...).

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Now built on getBackwardSlice, filtered to the loop, and named carriedArguments: the loop-carried arguments a value depends on, with an after-region argument standing for the condition operand forwarded to it, which is the one step the slice cannot take by itself.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Done.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: carriedArguments is now used both for its set (the condition's carried arguments, whose yields are checked) and through isInvariant; kept as one function for that reason.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Documented: evaluation 0 is outside the loop, 1 the first evaluation of the before region (fed by the inits), 2 the second (fed by the first's yields); nullopt for a non-integer value.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Made iterative: a worklist over (value, evaluation) keys with a memo, computing a key once its operand keys are known.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Asserted: a before-region argument is only ever asked at evaluations 1 and 2, and a value from outside the loop is answered before the operand lookup (its bits are known only when it is a constant).

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: IRMapping::lookupOrDefault returns the value itself when it is not mapped (the values from outside the loop), not null; lookupOrNull is the variant that would. Kept.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
ftynse: But is it a good idea to build on getBackwardSlice? More efficient? More robust? I needed an answer/explanation for why/why not rather than a blind implementation.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
ftynse: it can still take a flag and stop immediately after pushing one element to `carried`, no need to burn cycles when we already know the answer.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -766,6 +769,328 @@ struct StoreOfUndef
...
+
+  LogicalResult matchAndRewrite(scf::WhileOp whileOp,
+                                PatternRewriter &rewriter) const override {
+    Block &before = whileOp.getBefore().front();
+    Block &after = whileOp.getAfter().front();
+    auto conditionOp = cast<scf::ConditionOp>(before.getTerminator());
+    auto yieldOp = cast<scf::YieldOp>(after.getTerminator());
+
+    // The loop-carried arguments a value is a pure function of, together
+    // with values from outside the loop; nullopt when the dependence passes
```
ftynse: "together with values from outside the loop" is misleading, the function returns only block arguments.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -766,6 +769,328 @@ struct StoreOfUndef
...
+      return carried;
+    };
+    auto isInvariant = [&](Value value) {
+      auto carried = carriedArguments(value);
+      return carried && carried->empty();
+    };
+
+    // The bits known of an integer value in one evaluation of the loop:
+    // evaluation 0 is outside the loop, 1 the first evaluation of the before
+    // region (fed by the inits), 2 the second (fed by the first's yields).
```
ftynse: used named enum entries intead of magic numbers with unsigned

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -766,6 +769,328 @@ struct StoreOfUndef
...
+      Operation *definingOp = value.getDefiningOp();
+      if (!whileOp->isAncestor(blockArg ? blockArg.getOwner()->getParentOp()
+                                        : definingOp))
+        evaluation = 0;
+      return {value, evaluation};
+    };
+    // The values, with their evaluations, the bits of a key inside the loop
+    // follow from.
+    auto operandKeys = [&](Key key) -> SmallVector<Key> {
+      auto [value, evaluation] = key;
```
ftynse: Let's turn these into free functions instead of the long chain of nested lambdas.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -766,6 +769,328 @@ struct StoreOfUndef
...
+    auto keyOf = [&](Value value, unsigned evaluation) -> Key {
+      auto blockArg = dyn_cast<BlockArgument>(value);
+      Operation *definingOp = value.getDefiningOp();
+      if (!whileOp->isAncestor(blockArg ? blockArg.getOwner()->getParentOp()
+                                        : definingOp))
+        evaluation = 0;
+      return {value, evaluation};
+    };
+    // The values, with their evaluations, the bits of a key inside the loop
+    // follow from.
```
ftynse: Use plain, easily understandable grammar, not a multi-part sentence ending with a preposition.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
```diff
@@ -766,6 +769,328 @@ struct StoreOfUndef
...
+      if (matchPattern(value, m_ConstantInt(&constant)))
+        return KnownBits::makeConstant(constant);
+      if (llvm::any_of(operands, [](const Known &bits) { return !bits; }))
+        return KnownBits(width);
+      Operation *definingOp = value.getDefiningOp();
+      if (isa<scf::IfOp, arith::SelectOp>(definingOp)) {
+        const KnownBits &condition = *operands[0];
+        if (condition.isConstant())
+          return operands[condition.isZero() ? 2 : 1];
+        return operands[1]->intersectWith(*operands[2]);
```
ftynse: The naming is misleading here for scf.if: the operands are not really operands, but values yielded by `scf.yield` from both regions. This should be reflected in proper names here and elsewhere to avoid confusion.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
ftynse: just use an enum as suggested elsewhere

ftynse (approved):

ftynse:
This is okay for me, but I'd still want an answer/justificaiton for:

> We should see if it can be obtained as side effect of other opts: can we rotate a while loop so there's only the before or only the after part? can we convert such a loop to scf.for and then use the available optimizations to remove zero-iteration and inline single-iteration scf.for? Can we reuse integer range analysis and constant propagation passes (though KnownBits may be more powerful, it needs a justification)?

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: It is neither faster nor more robust than a hand-written walk; it is the same transitive closure over use-def edges. The reason to use it is that the question here is exactly what the utility answers, "every operation the value transitively depends on", filtered to the loop, and a reader recognizes it without reading a second worklist. The pattern still needs its own walk on top for the two things a slice does not do: read the block arguments among the sliced operands, and continue from an after-region argument at the condition operand forwarded to it. Cost is the size of the dependence cone either way.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Done: `anyIsEnough` returns at the first carried argument, and `isInvariant` passes it.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Reworded: it returns the loop-carried arguments the value is a pure function of, nullopt when the dependence passes through a region or a side effect.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: `enum class Evaluation { Outside, First, Second }` throughout; the asserts are on the enum.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: `enum class Evaluation { Outside, First, Second }` throughout; the asserts are on the enum.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: The lambdas are now free functions taking the loop: `carriedArguments`, `isInvariant`, `knownKey`, `knownInputKeys`, `knownBitsFrom`, `knownBitsIn`; the memo map is passed in.

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Reworded: "The keys whose bits decide the bits of a key inside the loop."

src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
wsmoses: Renamed: the function takes `inputs`, and for an scf.if or select they are read as `condition`, `thenBits`, `elseBits`, with the comment saying they are the values the two regions yield.

wsmoses:
On the alternatives:

- Rotating the loop does not change what is unknown. These loops are already in the shape where the exit test sits at the top of `before` and all the work in `after` (the do-while MFEM writes); the missing fact is the trip count, and rotation moves code between the regions without adding any.
- `scf.for` is not reachable for them. `scf-uplift-while-to-for` needs an induction variable updated as `iv + step` and compared against a bound. The strided-copy loop yields a constant (`scf.yield %c4_i32`) rather than `iv + step` once the range fold has rewritten the increment, the done-flag loops (`tmop_pa_h3d.cpp`, the rotated single-trip loop in `bilininteg_batchitrans.cpp`) carry an `i1` and no counter, and the `vectorfemass` reduction exits on `(k | odd) == 0`. None of them is a counted loop, so the zero- and single-iteration folds of `scf.for` never get a chance.
- SCCP and integer range analysis merge every evaluation of a block argument into one lattice value: the `before` argument joins the init with the yield. For `k = tid; ... yield 4` that join is unknown, and for the done flag it is `{true, false}`, so neither analysis can decide the exit test. What is needed is the value in the second evaluation specifically, which is what `knownBitsIn(value, Evaluation::Second)` computes: an unrolled, evaluation-indexed query rather than a fixpoint over the loop.
- Known bits versus ranges: the flag and the constant-yield cases would work with ranges as well. The reduction case reads the low bit of `k | odd` with `k = 1` in the second evaluation and `odd` unknown, which is a bit fact, not an interval fact, and it is what decides `bilininteg_vectorfemass_pa.cpp`. The arithmetic covered is the dozen `arith` ops plus `select`/`scf.if`, so reusing `llvm::KnownBits` costs one function; a range version would need the same evaluation-indexed driver around a different transfer function.

PR #3081, opened 2026-09-03
Fold a null-armed select per use

wsmoses:
Follow-up to #3080. `SelectOfNullPointer` folds `select(c, p, null)` to `p` only when every use of the select is a dereference, so one use that can observe the address keeps every other use on the select too. In mfem's `bilininteg_dgdiffusion_pa.cpp` the `DeviceTensor` pointer

```
%296 = arith.select %295, %285, %2 : !llvm.ptr        // %2 = llvm.mlir.zero
affine.store %296, %332[40] : memref<?x!llvm.ptr>      // lambda capture struct
%298 = llvm.getelementptr %296[%i] ...
"llvm.intr.memset"(%299, %c0_i8, %c24_i64)
```

is both stored into the `forall` lambda's capture struct (the host loop copy of the body, read later at dynamic indices) and memset through a gep. The stored copy blocks the fold, the raiser then distributes the gep over the select, and the wrapper fails on `gep(llvm.mlir.zero)`.

Decide per use instead. `dereferencesOrDerives` answers for one operand: a load, a store through it, an atomic, a memory intrinsic, or a carry into a derived value (gep base, cast, select arm, view, yield out of an `if`) whose own uses are then checked. Every use of the select that reaches only dereferences is rewritten to the real arm; the select survives only for the uses that can observe the address, and is erased once none remain. `onlyDereferenced` is the same walk over all uses, so `IfOfNullPointer` is unchanged.

The test covers a select that is stored as a value and memset through geps (the memset chain takes the pointer, the store keeps the select), a store through it plus a view plus an `icmp` (the accesses take the pointer, the compare keeps the select), a yield out of an `scf.if` feeding a store plus an escaping call, and a select that only escapes (untouched).

Part of #2968.

### DIFF
--- a/src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
+++ b/src/enzyme_ad/jax/Passes/CanonicalizeParallel.cpp
@@ -503,10 +503,67 @@ struct SelectOfDifferentBaseGEPs : public OpRewritePattern<arith::SelectOp> {
   }
 };
 
-// Whether nothing can observe the address itself: every use either reads or
-// writes the pointed-to memory, or carries the pointer somewhere that is in
-// turn only dereferenced. A comparison, an escape into a call, or a store of
-// the pointer as a value all answer no.
+// Whether `use` reads or writes the pointed-to memory, or carries the pointer
+// into a value (appended to `derived`) whose uses decide instead. A
+// comparison, an escape into a call, or a store of the pointer as a value all
+// answer no.
+static bool dereferencesOrDerives(OpOperand &use,
+                                  SmallVectorImpl<Value> &derived) {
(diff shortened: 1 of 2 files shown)


wsmoses:
CI's failure in `@yielded_and_escaping` was the test, not the fold: the branch only yielded values defined outside it, so upstream's if-to-select canonicalization turned the yield into a select and `LoadSelect` distributed the load into nested `scf.if`s of loads before the per-use fold saw the yield. The branch now yields a GEP computed inside it, which keeps the `scf.if` (no if-to-select, no `LoadSelect`), so the test exercises the yield path it was written for: the GEP's base takes the real pointer, the call keeps the select. Verified to pass with this branch and fail on main.

PR #3092, opened 2026-09-03
Unroll for loops whose trip count the loop bounds decide

wsmoses:
An `scf.for` or `affine.for` whose bounds prove exactly one trip on every point of the enclosing domain inlines its body at the lower bound (a strided-copy remainder loop whose extent the propagated block size made constant); one that provably never runs folds to its inits. The bounds and step compose into the block's `LoopDomain`, failing (no fold) when a bound does not compose through; the `affine.for` lower bound materializes as the operand itself when the map is a bare dim or symbol, otherwise as an `affine.apply`, or an `affine.max` for several results.

Registered in `populateAffineExprSimplificationPatterns`, so affine-cfg and simplify-affine-exprs both apply it.

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -1512,6 +1512,143 @@ decideOnDomain(LoopDomain &domain, arith::CmpIPredicate pred, Value lhs,
   return failure();
 }
 
+// The bounds of a for loop as expressions over `domain`: the loop runs while
+// every lower bound is below every upper bound, stepping by `step`.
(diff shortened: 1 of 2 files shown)


wsmoses:
Rebased atop main (after #3093/#3095: the decision helper is now the shared `decideOnDomain`, and the patterns register after `FoldNeverLoopingSCFWhileCondition`).

Also fixed a crash the MFEM sweep exposed: when the single-trip body yields its own induction variable or an iteration argument, the yield operands are body block arguments, which `inlineBlockBefore` replaces before `replaceOp` read them. They are now mapped through the argument replacements first; `@yield_args` in `loop_bound_unroll.mlir` covers it. Full lit suite green locally.

PR #3093, opened 2026-09-03
Fold an scf.while condition that is false on its first evaluation

wsmoses:
An `scf.while`'s condition is evaluated first with the inits as the before arguments, and only again if it held. So when the enclosing loop bounds show it false on the inits, it is false whenever evaluated: `FoldNeverLoopingSCFWhileCondition` replaces the `scf.condition` operand with `false`, and the generic `InlineNeverLoopingWhile` of #3096 turns the while into one execution of its before region. This is how a rotated do-while remainder loop that the propagated block size made single-shot disappears.

The decision is `decideOnDomain`, factored out of `FoldCmpUsingLoopBounds`: the comparison's operands compose into the block's `LoopDomain`, here with the before arguments substituted by the inits; an unsigned predicate is rewritten signed once both sides are proven non-negative. Registered in `populateAffineExprSimplificationPatterns`, so affine-cfg and simplify-affine-exprs both apply it.

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -9,6 +9,7 @@
 #include "mlir/Dialect/Affine/IR/AffineValueMap.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/IntegerSet.h"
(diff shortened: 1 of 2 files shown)


wsmoses:
Split per the discussion on #2946: the generic half — an `scf.while` whose `scf.condition` is a constant `false` is one execution of its before region — is `InlineNeverLoopingWhile` in CanonicalizeFor (#3096, on `main`, also imported into `populateAffineExprSimplificationPatterns`). What stays here is the domain-specific part: proving the condition false on its first evaluation, which needs the before arguments read as the inits (`decideOnDomain` with a substitution, factored out of `FoldCmpUsingLoopBounds`), and then only rewriting the `scf.condition` operand to `false`. Rebased onto `main` now that #2946 is in; carries #3096's commit until it merges.

PR #3094, opened 2026-09-03
affine-cfg: split a loop or guard on an affine.if of constants

wsmoses:
An `affine.if` over dimensions yielding constants is a select no affine expression writes — the sign-extended `c != 0` of LDV's `D1D + (c != 0)` bound once #3090 raises it — so an `scf.for` bounded by it or an `scf.if` guarded by it never raised. `SplitOnAffineIfConstants<scf::ForOp | scf::IfOp>` puts a copy of the op under each branch of the conditional with the yielded constant in place of the result, recomputing what derives from it through pure region-free ops, and each copy raises. Only the operands the raising needs (bounds and step, or the condition) are traced; a conditional whose set operands are all symbols is a symbol itself and is left alone.

With #3090 on top, the `d + extsi(c != k)` guards from LDV (`c` a parallel IV in `[0, 2)` and `[0, 3)`) raise to nested `affine.if`s with the index folded under each branch:

```mlir
affine.parallel (%t, %c) = (0, 0) to (8, 2) {
  affine.if #set(%c) {                 // c >= 1
    affine.if #set3(%t)[%d] {          // t <= d - 2
      affine.store %cst, %out[%t + 8]
    }
  } else {
    affine.if #set2(%t)[%d] {          // t <= d - 1
      affine.store %cst, %out[%t]
    }
  }
}
```

Test: `affinecfg_split_if_constants.mlir`. Full lit suite 1289/1289 locally. Independent of #3090 (the test starts from the `affine.if`).

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineCFG.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineCFG.cpp
@@ -2981,6 +2981,12 @@ struct SplitOnAffineIfConstants : public OpRewritePattern<OpTy> {
     Region *scope = getLocalAffineScope(op);
     if (!scope)
       return failure();
+    // Only scalar results are yielded through the affine.if as a select;
+    // a pointer or memref result would leave one no raising can select on.
+    if (!llvm::all_of(op->getResultTypes(), [](Type type) {
+          return isa<IntegerType, IndexType, FloatType>(type);
+        }))
+      return failure();
     affine::AffineIfOp ifOp = findConditional(op, scope);
     if (!ifOp)
       return failure();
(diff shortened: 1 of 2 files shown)


wsmoses:
Amended: the split now only applies to ops with scalar results (`isa<IntegerType, IndexType, FloatType>`). The clone under each branch of an `affine.if` yields the op's results, and an `scf.if` yielding a pointer out of arms with effects would become an `affine.if` yielding that pointer, which nothing downstream selects on; scalar results select as an `affine.if` does. Added the `@pointer_keep` case to the test.

On MFEM this PR alone turned `bilininteg_mixedvecgrad_pa`, `bilininteg_hcurl_kernels` and `bilininteg_hdiv_kernels` red; the cause was not the split itself but `CombineAffineIfs` merging the new guarded `affine.if` into an adjacent pure pointer multiplex on the same set. That is fixed independently in #3098; with both, the three raise again and the suite is at parity.

PR #3095, opened 2026-09-03
Fold a bitwise-or compared against a power of two using loop bounds

wsmoses:
`(a | b) ult 2^k` holds exactly when every operand is under `2^k`, and `(a | b) uge 2^k` when any operand is: bitwise-or sets a bit at or above position `k` exactly when some operand does. The or itself is not affine, so `FoldCmpUsingLoopBounds` cannot see through it; `FoldOrCmpUsingLoopBounds` instead asks the block's `LoopDomain` about each operand's own comparison against `2^k` (`decideOnDomain` of #3093, now taking the predicate and operands so a comparison that exists only as an operand and a bound can be asked):

- under `ult`, an operand decided under `2^k` drops out of the or; one decided at or above it settles the comparison `false`;
- under `uge`, an operand decided at or above `2^k` settles it `true`; one decided under drops out.

Nothing is rewritten when no operand is decided. The remaining operands are re-or'ed and compared as before (`(tid | b) ult 4` under `tid < 4` becomes `b ult 4`, which is affine and may itself fold).

Test: `or_cmp_pow2.mlir` under `--affine-cfg`. Full lit suite green locally.

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -1469,16 +1469,16 @@ struct FoldMinMaxUsingLoopBounds : public OpRewritePattern<MinMaxOp> {
   }
 };
 
-// The value `cmp` has on every point of `domain`, its operands read through
-// `substitution`; fails when the domain does not decide it. Values are
+// The value `lhs pred rhs` has on every point of `domain`, the operands read
+// through `substitution`; fails when the domain does not decide it. Values are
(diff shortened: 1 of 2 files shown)


wsmoses:
Reworked: no forced expansion. Each operand's comparison against `2^k` is put to `decideOnDomain`; a decided operand is dropped from the or (`ult`: decided under; `uge`: decided at or above) or settles the whole comparison (the opposite decision), and undecided operands stay in an or that is compared as before. `decideOnDomain` is factored out of `FoldCmpUsingLoopBounds` here (#3093 will rebase onto it and add its init substitution).

PR #3100, opened 2026-09-04
implement JIT lowering for `comm.mpi` ops

mofeing:
this PR effectively makes the transition from legacy EnzymeXLA MPI ops to the new `comm.mpi` ops, at least for the CPU path. NCCL will come soon after this PR is merged.

this new design features:
- MPI ops no longer return a `int` that requires checking. i'm working on throwing a XLA error just like the FFI handlers.
  - note that we were already ignoring error codes in the EnzymeXLA MPI ops
- MPI communicators are now values, so we are no longer restricted to working only with `MPI_COMM_WORLD`.
- `comm.mpi.waitall` accepts varargs, which should help with JIT fusion by avoiding intermediate `stablehlo.concatenate`.
- LLVM wrapper for JIT is no longer monomorphised / specialized to MPI datatype, buffer shape or other MPI constants.
  - instead, all required info is emitted as `stablehlo.constant` and passed as args of `enzymexla.jit_call`.
  - this should reduce compile time spent on JIT.
  - ... and also should allow us to support dynamic shapes in the future.
  - the only exception is `comm.mpi.waitall` since we need varargs, but i'm already experimenting with a version that uses va_list.

the only current problem is that since MPI constants are queried during lowering (and not during JIT), the lowering fails if the MPI constants are not registered. this is the case of lit tests on `comm.mpi.send/isend/recv/irecv/allreduce/bcast` as `enzymexlamlir-opt` doesn't register such symbols. i'm working on a solution for that at the same time.


src/enzyme_ad/jax/Passes/Comm/Passes.td
```diff
@@ -11,4 +11,14 @@ def LowerCommToStablehloPass : Pass<"lower-comm-to-stablehlo"> {
   ];
 }
 
+def LowerCommToJITPass : Pass<"lower-comm-to-jit"> {
+  let summary = "Lower Comm ops to `enzymexla.jit_call`";
+  let dependentDialects = [
+    "comm::CommDialect",
+    "stablehlo::StablehloDialect",
+    "enzymexla::EnzymeXLADialect",
```
romanlee: Why do we need each of these here?

test/lit_tests/comm/lower-comm-to-stablehlo/mpi_allreduce.mlir
romanlee, line 1: Looks like you moved all the other to-stablehlo tests to lower-comm. WHy is this one still here?

src/enzyme_ad/jax/Passes/Comm/LowerCommToJIT.cpp
```diff
@@ -0,0 +1,1677 @@
...
+
+extern "C" void *EnzymeJaXLookupSymbol(const char *name);
+
+const char *convertMlirTypeToMpiDatatypeName(Type type,
+                                             bool allow_cast = false) {
+  // case ffi::DataType::INVALID: return nullptr;
+  if (type.isInteger(1)) /*ffi::DataType::PRED:*/
+    return "MPI_C_BOOL";
+  // case ffi::DataType::S1: return nullptr;
+  // case ffi::DataType::S2: return nullptr;
```
romanlee, line 29: Are you sure we want to do it this way as opposed to the existing EnzymeXLA_MPIDatatype EnumAttr approach?

This way is more elegant for sure, but maybe the advantage of the other way is that we leave the raw datatype to mpi datatype mapping to the frontend (whether that's julia or something else in the future). Doing the mapping at the Julia-level in the Reactant case for example allows us to be more confident that the types map as the user expects. Not sure if bad mapping would actually be a concern, but just a thought.

Also using the enum was originally @wsmoses's idea, so curious what his thoghts are on changing that

test/lit_tests/comm/lower-comm/mpi_comm_rank.mlir
```diff
@@ -0,0 +1,25 @@
...
+// JIT-LABEL: llvm.func @enzymexla_jitwrap_MPI_Comm_rank
+// JIT-SAME:                                            (%arg0: !llvm.ptr, %arg1: !llvm.ptr) {
+// JIT-NEXT:   %0 = llvm.load %arg1 : !llvm.ptr -> !llvm.ptr
+// JIT-NEXT:   %1 = llvm.call @MPI_Comm_rank(%0, %arg0) : (!llvm.ptr, !llvm.ptr) -> i32
+// JIT-NEXT:   llvm.return
+// JIT-NEXT: }
+// JIT-LABEL: func.func @main
+// JIT-SAME:                 (%[[COMM:.*]]: tensor<i64>) -> tensor<i32> {
+// JIT-NEXT: %[[rank_pl:.*]] = stablehlo.constant dense<-1> : tensor<i32>
+// JIT-NEXT: %[[v0:.*]] = enzymexla.jit_call @enzymexla_jitwrap_MPI_Comm_rank (%[[COMM]], %[[rank_pl]]) {output_operand_aliases = [#stablehlo.output_operand_alias<output_tuple_indices = [], operand_index = 1, operand_tuple_indices = []>], xla_side_effect_free} : (tensor<i64>, tensor<i32>) -> tensor<i32>
```
romanlee, line 24: Typo? You pass the communicator as the second arg to `@MPI_Comm_rank`, I think it should be the first

test/lit_tests/comm/lower-comm-to-stablehlo/mpi_allreduce.mlir
mofeing: ups, removing it now.

test/lit_tests/comm/lower-comm/mpi_comm_rank.mlir
mofeing: good catch. fixing it now. it should be `@MPI_Comm_rank(%0, %arg1)` because `%arg1` is the pointer to the rank.

mofeing:
#3115 should fix the issue for testing JIT lowering of ops with datatype and MPI op arguments.

src/enzyme_ad/jax/Passes/Comm/LowerCommToJIT.cpp
mofeing: > This way is more elegant for sure, but maybe the advantage of the other way is that we leave the raw datatype to mpi datatype mapping to the frontend (whether that's julia or something else in the future).
> Doing the mapping at the Julia-level in the Reactant case for example allows us to be more confident that the types map as the user expects. Not sure if bad mapping would actually be a concern, but just a thought.

I wouldn't leave that up to the frontend. In the case that you want another type, the frontend should then emit `stablehlo.convert` or `stablehlo.bitcast_convert`.

The MPI_Datatype is not only related to the bitsize of the elements, but also to the operations (like reductions on `MPI_Reduce` or `MPI_Allreduce`). If you mismatch the MPI_Datatype and the MLIR type on reduction op, the data will be corrupted. This is also why this method has the `allow_cast` argument, which states whether it is safe to bitcast to another type (e.g. `bf16` to `i16`) on regular communications.

Also, the number of MLIR element types we deal with (`i32`, `f64`, `complex<f32>`, ...) is fixed by the MLIR side. You're not gonna be able to add new datatypes from the frontend because StableHLO/XLA is not gonna understand it.

The way I see it, you gain nothing from the fronted side, and you avoid messing it up by having it here.

mofeing: Something that we might need to do in the frontend (or here if we manage to push for a little "runtime") is to register custom `MPI_Datatype`s with `MPI_Register_datarep` for the MLIR element types that do not have a equivalent in MPI (e.g. `bf16`, quantised types, ...).

romanlee: Hm, yeah ok I think this makes sense. I guess the question is basically: who's a better judge of what the actual data layout is, the frontend user (eg julia) who requests a buffer of a certain type, or the backend (xla/mlir) who actually allocates the buffer and stores the data. And probably the answer is the backend — ie, if the actual layout doesn't match what the user expected, it might still work if you leave the datatype determination to the backend, but it definitely won't work if you leave it up to the front end.



Separately, it seems like you're missing a bunch of datatypes. E.g. openmpi defines way more than you have here. https://github.com/open-mpi/ompi/blob/a06f1c1a4b6212729da406de9f45ce357a5dac7e/ompi/datatype/ompi_datatype_internal.h#L40. What's your thinking here?



Also separately, could you explain why allreduce isn't allowed to cast, but the other ops are?

src/enzyme_ad/jax/Passes/Comm/LowerCommToJIT.cpp
```diff
@@ -0,0 +1,1677 @@
...
+    } else if (auto attr = cast<comm::MpiOpAttr>(value_attr)) {
+      name = comm::stringifyMpiOpEnum(attr.getValue());
+    } else {
+      return rewriter.notifyMatchFailure(
+          op, "MPI constant is not a valid attribute");
+    }
+
+    uint64_t value;
+    int found =
+        EnzymeJaXLookupSymbol(name.data(), reinterpret_cast<void **>(&value));
```
romanlee: Why do we have to do this here? It already happens in lowerjitpass. Seems redundant. Then you also probably don't need https://github.com/EnzymeAD/Enzyme-JAX/pull/3115 right?

mofeing: they are related, but are still different things. I'm avoiding specialisation of LLVM IR as much as possible so that we can use a single LLVM wrapper for all argument combinations and reduce compile time.

instead of doing `llvm.mlir.addressof` inside the LLVM wrapper for `jit_call`, we need to lookup the symbol during the lowering (and not during JIT), inject a `stablehlo.constant` with the value, and we pass it as a argument. this also has the benefit that switching from statically known values to runtime known values is trivial.

the problem is that by doing it this way, lowering fails on lit_tests because `enzymexlamlir-opt` doesn't initialise JIT nor registers the MPI symbols (so the `EnzymeJaxLookupSymbol` fails). #3115 solves it by adding a pass that (1) initialises the JIT and (2) registers the symbols with mock values.

src/enzyme_ad/jax/Passes/Comm/LowerCommToJIT.cpp
mofeing: note that the data layout is **already** set by MLIR/StableHLO/XLA, and that it is the frontend, Reactant, who adheres to it (e.g. MLIR is row-major, Julia is column-major, for which Reactant introduces `stablehlo.transpose` that we later optimize away).

> Separately, it seems like you're missing a bunch of datatypes. E.g. openmpi defines way more than you have here. https://github.com/open-mpi/ompi/blob/a06f1c1a4b6212729da406de9f45ce357a5dac7e/ompi/datatype/ompi_datatype_internal.h#L40. What's your thinking here?

They are not standard so if we change implementation (e.g. MPICH, IntelMPI, ...), it will fail. What is there I think is all the datatypes defined in the MPI 5 standard. I couldn't find datatypes in there that could be semantically similar to the Stablehlo ones we are now missing a mapping for. Also, that's clearly an internal header that I'm not sure we should rely on.

The MPI datatypes that we need for fitting other StableHLO primitive types should be handled during runtime initialisation with `MPI_Register_datarep`. And we can think about it after this PR.

Note also that if we can use MPI 5 ABI in the future, we will no longer need to think about implementations.

> Also separately, could you explain why allreduce isn't allowed to cast, but the other ops are?

It's for the reason I wrote above. `MPI_Allreduce` uses the datatype for dispatching to the correct implementation of the reduction operation. So if you bitcast, performing arithmetic operations onto it will be semantically wrong and your data corrupted.

This doesn't happen on just-communicating functions, where there's no arithmetic operation performed to them.

romanlee: All good as per our discussion earlier, thanks!

test/lit_tests/comm/lower-comm/mpi_allreduce.mlir
```diff
@@ -0,0 +1,30 @@
...
+// JIT-NEXT:   %3 = llvm.load %arg5 : !llvm.ptr -> !llvm.ptr
+// JIT-NEXT:   %4 = llvm.call @MPI_Allreduce(%arg0, %arg1, %0, %1, %2, %3) : (!llvm.ptr, !llvm.ptr, i32, !llvm.ptr, !llvm.ptr, !llvm.ptr) -> i32
+// JIT-NEXT:   llvm.return
+// JIT-NEXT: }
+// JIT-LABEL: func.func @main
+// JIT-SAME:                 (%[[BUF:.*]]: tensor<4xf64>, %[[COMM:.*]]: tensor<i64>) -> tensor<4xf64> {
+// JIT-NEXT: %[[RECVBUF:.*]] = stablehlo.constant dense<0.000000e+00> : tensor<4xf64>
+// JIT-NEXT: %[[COUNT:.*]] = stablehlo.constant dense<4> : tensor<i32>
+// JIT-NEXT: %[[DATATYPE:.*]] = stablehlo.constant dense<4> : tensor<i64>
+// JIT-NEXT: %[[OP:.*]] = stablehlo.constant dense<3> : tensor<i64>
```
romanlee: This seems pretty unfortunate: allocate a buffer on the gpu only to immediately copy it back to the host inside the wrapper. But maybe that's the best we can do with the constraints xla puts up?

PR #3109, opened 2026-09-05
Unroll for loops whose trip count the loop bounds decide

tyb0807:
An scf.for or affine.for whose bounds prove exactly one trip on every point of the enclosing domain inlines its body at the lower bound (a strided-copy remainder loop whose extent the propagated block size made constant); one that provably never runs folds to its inits. The bounds and step compose into the block's LoopDomain; the affine.for lower bound materializes as the operand itself when the map is a bare dim or symbol, otherwise as an affine.apply, or an affine.max for several results.

Registered in populateAffineExprSimplificationPatterns, so affine-cfg and simplify-affine-exprs both apply it.

Claude-Session: https://claude.ai/code/session_016zErYp7upmqr4NHfhod9UD

Resolved rebase conflict against `main`.


tyb0807:
Oh just realized this has been merged. Closing

PR #3114, opened 2026-09-05
fix `EnzymeJaXLookupSymbol` by looking on `MappedSymbols` instead of JIT

mofeing:
`EnzymeJaXLookupSymbol` was looking for registered symbols on the JIT engine instead of the `SymbolMap MappedSymbols`, so it was failing even when the symbols where registered.

the function was introduced as a way for other passes (mainly MPI lowering) to lookup for registered symbols.

### DIFF
--- a/src/enzyme_ad/jax/Passes/LowerJIT.cpp
+++ b/src/enzyme_ad/jax/Passes/LowerJIT.cpp
@@ -247,13 +247,17 @@ llvm::orc::SymbolMap MappedSymbols;
 
 bool initJIT();
 
-extern "C" MLIR_CAPI_EXPORTED void *EnzymeJaXLookupSymbol(const char *name) {
+extern "C" MLIR_CAPI_EXPORTED int EnzymeJaXLookupSymbol(const char *name, void **symbol) {
   if (!JIT)
-    return nullptr;
-  auto addr = JIT->lookup(name);
-  if (!addr)
-    return nullptr;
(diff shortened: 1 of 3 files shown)


src/enzyme_ad/jax/xla_ffi/host/mpi.cc
```diff
@@ -13,7 +13,7 @@
 // registered normally (as they are encoded as pointers)
 // - integer constants (MPI_SUCCESS, MPI_STATUS_SIZE, ...) should be registered
 // as pointers to int
-extern "C" void *EnzymeJaXLookupSymbol(const char *name);
+extern "C" int EnzymeJaXLookupSymbol(const char *name, void **symbol);
```
wsmoses, line 16: @mofeing sorry I have a couple of comments here, do you mind reverting for a proper review?

src/enzyme_ad/jax/Passes/LowerJIT.cpp
```diff
@@ -247,13 +247,18 @@ llvm::orc::SymbolMap MappedSymbols;
 
 bool initJIT();
 
-extern "C" MLIR_CAPI_EXPORTED void *EnzymeJaXLookupSymbol(const char *name) {
+extern "C" MLIR_CAPI_EXPORTED int EnzymeJaXLookupSymbol(const char *name,
```
wsmoses, line 250: why does this need to be extern C offhand?

PR #3115, opened 2026-09-05
add `MapSymbol` pass to mock JIT symbols on CLI for lit_tests

mofeing:
requires #3114

it introduces a way to map symbols for JIT through the CLI, like `enzymexlamlir-opt`.

:warning: it is not meant to be used in a regular pipeline (although you could), but as a way to mock, during testing, symbol registration that happens on runtime initialization.

it is specially helpful for testing the `comm.mpi` new lowering to JIT since it queries for such symbols during the lowering and not during JIT compilation (which reduces specialization), so it fails on `lit_tests`. also, it should allow to run `--lower-jit="jit=true"` on lit_tests.


src/enzyme_ad/jax/Passes/MapSymbol.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace mlir::enzyme {
+#define GEN_PASS_DEF_MAPSYMBOLPASS
+#include "src/enzyme_ad/jax/Passes/Passes.h.inc"
+} // namespace mlir::enzyme
+
+using namespace mlir;
+using namespace mlir::enzyme;
+
+// from LowerJIT.cpp
+extern "C" void EnzymeJaXMapSymbol(const char *name, void *symbol);
```
wsmoses: you shouldn't use extern C for this, just make sure to put EnzymeJaXMapSymbol and co into headers

wsmoses: sorry to be clear, just put this in a header included by both this pass and lowerjit

src/enzyme_ad/jax/Integrations/c/EnzymeXLA.h
```diff
@@ -269,6 +269,16 @@ enzymexlaGetTransformPassesList(const EnzymeXLATransformPassesOptions *options,
 /// Free a string returned by enzymexlaGetTransformPassesList.
 MLIR_CAPI_EXPORTED void enzymexlaFreeTransformPassesList(char *passes);
 
+//===----------------------------------------------------------------------===//
+// JIT runtime support
+//===----------------------------------------------------------------------===//
+MLIR_CAPI_EXPORTED int enzymexlaInitJIT();
```
wsmoses: generally we should try to avoid c abi fns returning an int of unknown size [eg int instead of int32/int64], since int is platform dependent

wsmoses: do these actually need to be exported c functions, or just available in lowercomm too?

mofeing: Right, in reality it is a bool which I usually see it as int on the other C APIs I've checked. Anyway, changing it to int32.

mofeing: So we need enzymexlaMapSymbol to register symbols from Reactant. The other two can be removed as the C++ API in Runtime/jit/jit.h is used instead for lowercomm.

wsmoses: yeah just leave the existing symbols as is, and add the existing EnzymeXlaMapSymbol declaration to Utils.h to be used in all passes

mofeing: okay, refactored to that. I've left the C++ version for looking up symbols, as returning a `llvm::Expected<void *>` is very helpful for debugging (we were the only users of `EnzymeJaXLookupSymbol` and it wasn't used by Reactant).

This bindings are also used by the XLA MPI FFI handlers, so I've had to add some deps here and there. That's why the first version added the "Runtime" folder, as a way to share code between the passes and xla_ffi without bazel annoying.

wsmoses (approved):

PR #3122, opened 2026-09-06
implement JIT lowering for `comm.nccl` ops

mofeing:
i've skipped the implementation of some ops that we may not need right now (`ncclCommSplit`, `ncclCommDestroy`, ...).

waiting for #3100 to be merged first (or fuse both PRs into one)

# to do
- [ ] add deps to bazel target
- [ ] add compile time guard for only adding CUDA and NCCL headers and users if CUDA backend is compiled
- [ ] add tests


src/enzyme_ad/jax/Passes/Comm/LowerCommToJIT.cpp
```diff
@@ -1617,10 +1656,1100 @@ struct LowerCommMpiBcastOpToJIT : public OpConversionPattern<comm::MpiBcastOp> {
...
+        /*arg_attrs=*/nullptr,
+        /*res_attrs=*/nullptr,
+        /*output_operand_aliases=*/aliases,
+        /*xla_side_effect_free=*/nullptr);
+
+    return success();
+  }
+};
+
+struct LowerCommNcclCommUserRankOpToJIT
```
romanlee: This will always run on the host so we can always just use MPICommRank, so I don't think we need this Op or a corresponding lowering

src/enzyme_ad/jax/Passes/Comm/LowerCommToJIT.cpp
```diff
@@ -1631,46 +1671,1208 @@ struct LowerCommMpiBcastOpToJIT : public OpConversionPattern<comm::MpiBcastOp> {
...
+      Value datatype =
+          LLVM::LoadOp::create(rewriter, op.getLoc(), type_nccl_datatype,
+                               host_datatype_ptr)
+              .getResult();
+      Value redop = LLVM::LoadOp::create(rewriter, op.getLoc(), type_nccl_redop,
+                                         host_redop_ptr)
+                        .getResult();
+      Value comm =
+          LLVM::LoadOp::create(rewriter, op.getLoc(), type_ptr, host_comm_ptr)
+              .getResult();
```
romanlee, line 2565: This seems pretty unfortunate: allocate a buffer on the device, only to immediately copy it back to the host in the jit wrapper? But maybe this is the best we can do given xla constraints?

mofeing: we should be able to avoid this memcpys by using attributes, but i don't know yet how to do it with jit_call. i think it's fine for now and can be improved once we learn how to pass attributes with jit_call.

PR #3127, opened 2026-09-06
Add module-owned persistent XLA temporaries

vimarsh6739:
Host code needs XLA temporary buffers that survive repeated calls without allocating and freeing storage on every invocation. Add `enzymexla.temp_alloc` to declare module-owned storage and `enzymexla.get_global_temp` to retrieve its stable, borrowed handle.

`convert-polygeist-to-llvm` lowers declarations to internal handle slots and LLVM constructors/destructors using the existing XLA allocation ABI. Runtime initialization has priority 65534 and temporary initialization has priority 65535, so the shared runtime outlives every module's temporaries. Internal linkage keeps identically named temporaries in separate translation units distinct.

The operations can be emitted independently of `XLAMegakernelizePass`. Copies remain explicit at their original sites; switching LBM's producer to these operations and copy hoisting are separate follow-ups. The initial implementation supports static, identity-layout device memrefs with C-style lowering and documents synchronization requirements for overlapping uses.

Validation:

- Added four LIT tests covering XLA CPU/GPU lowering, constructor/destructor priorities, repeated access/CSE, symbol retention and name collisions, and invalid declarations/types/backend options. Updated the existing XLA lowering test for the runtime priority.
- All 14 selected LIT regression targets passed locally.
- Built `enzymexlamlir-opt` and `mlir-translate`. Linked two separately lowered translation units and exercised them against both a counting mock and the locally built ReactantExtra on an RTX 5090: one shared runtime, two distinct persistent allocations, 200 updates, 400 copies, stable handles, and exactly two frees before runtime teardown.

### DIFF
--- a/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
+++ b/src/enzyme_ad/jax/Dialect/EnzymeXLAOps.td
@@ -250,6 +250,49 @@ let skipDefaultBuilders = 1;
 let builders = [OpBuilder<(ins "ValueRange" : $blockSizes)>, OpBuilder<(ins)>];
 }
 
+def TempAllocOp : EnzymeXLA_Op<"temp_alloc", [Symbol, HasParent<"ModuleOp">]> {
+  let summary = "declare a module-owned persistent XLA allocation";
+  let description = [{
+    Declares uninitialized, statically shaped device storage allocated when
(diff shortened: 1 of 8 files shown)


wsmoses (approved):
presumably in a follow up you'll update the raise to shlo pass to use this for its temporaries [and also do some hoisting]

PR #3131, opened 2026-09-06
EnzymeHLOOpt: bound the provenance walk behind DUSDUSSubsuming; fix its dangling insert

wsmoses:
`computeTensorValueProvenanceImpl` follows a chain of slices, updates and pads back to its sources with a Presburger relation that grows with the chain, and `DUSDUSSubsuming` recomputes it from scratch for every update it visits: cubic in the chain. mfem's TMOP metric kernels write a scratch buffer a few hundred times. One such kernel (270 updates, 14k lines) spends 132 s in this pattern and 3 s in everything else (`passses` without bit 512), and the eight `fem/tmop/metrics/3xx.cpp` translation units carry the whole cost of running `enzyme-hlo-opt` on raised kernels at compile time (#3129): 464 s of a 533 s compile, per the pass timing.

The walk now stops after 128 steps and treats what is left as an opaque source. That only loses precision: both provenances the pattern compares are computed against the same sources, so an opaque prefix makes the comparison conservative, never wrong. The kernel takes 3 s; the 45-kernel translation unit's optimizer time goes from 411 s to 78 s.

The slice case also inserted into the provenance map with arguments that were references into that map (`try_emplace(value, it->second.sources, it->second.provenanceRelation)`); the insertion can rehash, after which they dangle and the copy reads freed memory. Nine of the 45 kernels segfault on that, in `PresburgerRelation`'s copy constructor. Copy first.

Test: a three-update chain read back through overlapping slices, checked line for line. The reduced shapes that exercise the pattern fold before the walk gets deep, and the kernels that hit the cost and the crash are 8-14k lines each (`fem/tmop/metrics/315.cpp`, kernels 43 and 44 of its pre-optimizer dump; I can attach them if useful). Lit passes.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -84,6 +84,12 @@ using namespace mlir;
 using namespace mlir::enzyme;
 using namespace mlir::stablehlo;
 
+// Steps the tensor provenance walk (DUSDUSSubsuming) takes through a chain of
+// slices, updates and pads before treating the rest as an opaque source. Each
+// step is Presburger work on a relation that grows with the chain; the cutoff
+// only loses precision.
+constexpr size_t kProvenanceWalkBudget = 128;
+
 namespace mlir {
 // Implementation of helper function to lower MultiRotateOp into individual
 // RotateOps
@@ -19119,10 +19125,15 @@ addSelfProvenance(Value value,
 void computeTensorValueProvenanceImpl(
     SmallVectorImpl<Value> &pendingValues,
     DenseMap<Value, TensorValueProvenance> &provenanceInfo2) {
+  size_t steps = 0;
   while (!pendingValues.empty()) {
     Value value = pendingValues.pop_back_val();
     if (provenanceInfo2.contains(value))
       continue;
+    if (steps++ >= kProvenanceWalkBudget) {
+      addSelfProvenance(value, provenanceInfo2);
+      continue;
+    }
 
     if (auto slice = value.getDefiningOp<stablehlo::SliceOp>()) {
       // If we don't know the provenance of the operand, figure it out before
@@ -19143,8 +19154,12 @@ void computeTensorValueProvenanceImpl(
       presburger::IntegerRelation offset =
           getOffsetRelation(domainSpace, slice.getStartIndices());
 
+      // The insertion may rehash the map and invalidate references into it.
+      SmallVector<Value> operandSources = it->second.sources;
+      presburger::PresburgerRelation operandProvenance =
+          it->second.provenanceRelation;
       auto [insertIt, _] = provenanceInfo2.try_emplace(
-          value, it->second.sources, it->second.provenanceRelation);
+          value, std::move(operandSources), std::move(operandProvenance));
       insertIt->second.provenanceRelation =
           insertIt->second.provenanceRelation.intersectDomain(
               presburger::PresburgerSet(restriction));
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -19116,13 +19116,28 @@ addSelfProvenance(Value value,
...
 // Compute provenance for all given values and popupate the provenance info data
 // structure with it.
+// How far back the provenance walk follows a chain of slices, updates and
+// pads before it treats what is left as an opaque source. Each step is a
+// Presburger union or intersection over a relation that grows with the
+// chain, and the pattern below recomputes it from scratch for every update
+// it visits, so an unbounded walk over a kernel's hundreds of updates into
+// one scratch buffer is cubic in the chain (a 270-update kernel took over
+// two minutes). An opaque source only loses precision: both provenances the
+// pattern compares are computed against the same sources.
```
ftynse: Use https://github.com/ftynse/skills/blob/main/llvm-comment-hygiene/SKILL.md with Claude for LLVM code.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -19116,13 +19116,28 @@ addSelfProvenance(Value value,
...
+constexpr size_t kProvenanceWalkBudget = 128;
```
ftynse: Can we hoist it at least to the top of the file if not to a separate header where such constants can be easily found and configured?

ftynse (approved):
I don't quite follow the claim of cubic complexity, don't we store/reuse provenanceInfo between calls?

This can be also done as a dataflow analysis to improve caching, but we currently don't have a stopping mechanism for that either.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: Trimmed per the guide: the constant's doc comment now says what it bounds and that the cutoff only loses precision; the timings and the kernel it came from stay in the PR description. The inline comment at the copy is down to the one reason the copy exists.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
wsmoses: Hoisted to the top of the file, right after the using-declarations, with a doc comment naming the pattern it bounds.

wsmoses:
On the cubic claim: `provenanceInfo` is a local `DenseMap` in `DUSDUSSubsuming::matchAndRewriteImpl`, so nothing is reused between pattern applications. For a chain of N updates into one buffer the pattern fires once per update whose operand is an update, each application walks the chain back to its sources (N steps for the N-th update), and each step is a union or subtraction on a relation whose piece count grows with the depth. That is N applications x N steps x N-ish per step. Within one application the map does cache: the slices computed afterwards reuse it, which is why the budget applies per walk rather than per lookup.

Sharing the map across applications would take invalidation, since this pattern and its neighbours in the same driver replace and erase the values it is keyed on; a dataflow analysis with a proper invalidation story would be the real fix, and I agree it needs a stopping mechanism too. The budget is the stopgap that makes the TMOP translation units compile in a reasonable time.

Also rebased on main (7703fa2d3) so lit is clean: 1315/1315.

ftynse:
Maybe you can attempt having one megapattern that does all such rewrites in an isolated-from-above region in one call and thus can reuse the map?

wsmoses:
I agree, but for now I may just merge the forced compile time cap for the paper deadline push to continue, and revisit in post

PR #3138, opened 2026-09-07
EnzymeHLOOpt: a masked pick reduce over a pad with a guaranteed-true lane is a slice

wsmoses:
Follow-up to #2960's large case. The raiser picks one lane's value out of a reduced axis (a `tid == 0` broadcast store) with a `(value, mask)` `stablehlo.reduce` whose body is `select(mask, new, acc)` / `or(mask, acc)`. After #3137 the pure mask reduce folds, but the value pick stayed a full reduction: at 4096 × 256 it reduced a padded `256x4096xi1` mask against a transposed copy of the input.

Which of several live lanes such a pick returns is unspecified (it is the reduction order), so any lane guaranteed live may be returned. `OneHotMaskedReduce`: when the mask operand is a `stablehlo.pad` with a lane that is guaranteed true for every kept position, the pick is the slice at that lane (reshaped to the result shape) and the reduced mask is `true`. A lane is guaranteed true either because the padding value is true and every reduced dim is padded (a padded index), or because the operand is all true and the kept dims are unpadded or padded with true (the operand's first index). Negative or interior padding is left alone. The slice then folds through the transpose with the existing patterns, so the large case becomes

```
%0 = stablehlo.reshape %arg0 : (tensor<1048576xf64>) -> tensor<4096x256xf64>
%1 = stablehlo.slice %0 [0:4096, 0:1] : (tensor<4096x256xf64>) -> tensor<4096x1xf64>
%2 = stablehlo.reshape %1 : (tensor<4096x1xf64>) -> tensor<4096xf64>
```

Per DEVDOCS: registered as the `one_hot_masked_reduce` transform op, added to the pass list in `EnzymeXLA.cpp` and to `test/test_utils.py`. Test: the exact optimized form of #2960's `lane0_large`, a lane-3 pad with the mask result feeding a select, a true-padded mask over an arbitrary operand, and an all-true operand false-padded along the kept axis that must not fold; a second RUN line drives the pattern alone through the transform interpreter (`TD:` goldens).

### DIFF
--- a/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
+++ b/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
@@ -569,6 +569,7 @@ static void addBaseTransformPasses(std::vector<std::string> &list,
   list.push_back("multiply_negated_operands_simplify");
   list.push_back("factor_scalars_in_dot_general");
   list.push_back("reduce_mul_to_dot_general");
+  list.push_back("one_hot_masked_reduce");
   list.push_back("dot_general_broadcast_in_dim");
   list.push_back("dot_general_broadcast_in_dim_sort_dims");
   list.push_back("dus_dynamic_slice_simplify");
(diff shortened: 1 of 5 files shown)


wsmoses:
Parked as a draft; tracked in #3139. The reduce this simplifies is #2949's masked lane pick (value varies along the lane axis), not #2960's or-reduce, which gets its own pad-aware elimination in #2960.

PR #3141, opened 2026-09-07
Lower wrapping affine scatter grids to dynamic update slices

vimarsh6739:
The existing `ScatterOpCanon::tryRewriteScatterWithIotaIndexing` rewrite rejects affine grids that cross a boundary in the reshaped destination, leaving point scatters in LBM. Extend it to split the update into in-bounds rectangular pieces and emit one dynamic update slice per piece.

For example, `7 + i + 8*j`, with `i` in `[0,2)` and `j` in `[0,3)`, writes `[0:3,7:8]` and `[1:4,0:1]` in an 8-column view. A single DUS would clamp its start and write different cells. Preserve index/update pairing through axis permutations, reversals, and reshaped grids; reject out-of-bounds writes, overflowing index arithmetic, repeated indices, and unsupported window/batching semantics.

Add explanatory FileCheck and StableHLO interpreter coverage for row wrapping, multiple-axis wrapping, reversed axes, reshaped grids, and rejected cases. Validation: the new test and six related gather/scatter LIT targets pass against current main.

This PR contains only the iota-grid extension. Full-LBM integration also uses [separate DUS-chain sharing/convergence fixes](https://github.com/EnzymeAD/Enzyme-JAX/commit/d31361ea) on `vim/lbm-staging`; those are outside this PR. PR #3130 has been reverted from staging for this comparison.

Fresh staging validation: unchanged LBM, 20 timesteps, three Nsight runs per backend. The raised loop body has zero scatters/gathers and 48 DUS operations per two timesteps; output is byte-identical to the previous XLA result and agrees with CUDA. Median summed GPU kernel time is 8.704 ms / 511 launches for XLA versus 5.781 ms / 20 launches for CUDA. The separate sharing fix removes the full-buffer copies, but DUS-write fusion remains unresolved. Compilation and transfers are excluded from these GPU kernel timings.


test/lit_tests/scatter_wrapping_grid.mlir
```diff
@@ -0,0 +1,213 @@
...
+// A flat affine grid can cross boundaries of its reshaped operand. Split the
+// update at those boundaries, preserving both the destination addresses and
+// their corresponding update values. The runtime checks at the end distinguish
+// correct wrapping from DUS clamping and cover out-of-bounds scatter semantics.
+
+// indices[i,j] = 7 + i + 8*j. Split into [0:3,7:8] and [1:4,0:1].
+// CHECK-LABEL: func.func @wrap_row
+// CHECK-NOT: stablehlo.scatter
+// CHECK-COUNT-2: stablehlo.dynamic_update_slice
+// CHECK-NOT: stablehlo.scatter
```
Pangoraw, line 13: nit: can you put the full lines with sliced starts here?

Pangoraw (approved):
feel free to merge

PR #3146, opened 2026-09-07
Cancel partial inverse transposes across elementwise operations

vimarsh6739:
`TransposeElementwiseTransposeSimplify` currently requires every operand to be an inverse transpose, leaving patterns such as `T(select(T^-1(p), x, T^-1(y)))` unchanged. Extend it to produce `select(p, T(x), y)`, reducing three transposes to one, including when the select predicate is scalar.

For partial cancellation, require a single-use elementwise result and a strict decrease in transpose count, accounting for shared inputs. Preserve operation properties when rebuilding the elementwise op. Run the partial form in a final cleanup inside EnzymeHLOOpt: including it in the main factoring/CSE rewrite set caused cycling on the full LBM input. Existing full cancellation remains in the main rewrite set.

Add explanatory FileCheck and StableHLO interpreter coverage for partial cancellation, scalar predicates, non-self-inverse permutations, shared arithmetic, and unprofitable rewrites. Validation: all 29 related transpose LIT targets pass on this independent branch based on current main. The combined staging build also passed all 44 related transpose/comparison/AND tests and optimized the full LBM pre-pass input in approximately 0.11 seconds.

The same change is on `vim/lbm-staging`. In the existing all-DUS LBM pipeline, this and the masked-comparison change in #3147 reduced StableHLO transposes from 130 to 72 and i8-to-i32 conversions from two to zero. The unchanged 20-timestep benchmark still launched 511 kernels; median summed GPU kernel time was 9.258 ms versus 8.707 ms before these two changes (three captures on an RTX 5090, excluding compilation and transfers). XLA had already removed explicit transpose/reshape HLO operations in the baseline, and output remained byte-identical. These combined measurements do not establish an isolated performance benefit for this PR.

CI note at submission: the formatting check reports a pre-existing line-wrap difference in the DUSDUS `LLVM_DEBUG` block on main (`fd510e64`). Running clang-format 16 on main and this branch gives the identical diff.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -7734,37 +7734,85 @@ using BroadcastingElementwiseAllTransposeOperandsSimplify =
 struct TransposeElementwiseTransposeSimplify
     : public CheckedOpRewritePattern<stablehlo::TransposeOp,
                                      TransposeElementwiseTransposeSimplify> {
-  using CheckedOpRewritePattern<
-      stablehlo::TransposeOp,
-      TransposeElementwiseTransposeSimplify>::CheckedOpRewritePattern;
+  bool allowPartial;
+
+  TransposeElementwiseTransposeSimplify(MLIRContext *context,
+                                        PatternBenefit benefit = 1,
+                                        bool allowPartial = true)
+      : CheckedOpRewritePattern(context, benefit), allowPartial(allowPartial) {}
 
   LogicalResult matchAndRewriteImpl(stablehlo::TransposeOp op,
                                     PatternRewriter &rewriter) const {
     auto elem = op.getOperand().getDefiningOp();
-    if (!elem)
+    if (!elem ||
+        (!stablehlo::hasTraitElementwise(elem) &&
+         !isa<stablehlo::SelectOp>(elem)) ||
+        elem->getNumResults() != 1 || elem->getNumRegions() != 0)
       return failure();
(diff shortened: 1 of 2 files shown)


vimarsh6739:
@Pangoraw can you take a look here

test/lit_tests/transpose_elementwise_partial.mlir
```diff
@@ -0,0 +1,100 @@
+// RUN: enzymexlamlir-opt %s --enzyme-hlo-opt | FileCheck %s
+// RUN: enzymexlamlir-opt %s --enzyme-hlo-opt --inline --canonicalize --symbol-dce | stablehlo-translate --interpret
```
Pangoraw, line 2: default allowPartial is false so it would be nice to have the more complete lit checks (with operands) so to make sure it tests the feature

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -37479,6 +37528,20 @@ struct EnzymeHLOOptPass
...
+    // those patterns cannot recreate the intermediate transposes it removes.
+    RewritePatternSet transposeCleanup(context);
+    transposeCleanup.add<TransposeElementwiseTransposeSimplify>(context);
+    if (passses & 2048)
+      transposeCleanup.add<TransposeTranspose>(context);
+    if (failed(applyPatternsGreedily(getOperation(),
+                                     std::move(transposeCleanup), config))) {
+      signalPassFailure();
+      return;
     }
```
Pangoraw, line 37545: this feels a bit wonky

PR #3147, opened 2026-09-07
Narrow masked integer equality comparisons before conversion

vimarsh6739:
Extend `CompareConvert` to simplify integer flag tests such as `(convert<i8 -> i32>(x) & 1) == 0` to `(x & 1) == 0` at i8 width, preserving element order through intervening reshapes.

Apply this to EQ/NE when every mask and expected-value element fits in the source width. The mask discards extension bits, so the comparison remains valid for negative inputs and high-bit masks. Handle swapped operands and reject ordered comparisons or constants needing wider bits.

The high-bit test exposed a necessary supporting correction in `NonNegativeResultAnalysis`: an AND result is not unconditionally nonnegative. Require at least one operand to be provably nonnegative before making that deduction, so narrowing `(convert(x) & 128) == 128` does not incorrectly fold to false. This part overlaps with #2652 while retaining proofs for masks that clear the sign bit.

Add explanatory FileCheck and StableHLO interpreter coverage for low/high-bit masks, negative inputs, swapped operands, and rejected cases. Validation: all 15 related comparison/AND LIT targets pass on this independent branch based on current main. The combined staging build also passed all 44 related transpose/comparison/AND tests.

The same change is on `vim/lbm-staging`. Together with the transpose cancellation change in #3146, it removes both i8-to-i32 flag conversions from the all-DUS LBM loop body. The unchanged 20-timestep benchmark still launched 511 kernels, with byte-identical output; the combined changes measured 9.258 ms median summed GPU kernel time versus 8.707 ms before (three RTX 5090 captures, excluding compilation and transfers). This PR has not been benchmarked in isolation.

CI note at submission: the formatting check reports a pre-existing line-wrap difference in the DUSDUS `LLVM_DEBUG` block on main (`fd510e64`). Running clang-format 16 on main and this branch gives the identical diff.


src/enzyme_ad/jax/Utils.cpp
```diff
@@ -970,11 +970,19 @@ NonNegativeResultAnalysis::State NonNegativeResultAnalysis::localGuaranteed(
 
   // integer ops
   if (isa<stablehlo::AbsOp, stablehlo::SqrtOp, stablehlo::ExpOp,
-          stablehlo::IotaOp, stablehlo::AndOp, stablehlo::OrOp,
-          stablehlo::XorOp, stablehlo::NotOp>(op)) {
+          stablehlo::IotaOp, stablehlo::OrOp, stablehlo::XorOp,
+          stablehlo::NotOp>(op)) {
```
Pangoraw, line 974: in a follow-up, the same kind of fix should apply to or, xor and not

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -9575,6 +9575,75 @@ struct CompareConvert
...
                                     PatternRewriter &rewriter) const {
+    // An equality test of a masked integer extension can be performed in the
+    // source width when both constants use only source-width bits. The mask
+    // discards all extension bits, including the sign extension of negative
+    // inputs. Reshapes preserve the element order and can use the narrow type.
+    auto direction = cmpOp.getComparisonDirection();
+    if (direction == stablehlo::ComparisonDirection::EQ ||
+        direction == stablehlo::ComparisonDirection::NE) {
+      for (int i = 0; i < 2; ++i) {
+        auto andOp = cmpOp->getOperand(i).getDefiningOp<stablehlo::AndOp>();
```
Pangoraw, line 9586: maybe break if i1 type? there are other patterns in this case

Pangoraw (approved):
looks good

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
vimarsh6739: isn't the result of a stablehlo.compare always an i1? which pattern are you referring to?

PR #3149, opened 2026-09-08
EnzymeHLOOpt: format the DUSSliceSimplify debug check

wsmoses:
CI's clang-format wants the range-for inside this `LLVM_DEBUG` broken differently from the local formatter, so main fails the Format check since #3136 (e.g. https://github.com/EnzymeAD/Enzyme-JAX/actions/runs/34172682365/job/101895892480 on #3124). Bind the two shapes first so the range line is short and both versions agree.


wsmoses:
Superseded: the format was fixed directly on main.

PR #3151, opened 2026-09-08
Raising: lane-private scratch gets the batched axes as leading dimensions

wsmoses:
Replaces #2996. Scratch allocated inside batched axes is one buffer per lane, but the raising modeled it as a single tensor: a store whose indices do not involve the lane axes looked uniform, so every lane read lane 0's values (NormalTraceJump's per-thread `grad[3]` accumulators; 8-assertion miscompile in the MFEM GPU suite without a fix).

Rather than rewriting the alloca before raising, the buffer is shaped at the point it is raised: one leading dimension per batched axis whose loop encloses the alloca (parallel axes and lockstep `affine.for`s alike, from `ParallelContext::ivs`), with the axes recorded in the alloca's value map. Every affine or memref load, store and atomic add then prepends the lane induction variables to its indices (`prependLaneDims` / `laneIndices`), so the existing slice, gather, scatter and masked-update machinery sees an ordinary access into a `[lanes..., n]` buffer. Any other kind of user never raised, so there is no gating on the alloca's users.

Shared memory sits between the grid parallel and the thread parallel, so it batches over the grid alone: scratch between two parallels gets only the outer axes and stays shared by the inner lanes. `raise_masked_store_overhang.mlir` and `raise_pad_guarded_lane_load.mlir` model such buffers, so their allocas move above the lane parallel (their goldens only reorder the zero constant).

The masked dynamic-update-slice store path re-read the store op's own map and indices; it now works from the extended access map, which the bounded-parallel test (a masked axis under a size-1 grid axis) exercises.

Test: `raise_lane_private_scratch.mlir`, four cases (per-lane scratch, the same through memref.load/store, per-block scratch shared by inner lanes, atomic adds into per-lane scratch) with full goldens for the raised and the `--enzyme-hlo-opt` output; the optimizer folds each to the expected slice or reverse.

### DIFF
--- a/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
+++ b/src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
@@ -883,6 +883,49 @@ bool isSafeToSpeculativelyExecuteAtScope(Operation *scope, Operation *op) {
   return inBounds;
 }
 
+// A buffer raised with leading lane dimensions (see the memref.alloca case)
+// is indexed by the lane induction variables before its own indices.
+static void prependLaneDims(Value memref, affine::AffineValueMap &avm,
+                            DenseMap<Value, affine::AffineValueMap> &maps) {
+  auto it = maps.find(memref);
+  if (it == maps.end())
+    return;
+  const affine::AffineValueMap &lane = it->second;
+  unsigned K = lane.getNumResults();
+  AffineMap map = avm.getAffineMap();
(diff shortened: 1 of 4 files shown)


wsmoses:
Updated per review (719446f78): the axes are every batched loop enclosing the alloca, no barrier heuristic and no gating on the alloca's users; `memref.atomic_rmw` prepends the lane indices too, with a fourth test case for it. The two barrier tests now keep their shared scratch above the lane parallel, where shared memory sits.

PR #3152, opened 2026-09-08
Factor transposes through select and clamp

vimarsh6739:
`ElementwiseAllTransposeOperandsSimplify` skips `select` and `clamp` because StableHLO gives them the `BroadcastingElementwise` trait. This blocks transpose factoring through shared expressions; LBM retained 72 transposes after optimization.

Reuse the existing pattern for broadcasting elementwise operations. Preserve scalar predicates and clamp bounds, derive the permuted result shape from the original result, and use `Operation::create` with the rewritten operands, result type, and original attributes. The ordinary elementwise transform registration keeps its existing name. Register the broadcasting variant as `broadcasting_elementwise_all_transpose_operands_simplify` in the transform dialect and immediately after the ordinary variant in the C API downward transpose propagation list.

The new LIT test explains the before/after behavior for shared select arithmetic, scalar operands, non-splat constants, non-self-inverse permutations, and incompatible permutations, with numerical checks against independent expected values in the StableHLO interpreter.

Validation:

- 73 related transpose/select/clamp and transform LIT tests pass on this branch based on main; 74 pass on LBM staging. The regression directly applies the new named transform pattern and checks that a sharding attribute is preserved.
- Clang-format 16 passes for the changed C++ file; `git diff --check` passes.
- In the normal LBM staging pipeline, post-optimization transposes drop from 72 to 0 and reshapes from 22 to 12. LBM sources and executable are unchanged, and output is byte-identical to the validated baseline.
- A 20-timestep LBM profile still has 511 GPU kernel launches and 48 DUS operations in the optimized loop body.

The same logical change is included on `vim/lbm-staging` as `f356ef738e2782efb8f36a52137f47d1cc665bff`; the only integration difference is retaining each branch's existing transpose-cancellation registration.

Fresh LBM staging benchmark (20 timesteps on an RTX 5090; medians of three captures per configuration):

| Version | Transposes | GPU launches | Summed GPU kernel time |
|---|---:|---:|---:|
| XLA baseline | 72 | 511 | 9.398 ms |
| XLA with this change | 0 | 511 | 9.403 ms |
| CUDA reference | — | 20 | 5.756 ms |

Before/after XLA runtimes were profiled in alternating order after builds completed. The +0.06% difference is within the observed variation; this change simplifies the input IR but shows no measurable LBM speedup. GPU kernel time excludes compilation, transfers and launch gaps. All XLA outputs are byte-identical to the baseline and agree with CUDA within atol=1e-6, rtol=1e-5 (maximum absolute error 5.96e-7).

After the review fixes, the rebuilt runtime generates byte-identical optimized LBM MLIR to the benchmarked version (zero transposes). The C API was checked directly to confirm that the new pattern immediately follows the ordinary elementwise pattern when downward transpose propagation is selected.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -7629,12 +7629,13 @@ struct ScatterToDynamicUpdateSlice final
   }
 };
 
-struct ElementwiseAllTransposeOperandsSimplify
+template <template <typename> class Trait>
+struct ElementwiseAllTransposeOperandsSimplifyBase
     : public CheckedOpTraitRewritePattern<
-          OpTrait::Elementwise, ElementwiseAllTransposeOperandsSimplify> {
+          Trait, ElementwiseAllTransposeOperandsSimplifyBase<Trait>> {
   using CheckedOpTraitRewritePattern<
-      OpTrait::Elementwise,
-      ElementwiseAllTransposeOperandsSimplify>::CheckedOpTraitRewritePattern;
+      Trait, ElementwiseAllTransposeOperandsSimplifyBase<Trait>>::
+      CheckedOpTraitRewritePattern;
 
   LogicalResult matchAndRewriteImpl(Operation *op,
                                     PatternRewriter &rewriter) const {
@@ -7647,6 +7648,16 @@ struct ElementwiseAllTransposeOperandsSimplify
     DenseI64ArrayAttr permutation;
     bool foundTranspose = false;
     for (auto operand : op->getOperands()) {
+      auto type = dyn_cast<RankedTensorType>(operand.getType());
+      if (!type)
+        return failure();
+      // Select predicates and clamp bounds may be scalars. They have no axes
+      // to permute and do not determine the result's shape.
+      if (type.getRank() == 0) {
+        kinds.push_back(OperandKind::Scalar);
+        operands.push_back(operand);
+        continue;
+      }
       if (matchPattern(operand, m_Constant())) {
         kinds.push_back(OperandKind::Const);
         operands.push_back(operand);
@@ -7683,6 +7694,7 @@ struct ElementwiseAllTransposeOperandsSimplify
     for (size_t i = 0; i < operands.size(); i++) {
       switch (kinds[i]) {
       case OperandKind::Transpose:
+      case OperandKind::Scalar:
         break;
       case OperandKind::Const:
         // This will be eliminated by a transpose(constant) -> constant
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -7629,24 +7629,42 @@ struct ScatterToDynamicUpdateSlice final
...
+      CheckedOpTraitRewritePattern;
 
   LogicalResult matchAndRewriteImpl(Operation *op,
                                     PatternRewriter &rewriter) const {
     if (op->getNumOperands() < 2)
       return rewriter.notifyMatchFailure(
           op, "ElementwiseAllTransposeOperandsSimplify needs >= 2 operands");
 
+    auto resultType = dyn_cast<RankedTensorType>(op->getResult(0).getType());
+    if (!resultType || resultType.getEncoding() ||
```
wsmoses: can you remove this check?

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -37073,6 +37092,8 @@ struct EnzymeHLOOptPass
 
     patterns.add<ElementwiseAllTransposeOperandsSimplify,
                  TransposeElementwiseTransposeSimplify,
+                 ElementwiseAllTransposeOperandsSimplifyBase<
```
wsmoses: make sure to add this to transform patterns and enzymexla.cpp list of patterns at the same place as this

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
vimarsh6739: Removed the check, including the operand-encoding exclusion, in 434a854f. The direct transform regression now factors a select carrying `mhlo.sharding` and checks that the attribute is preserved.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
vimarsh6739: Added the broadcasting variant to `TransformOps.td` and added `broadcasting_elementwise_all_transpose_operands_simplify` immediately after the existing elementwise entry in `EnzymeXLA.cpp`'s downward-propagation list, in 434a854f. The LIT regression directly applies the named transform pattern; all 73 related PR tests pass. Also checked the rebuilt runtime's C API list directly.

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -7694,26 +7706,30 @@ struct ElementwiseAllTransposeOperandsSimplify
...
-    auto newOp = Operation::create(
-        op->getLoc(), op->getName(),
-        {RankedTensorType::get(
-            elemResShape,
-            cast<TensorType>(op->getResult(0).getType()).getElementType())},
-        operands, op->getAttrs(), mlir::PropertyRef(), op->getSuccessors(), 0);
-    rewriter.insert(newOp);
+    auto resultType = cast<RankedTensorType>(op->getResult(0).getType());
+    SmallVector<int64_t> elemResShape;
+    for (int64_t dim : invPerm.asArrayRef())
```
wsmoses, line 7711: can you keep the old Operation::create instead of doing a clone and set operands

vimarsh6739: ig we need that for perf

wsmoses: no I'm more worried about the rewriter knowing which operands to visit since the set operands are outside of the rewrtier call [in contrast to old one which presumably traverses all when it adds operands]

vimarsh6739: Restored the original `Operation::create` plus `rewriter.insert` construction in 6313f882, passing the final operands and result type directly. The result shape still comes from the original result so scalar predicates and clamp bounds remain supported. All 73 related PR tests pass, including attribute preservation; the rebuilt runtime produces byte-identical optimized LBM MLIR (zero transposes).

PR #3157, opened 2026-09-08
Pin julia version to 1.12.6 for Reactant integration

Pangoraw:
CI has been red in Reactant for the 1.12.7 bump due to a change in LinearAlgebra relying on invoke (that Reactant does not seem to handle correctly). Anyway, this pins the Julia version to 1.12.6 to see if gets green again while the bug is fixed in Reactant proper.


Pangoraw:
Reactant integration is green again, merging.

PR #3159, opened 2026-09-08
Raising pipeline: inline kernel functor methods and discard their dead copies

wsmoses:
Why #2933's shape exists, and a pipeline-side way through it.

MFEM's `CuWrapND` takes the lambda capture by reference. The functor's method (`DerefineMatrixOpMultFunctor<…>::Run(int)` in `fem/derefmat_op.cpp`) copies the capture out of `this` as one aggregate load taken apart by `extractvalue`, while its caller stored the fields one by one into the functor alloca (12 typed-view stores, no escapes). mem2reg forwards those fields as soon as the load and the stores meet in one function; on the pre-raise dump of that TU, after the pipeline's own `inline` and `polygeist-mem2reg`, no `extractvalue` remains in the caller. Two things kept them apart:

- The method still carries the `personality` the front end gave it after its invokes were lowered, and the LLVM dialect inliner refuses any callee with a personality (`InlinerInterfaceImpl.cpp`, "TODO: Handle exceptions"). A function without a landing pad has no use for one: unwinding passes straight through it either way. `strip-dead-personality` drops it.
- Once inlined, the `linkonce_odr` definition stays behind (symbol-dce keeps anything not private), and its own copy of the aggregate load still fails the raising. Every translation unit that uses such a function defines its own copy, so an unreferenced one is discardable, as LLVM's GlobalDCE does. `discard-unreferenced-linkonce` erases linkonce/weak definitions with no symbol uses (comdat selectors are not uses), to a fixpoint.

Both run around the pipeline's second `inline`. On the `derefmat_op.cpp` dump, with these two steps and no #2933, all four kernels raise.

Tests: `strip_dead_personality.mlir` (a landing-pad function keeps its personality), `discard_unreferenced_linkonce.mlir` (referenced, external and declared functions stay), `inline_host_functor.mlir` (the by-reference functor shape end to end: the call ends up with the stored SSA values and no aggregate load).

Validation on the 18 TUs that currently need #2933, and on the full suite, is running; I'll post the numbers here.

### DIFF
--- a/src/enzyme_ad/jax/Passes/DiscardUnreferencedLinkonce.cpp
+++ b/src/enzyme_ad/jax/Passes/DiscardUnreferencedLinkonce.cpp
@@ -1,0 +1,66 @@
+//===- DiscardUnreferencedLinkonce.cpp - GlobalDCE for linkonce functions -===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===---------------------------------------------------------------------===//
+//
+// symbol-dce keeps every symbol that is not private, so a linkonce or weak
+// definition the inliner has folded into all its callers survives it. Each
+// translation unit that uses such a function defines its own copy, so an
+// unreferenced definition is discardable, as in LLVM's GlobalDCE.
+//
+//===---------------------------------------------------------------------===//
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "src/enzyme_ad/jax/Passes/Passes.h"
+
+namespace mlir {
+namespace enzyme {
+#define GEN_PASS_DEF_DISCARDUNREFERENCEDLINKONCEPASS
+#include "src/enzyme_ad/jax/Passes/Passes.h.inc"
+} // namespace enzyme
+} // namespace mlir
+
+using namespace mlir;
+
+namespace {
+static bool isDiscardableLinkage(LLVM::Linkage linkage) {
+  switch (linkage) {
+  case LLVM::Linkage::Linkonce:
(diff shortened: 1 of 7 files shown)


wsmoses:
Split into StripDeadPersonality.cpp and DiscardUnreferencedLinkonce.cpp. On unwinding: a personality is consulted only through landing pads, so a function without one has no exception table and a throwing callee unwinds straight through it as it would through any function that never had a personality; there is no call to prove nounwind. An invoke needs a landing pad to unwind to, so no landing pad also means every call in the function is a plain llvm.call, which is the only kind of call site the inliner takes: the caller had no handler for that call either, and the inlined calls keep unwinding through. This is written up in the file header.

wsmoses:
Validation on the union without #2933 (union17 + this PR): of the 18 TUs that need #2933, only `fem/derefmat_op.cpp` compiles with this; 17 still fail on the same aggregate load. Why the two steps reach only that one:

- `fem/pderefmat_op.cpp` has the same by-reference `Run` shape (8 variants, plain call sites, no landing pads), and `strip-dead-personality` does drop the personalities, but the calls sit inside `scf.if`/`scf.for` regions of the caller. `Run` starts with the aggregate `affine.load`, and `AffineInlinerInterface::isLegalToInline(op, region)` only admits an affine op into a region whose parent has the `AffineScope` trait or is an affine for/parallel/if, so the inliner declines the whole call. In `derefmat_op.cpp` the calls are at the callers' top level.
- The TMOP metrics (`TMOPAssembleGradPA2D::Mult<…>` and friends) are address-taken through the kernel dispatch table and carry 20 real landing pads each, so they are never inlined; there the lambda object is an alloca in the same function with 36 field stores, so the aggregate load could be forwarded in place once split by field.

So this PR is a general improvement (more host functors inline, dead copies go), but #2933's field split stays necessary for the other 17. The natural follow-up is to run that split before `polygeist-mem2reg` as well, so the in-function cases (TMOP) forward to SSA values instead of reaching the raising as loads.

PR #3160, opened 2026-09-08
Raising pipeline: lower affine around the second inline

wsmoses:
Follow-up to #3159. Its two passes reached only `fem/derefmat_op.cpp` of the 18 TUs that need #2933: in `fem/pderefmat_op.cpp` the functor methods are called under `scf.if`/`scf.for`, and `AffineInlinerInterface::isLegalToInline(op, region)` only admits an affine op into a region whose parent is an affine scope or an affine loop/conditional, so a callee that starts with the aggregate `affine.load` stays out of line there.

Lowering affine before the second inline removes the affine ops from the callees, and `affine-cfg` right after `discard-unreferenced-linkonce` raises them back before mem2reg. On the pderefmat dump every `Run` folds into its callers, the parallels come back as `affine.parallel`, and mem2reg forwards all the fields.

Test: `inline_host_functor_nested.mlir`, the functor call under an `scf.if`.

### DIFF
--- a/src/enzyme_ad/jax/raise.cpp
+++ b/src/enzyme_ad/jax/raise.cpp
@@ -190,9 +190,9 @@ extern "C" std::string runLLVMToMLIRRoundTrip(std::string input,
     // the primal that carries the user's marker -- and any left behind
     // fail translation to LLVM IR.
     "lower-llvm-ext,"
-    "strip-dead-personality,"
+    "strip-dead-personality,lower-affine,"
     "inline{default-pipeline=canonicalize max-iterations=4},"
-    "discard-unreferenced-linkonce,"
+    "discard-unreferenced-linkonce,affine-cfg,"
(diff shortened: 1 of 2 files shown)


wsmoses:
Validation on the union without #2933: this gets `fem/pderefmat_op.cpp` through (16 of the 18 TUs remain, all TMOP metrics plus curlcurl_pa and normal_deriv_restriction, where the aggregate load is not behind a call at all), but the full sweep regresses 16 other TUs that compiled before (bilinearform_ext, hdiv_ea, mass_ea, prestriction, restriction, …). They all die the same way: a stack overflow in `affine-cfg`, `MoveLoadToAffine` → `AffineApplyNormalizer` recursing through the arith-expanded index chains that `lower-affine` leaves behind. Reproduces standalone with `--affine-cfg` on the pre-cfg dump of prestriction. That is a separate bug in the normalizer's recursion; fixing it first, then re-validating this.

PR #3161, opened 2026-09-08
Hoist loop-carried reshapes out of StableHLO while loops

vimarsh6739:
Loops that reshape a carried buffer on entry and flatten it again at the yield now carry the body shape directly. For example, LBM can carry `tensor<379200x128xf32>` instead of `tensor<48537600xf32>`. Every use of the carried argument in both the body and condition regions must be a reshape or bitcast-convert; having one matching reshape is insufficient. Arguments used only by bitcast-convert are eligible too. Input/output reshapes retain the original boundary types and zero-iteration behavior.

Add `ReshapeWhile` and a companion `ReshapeBitcastConvert` pattern that cancels `reshape(bitcast_convert(reshape(x)))` when the direct bitcast preserves the element grouping. This handles LBM's dynamic intermediate byte-view type while retaining the trailing byte dimension. Both patterns are registered in the default EnzymeHLOOpt pass, the transform dialect, and the C API reshape pipeline.

Validation: 93 related reshape, while, bitcast, and transform LIT tests pass. New tests check the resulting IR and interpret zero, one, and multiple iterations, two buffers, buffer swaps, bitcast-only and unused arguments, and rejection of mixed arithmetic users or a direct slice user in the condition. Bitcast cases cover narrowing, widening, equal widths, mismatched grouping, and shared results.

LBM validation on the development staging branch: both buffers carry `tensor<379200x128xf32>` through the while; only the two flag-slice reshapes remain inside the body. The runtime result matches the staging optimizer output. Source files were unchanged, and the before/after output is byte-identical (6,480,000 floats; max absolute difference from native CUDA: 5.96e-7).

On an RTX 5090, 20 timesteps over three alternating captures remain at 511 launches. Median summed GPU kernel time is 9.246 ms before versus 9.249 ms after (CUDA reference: 20 launches, 5.798 ms). Both optimized HLO bodies still contain 51 fusions; this change removes views but does not produce a measurable speedup on this case. GPU timings exclude compilation, transfers, and launch gaps.

The all-uses guard produces byte-identical LBM MLIR to the version measured above.

### DIFF
--- a/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
+++ b/src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
@@ -16877,6 +16877,147 @@ private:
   }
 };
 
+// Cancel reshape(bitcast_convert(reshape(x))) when the outer shape is exactly
+// the bitcast shape of x. In particular, a narrowing bitcast must retain its
+// extra innermost dimension, which groups the pieces of each source element.
+struct ReshapeBitcastConvert final
+    : CheckedOpRewritePattern<stablehlo::ReshapeOp, ReshapeBitcastConvert> {
+  using CheckedOpRewritePattern::CheckedOpRewritePattern;
+
+  // The intermediate bitcast may have a dynamic shape even when the source
+  // and final result statically determine the complete element grouping.
+  bool supportsDynamicShapes() const { return true; }
+
+  LogicalResult matchAndRewriteImpl(stablehlo::ReshapeOp op,
+                                    PatternRewriter &rewriter) const {
+    auto bitcast = op.getOperand().getDefiningOp<stablehlo::BitcastConvertOp>();
+    if (!bitcast || !bitcast->hasOneUse())
+      return failure();
+    auto reshape = bitcast.getOperand().getDefiningOp<stablehlo::ReshapeOp>();
+    if (!reshape)
+      return failure();
+    auto inputType = cast<RankedTensorType>(reshape.getOperand().getType());
+    auto resultType = cast<RankedTensorType>(op.getType());
+    if (!inputType.hasStaticShape() || !resultType.hasStaticShape() ||
+        !isa<IntegerType, FloatType>(inputType.getElementType()) ||
+        !isa<IntegerType, FloatType>(resultType.getElementType()))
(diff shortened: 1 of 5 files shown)


src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -16877,6 +16877,138 @@ struct WhileOpInductionReplacement
...
+      auto newType = cast<RankedTensorType>(reshape.getOperand().getType());
+      if (oldType == newType || !oldType.hasStaticShape() ||
+          !newType.hasStaticShape() ||
+          oldType.getElementType() != newType.getElementType())
+        continue;
+      Candidate candidate{static_cast<unsigned>(i), oldType, newType, {}};
+      for (Operation *user : body.getArgument(i).getUsers())
+        if (auto view = dyn_cast<stablehlo::ReshapeOp>(user))
+          if (view.getType() == newType)
+            candidate.inputViews.push_back(view);
```
wsmoses, line 16972: the condition here shouldn't be if there is a reshape argument of the ind var, but if all users of the ind var are reshapes [or the bitcast convert]

vimarsh6739: fixed

src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp
```diff
@@ -16877,6 +16877,147 @@ struct WhileOpInductionReplacement
...
+        continue;
+      // One matching reshape is not enough: other consumers would still need
+      // the old shape inside the loop. Check both regions, allowing bitcast
+      // users even when there is no direct reshape of the argument.
+      auto onlyViews = [](BlockArgument arg) {
+        return llvm::all_of(arg.getUsers(), [](Operation *user) {
+          return isa<stablehlo::ReshapeOp, stablehlo::BitcastConvertOp>(user);
+        });
+      };
+      if (!onlyViews(body.getArgument(i)) ||
```
wsmoses, line 16965: why have two checks here and also in the `for (Operation *user : body.getArgument(i).getUsers())` loop

PR #3174, opened 2026-09-09
simplify-affine-exprs: simplify against domains with symbols

wsmoses:
`simplify-affine-exprs` gists every access map and `affine.if` condition against the enclosing loop and branch domain with ISL, which is what folds `(%j + %i floordiv 3) floordiv 3` to `0` when `%i, %j` range over `[0, 3)`. It gave up whenever the domain carried a symbol, so any access under a loop with a symbolic bound was skipped. Every GPU kernel's grid loop has a symbolic bound, which left the pass dead on the raising path: on MFEM's `nonlininteg_vecconvection_pa` the shared-memory scratch accesses stay as

```
affine.store %v, %alloca[(%j + %i floordiv 3) floordiv 3, (%j + %i floordiv 3) mod 3, %i mod 3] : memref<4x3x3xf64>
```

under `affine.parallel (%i, %j) = (0, 0) to (3, 3)`, where `[0, %j, %i]` is the answer.

Map operands the domain does not know (an access's own symbol operand, or a dim no enclosing affine op binds) are appended to the constraint system and the ISL set as unconstrained variables, symbols get parameter identifiers on both spaces so the gist aligns them, semi-affine expressions (a division or modulus by a symbol) are left alone, and a simplified result that reads a domain variable the map has no operand for (which a domain equality can produce) is rejected instead of being decoded as position 0.

Test `simplify-affine-exprs-symbols.mlir`: a symbolic grid bound above the folding loop, an access with its own symbol operand, and a condition under a symbolic bound.

### DIFF
--- a/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
+++ b/src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
@@ -273,29 +273,35 @@ struct AffineExprToIslAffConverter {
       case mlir::AffineExprKind::Add:
         return isl_aff_add(lhs, rhs);
       case mlir::AffineExprKind::CeilDiv:
+        if (isl_aff_is_cst(rhs) != isl_bool_true)
+          break;
         return isl_aff_ceil(isl_aff_div(lhs, rhs));
       case mlir::AffineExprKind::FloorDiv:
+        if (isl_aff_is_cst(rhs) != isl_bool_true)
+          break;
         return isl_aff_floor(isl_aff_div(lhs, rhs));
       case mlir::AffineExprKind::Mod: {
-        if (isl_aff_is_cst(rhs) == isl_bool_true) {
-          isl_aff *r = isl_aff_mod_val(lhs, isl_aff_get_constant_val(rhs));
-          isl_aff_free(rhs);
-          return r;
-        } else {
-          isl_aff_free(lhs);
-          isl_aff_free(rhs);
-          return nullptr;
-        }
+        if (isl_aff_is_cst(rhs) != isl_bool_true)
+          break;
+        isl_aff *r = isl_aff_mod_val(lhs, isl_aff_get_constant_val(rhs));
+        isl_aff_free(rhs);
+        return r;
       }
       case mlir::AffineExprKind::Mul:
+        if (isl_aff_is_cst(lhs) != isl_bool_true &&
+            isl_aff_is_cst(rhs) != isl_bool_true)
+          break;
         return isl_aff_mul(lhs, rhs);
       default:
         LLVM_DEBUG(llvm::dbgs()
                    << "Unhandled kind " << (unsigned)bo.getKind() << "\n");
-        isl_aff_free(lhs);
-        isl_aff_free(rhs);
-        return nullptr;
+        break;
       }
+      // Semi-affine: a division or modulus by a symbol, or a product of two
+      // non-constants, has no isl affine form.
+      isl_aff_free(lhs);
+      isl_aff_free(rhs);
+      return nullptr;
     } else if (auto c = dyn_cast<AffineConstantExpr>(expr)) {
       return isl_aff_val_on_domain(isl_local_space_copy(ls),
                                    isl_val_int_from_si(ctx, c.getValue()));
@@ -586,6 +592,8 @@ struct IslToAffineExprConverter {
   unsigned symOffset;
   PosMapTy dimPosMap;
   PosMapTy symPosMap;
+  // Set when an expression has no isl form or decodes to no map expression.
+  bool incomplete = false;
 
   AffineExpr createOpBin(__isl_take isl_ast_expr *Expr) {
     AffineExpr LHS, RHS, Res;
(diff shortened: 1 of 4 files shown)


src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -273,29 +273,35 @@ struct AffineExprToIslAffConverter {
...
-          isl_aff_free(rhs);
-          return r;
-        } else {
-          isl_aff_free(lhs);
-          isl_aff_free(rhs);
-          return nullptr;
-        }
+        if (isl_aff_is_cst(rhs) != isl_bool_true)
+          break;
+        isl_aff *r = isl_aff_mod_val(lhs, isl_aff_get_constant_val(rhs));
```
ftynse, line 286: Nit: can't we just do `isl_aff_mod` without extracting the constant?

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -758,10 +766,16 @@ struct IslToAffineExprConverter {
...
     id = id - 1;
-    if (id < symOffset)
-      V = getAffineDimExpr(dimPosMap[id], mlirContext);
-    else
-      V = getAffineSymbolExpr(symPosMap[id - symOffset], mlirContext);
+    PosMapTy &posMap = id < symOffset ? dimPosMap : symPosMap;
+    auto pos = posMap.find(id < symOffset ? id : id - symOffset);
+    if (pos == posMap.end()) {
+      unmapped = true;
+      V = getAffineConstantExpr(0, mlirContext);
```
ftynse: We should rather assert here, we expect everything to be mapped.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -959,44 +980,33 @@ std::optional<AffineMap> handleAffineValueMap(IslAnalysis &islAnalysis,
...
-
-  if (avm.getNumSymbols() != 0 || cst.getNumSymbolVars() != 0) {
-    // TODO While the fact that all dims from the map _must_ appear in the cst,
-    // this is not the case for symbols. We do not handle that case correctly
-    // currently, thus we abort early.
-    domain = isl_set_free(domain);
-    return {};
+  unsigned numDomainDims = cst.getNumDimVars();
+  unsigned numDomainSymbols = cst.getNumSymbolVars();
+  for (unsigned origDim = 0; origDim < map.getNumDims(); origDim++) {
```
ftynse: Remember the LLVM coding style requirement to never evaluate the loop upper bound in the loop

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
```diff
@@ -1005,11 +1015,22 @@ std::optional<AffineMap> handleAffineValueMap(IslAnalysis &islAnalysis,
     isl_id *id = isl_id_alloc(ctx, "dim", (void *)(size_t)(i + 1));
     space = isl_space_set_dim_id(space, isl_dim_set, i, id);
   }
+  // isl aligns the parameters of the expression and the domain by name, and
+  // refuses to add parameters to a set whose existing ones are unnamed.
```
ftynse: Is it _only_ the name, or name + user payload (the `void *` pointer argument).

ftynse (approved):

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: isl only has `isl_aff_mod_val` (aff mod val); there is no aff-by-aff mod in the API, so the constant has to come out of the rhs aff. Kept as is.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Hoisted (both loops).

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: Both: isl interns identifiers by name and user pointer together, and aligns parameters by identifier identity. The two spaces get ids built from the same name and payload, so they are the same isl_id objects. Comment updated to say so.

src/enzyme_ad/jax/Passes/SimplifyAffineExprs.cpp
wsmoses: It is reachable: with an abort in that branch, 6 MFEM TUs crash (five tmop metrics files and diffusion_kernels). The case is a domain equality between two loop ivs, e.g. a single-trip inner loop `affine.for %j = %i to %i + 1`: the gist of `%m[%j]` comes back reading `%i`, which the map had no operand for (tmop/metrics/315 alone hits it 52 times). Rather than reject those, the map is now extended: every domain variable decodes to a position past the map's own, and the ones a result reads are appended as operands (`handleAffineValueMap` returns a value map; both callers set the operands). The id lookup is an assert now, as you asked. New test case `@single_trip`.

PR #3179, opened 2026-09-09
Raising: drop the trunc and shift extent derivations

wsmoses:
Follow-up to #3178. Now that canonicalize-parallel unpacks the dim3 replication before the raiser runs, the extent derivation sees the launch dimension itself and no longer meets the `trunc` and `shrui` that read the packed halves back; the dim3 special case in the trunc arm and the two generic arms go with it.

Probe on the union with these arms switched off: all 25 runtime-dispatch TUs still compile, and the raised IR (loop count and batched-axis widths per TU) is compared against the same library with the arms on; that comparison is reported in the PR comments as it completes.


wsmoses:
Shape check done: with these two arms switched off (same library, on the union with #3178), all 25 runtime-dispatch TUs compile and their raised IR is identical, loop for loop and batched-axis width for batched-axis width, to the run with the arms on.

PR #3181, opened 2026-09-09
Add Z3 to Bazel build rules

javathunderman:
May want to make this optional but opening a PR to eventually stage #2429 for merging


javathunderman:
Made Z3 an optional dependency via `--define=enable_z3=false` similar to Triton

.github/workflows/build.yml
```diff
@@ -135,6 +135,11 @@ jobs:
 
       - name: Build enzymexlamlir-opt
         timeout-minutes: 180
+        runs-on: macOS-latest
+        steps:
+          - uses: actions/checkout@v4
+          - name: Setup tmate session
```
giordano: You only need this step, nothing from above

giordano: Also, the step goes below, with the other steps

.github/workflows/build.yml
```diff
@@ -132,7 +129,9 @@ jobs:
       - name: Update Python requirements files
         run: |
           bazel run $BAZEL_FLAGS --color=yes //builddeps:requirements.update
-
+      - uses: actions/checkout@v4
```
giordano: As I said above, you don't need this, you already check out the repo
```suggestion
```

PR #3195, opened 2026-09-11
Add derivative rules for stablehlo CaseOp

### LINKED ISSUE #2939: Missing autodiff rules for stablehlo.case: reverse fails with “could not compute the adjoint”, forward with “RegionBranchOpInterface not implemented”
### Environment

**Environment** — Julia 1.12.6 x86_64-linux-gnu · Reactant 0.2.280 · Reactant_jll 0.0.405+0 · Enzyme 0.13.199 · EnzymeCore 0.8.21 · Enzyme_jll 0.0.290+0 · Backend CPU (PJRT) · measured 2026-08-24

### Reproducer

```julia
using Reactant, Enzyme

f_case(x, idx) = Reactant.Ops.case(idx,
                     [a -> sum(a .^ 2),   # branch 0
                      a -> sum(a .^ 3),   # branch 1
                      a -> sum(a .^ 4)],  # branch 2
                     x; track_numbers = Number)

g = (x, s) -> Enzyme.gradient(Enzyme.Reverse, f_case, x, Enzyme.Const(s))[1]

Reactant.@compile sync=true g(Reactant.ConcreteRArray([0.6,0.7,0.8,0.9,1.0,1.1]),
                              Reactant.ConcreteRNumber(0))
```

### Observed — reverse

```
CompilationError: MLIR pass pipeline "all" failed
error: could not compute the adjoint for this operation
%1:2 = "stablehlo.case"(%0) ({ ... }, { ... }, { ... })
     : (tensor<i32>) -> (tensor<6xf64>, tensor<f64>)
  at Reactant/src/Ops.jl:2927  (Ops.case)
```

### Observed — forward

```
error: RegionBranchOpInterface not implemented for ...
```

All three branches fail, as does an out-of-range index that falls through to the last branch. It is a hard failure at compile time, not a silently wrong gradient — checked against analytic gradients on every branch.

### Expected

The analytic gradients `2x`, `3x^2`, `4x^3`. Two control constructs in the same script already achieve exactly this, at relative error 0:

| Construct | Emits | Reverse mode |
|---|---|---|
| `@trace if` (2-arm) | `stablehlo.if` | exact, rel err 0 |
| `@trace if` (3-arm) | 2x `stablehlo.if` | exact, rel err 0 |
| `ifelse` | `stablehlo.select` | exact, rel err 0 |
| `Ops.case` | `stablehlo.case` | **fails** |

So the gap is narrow and the surrounding constructs all work — which is what makes implementing the rule tractable.

> Already tracked, but unchecked and with no reproducer: `CaseOp` appears on Enzyme-JAX #88, "Tracking issue for missing HLO derivatives". Behaviour is identical on 0.2.274, so this is a long-standing gap rather than a regression. Precedent is #1579 (`stablehlo.sort`), closed by implementing the rule after a concrete case arrived.

awiteck:
Closes #2939 
- `AutoDiffCaseFwd`, `AutoDiffCaseCF`, `AutoDiffCaseRev`, and the associated `CaseOpEnzymeOpsRemover`. Basically just generalizing the existing IfOp case to multi-region
- Remove the `CaseOp` stub from `StableHLODerivatives.td`. It auto-registered the generic forward handler, which requires `RegionBranchOpInterface`; StableHLO's `CaseOp` does not implement it
-  Tests for forward, reverse, and the ops remover (`case.mlir`, `case_remove.mlir`)

### DIFF
--- a/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
+++ b/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp
@@ -450,6 +450,39 @@ public:
   }
 };
 
+class AutoDiffCaseFwd
+    : public AutoDiffOpInterface::ExternalModel<AutoDiffCaseFwd,
+                                                stablehlo::CaseOp> {
+public:
+  LogicalResult createForwardModeTangent(Operation *orig, OpBuilder &builder,
+                                         MGradientUtils *gutils) const {
+    llvm::SmallDenseSet<unsigned> operandPositionsToShadow;
+    llvm::SmallDenseSet<unsigned> resultPositionsToShadow;
+
+    for (auto res : orig->getOpResults()) {
+      if (!gutils->isConstantValue(res))
+        resultPositionsToShadow.insert(res.getResultNumber());
+    }
+    return mlir::enzyme::detail::controlFlowForwardHandler(
+        orig, builder, gutils, operandPositionsToShadow,
+        resultPositionsToShadow);
+  }
+};
+
+class AutoDiffCaseCF
+    : public ControlFlowAutoDiffOpInterface::ExternalModel<AutoDiffCaseCF,
+                                                           stablehlo::CaseOp> {
+public:
+  Operation *createWithShadows(Operation *op, OpBuilder &builder,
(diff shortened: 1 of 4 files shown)


Pangoraw:
cool thanks 💯

PR #3197, opened 2026-09-14
Raise to DUS/DS affine store/load with multi ivs per dim

Pangoraw:
consider

```mlir
module {
  func.func @main(%arg9: memref<20x1536x3072xf32, 1>, %arg0: memref<20x1536x3072xf32>) {
    affine.parallel (%arg11, %arg12, %arg13, %arg14) = (0, 0, 0, 0) to (4, 16, 95, 3056) {
      %51 = affine.load %arg9[%arg11 + 9, %arg12 + %arg13 * 16 + 9,  %arg14 + 8] : memref<20x1536x3072xf32, 1>
      %52 = arith.mulf %51, %51 : f32
      affine.store %52, %arg0[%arg11, %arg12 + %arg13 * 16 + 9, %arg14 + 8] : memref<20x1536x3072xf32>
    }
    return
  }
}
```

which we could not raise before this PR, if mergeparallelinductions where to kick in we could raise it. Now the raising identifies induction variables that can be merged


wsmoses:
This is obviously good and helpful and we should merge, but this doesn’t explain what broke elsewhere causing the need for it, as it wasn’t previously required.

Eg didn’t merge inductions used to fire for this?

wsmoses:
(I mostly raise this to make sure we don’t neglect the other root cause, which may perhaps be harder to find/isolate once this is merged)

Pangoraw:
Yes the merge does not apply because being in the affine if, there are cases where one the two values to merge are replaced with a constant because it's a single value within that affine if branch.

wsmoses:
Do you have a mwe of the affine if part?

Pangoraw:
yes based on the GB IR (just run `enzymexlamlir-opt %s --affine-cfg`):

```mlir
module {
  func.func @split_blocks_merge(%m: memref<1536x3072xf32, 1>, %a: f32, %b: f32) {
    %true = arith.constant true
    %false = arith.constant false
    %cst = arith.constant 0.000000e+00 : f32
    affine.parallel (%i, %j, %k) = (0, 0, 0) to (16, 95, 3056) {
      // The halo masks: `%i + %j * 16` in [0, 2] and <= 1517.
      %lo = affine.if affine_set<(d0, d1) : (d0 + d1 * 16 + 1 >= 0, -d0 - d1 * 16 + 2 >= 0)>(%i, %j) -> i1 {
        affine.yield %false : i1
      } else {
        affine.yield %true : i1
      }
      %hi = affine.if affine_set<(d0, d1) : (-d0 - d1 * 16 + 1517 >= 0)>(%i, %j) -> i1 {
        affine.yield %true : i1
      } else {
        affine.yield %false : i1
      }
      %interior = arith.andi %lo, %hi : i1
      %sel = arith.select %interior, %a, %b : f32
      %c = arith.cmpf olt, %cst, %sel : f32
      %r = scf.if %c -> f32 {
        %l = affine.load %m[%i + %j * 16 + 8, %k + 8] : memref<1536x3072xf32, 1>
        scf.yield %l : f32
      } else {
        %l = affine.load %m[%i + %j * 16 + 9, %k + 8] : memref<1536x3072xf32, 1>
        scf.yield %l : f32
      }
      // The bulk of the body: these are the accesses that lose their slice when
      // the pair fails to merge.
      %p = affine.load %m[%i + %j * 16 + 7, %k + 8] : memref<1536x3072xf32, 1>
      %q = arith.addf %r, %p : f32
      affine.store %q, %m[%i + %j * 16 + 8, %k + 8] : memref<1536x3072xf32, 1>
    }
    return
  }
}

Pangoraw:
On previous main, with the split notice how a few loads only reference arg3 and not arg4 because of the affine constraint propagation:

```mlir
#set = affine_set<(d0, d1) : (-d0 - d1 * 16 + 1517 >= 0)>
#set1 = affine_set<(d0, d1) : (d0 + d1 * 16 + 1 >= 0, -d0 - d1 * 16 + 2 >= 0)>
module {
  func.func @split_blocks_merge(%arg0: memref<1536x3072xf32, 1>, %arg1: f32, %arg2: f32) {
    %cst = arith.constant 0.000000e+00 : f32
    affine.parallel (%arg3, %arg4, %arg5) = (0, 0, 0) to (16, 95, 3056) {
      %0 = affine.if #set(%arg3, %arg4) -> f32 {
        %3 = affine.if #set1(%arg3, %arg4) -> f32 {
          %4 = arith.cmpf olt, %cst, %arg2 : f32
          %5 = scf.if %4 -> (f32) {
            // CANNOT BE MERGED BECAUSE OF HERE
            %6 = affine.load %arg0[%arg3 + 8, %arg5 + 8] : memref<1536x3072xf32, 1>
            scf.yield %6 : f32
          } else {
            // CANNOT BE MERGED BECAUSE OF HERE
            %6 = affine.load %arg0[%arg3 + 9, %arg5 + 8] : memref<1536x3072xf32, 1>
            scf.yield %6 : f32
          }
          affine.yield %5 : f32
        } else {
          %4 = arith.cmpf olt, %cst, %arg1 : f32
          %5 = scf.if %4 -> (f32) {
            %6 = affine.load %arg0[%arg3 + %arg4 * 16 + 8, %arg5 + 8] : memref<1536x3072xf32, 1>
            scf.yield %6 : f32
          } else {
            %6 = affine.load %arg0[%arg3 + %arg4 * 16 + 9, %arg5 + 8] : memref<1536x3072xf32, 1>
            scf.yield %6 : f32
          }
          affine.yield %5 : f32
        }
        affine.yield %3 : f32
      } else {
        %3 = arith.cmpf olt, %cst, %arg2 : f32
        %4 = scf.if %3 -> (f32) {
          // CANNOT BE MERGED BECAUSE OF HERE
          %5 = affine.load %arg0[%arg3 + 1512, %arg5 + 8] : memref<1536x3072xf32, 1>
          scf.yield %5 : f32
        } else {
          // CANNOT BE MERGED BECAUSE OF HERE
          %5 = affine.load %arg0[%arg3 + 1513, %arg5 + 8] : memref<1536x3072xf32, 1>
          scf.yield %5 : f32
        }
        affine.yield %4 : f32
      }
      %1 = affine.load %arg0[%arg3 + %arg4 * 16 + 7, %arg5 + 8] : memref<1536x3072xf32, 1>
      %2 = arith.addf %0, %1 : f32
      affine.store %2, %arg0[%arg3 + %arg4 * 16 + 8, %arg5 + 8] : memref<1536x3072xf32, 1>
    }
    return
  }
}

wsmoses:
Why was this able to be merged earlier? Eg I feel like we may need to still generalize some things

Pangoraw:
In the initial mwe, `%i` and `%j` are always used in pair `%i + %j * 16` with `%i in 0:15` so it uses

https://github.com/EnzymeAD/Enzyme-JAX/blob/d9edce8fab2bb760df761de465fe7bdae93da0b1/src/enzyme_ad/jax/Passes/AffineCFG.cpp#L5292-L5294

to collapse to a single iv in `%new_iv in 0:1519`

wsmoses:
Yeah I guess do we have the code before that invariant is broken?

wsmoses:
Ah it is above. Okay @Pangoraw go ahead and open a PR to revert the affine split, and open an issue for all this info and I will figure out how/if to re land.

Though also did GB ci here break (if not why)

Pangoraw:
GB CI is fine (unless for the expected TPU F64 failure) here because with this PR, the affine loads / stores are still loaded to ds / dus. #3199 mitigates the split when not it can be non-beneficial.

wsmoses:
Yeah it was more a question of did that PR that broke it initially also break GB ci and we didn't notice or is ci insufficient to catch (it was in the rush so may have been missed)

Pangoraw:
CI will still be green if there are less than 1000 collectives of the different kinds so it never went red. I added so that it will fail if there are any gather / scatter.

PR #3199, opened 2026-09-14
Bail out of if splitting if any operand is a float

Pangoraw:
while #3197 fixes the scatter / gather emission for https://github.com/EnzymeAD/Reactant.jl/issues/3276, the splitonaffineifconstants may not be beneficial if there are non integer in the chain since it will not recover an affine expr anyway. In gb the affine if is downstream of a cmpf with a runtime value.


wsmoses (commented):
For now I'm okay just disabling thr pattern entirely (or maybe we can temporarily flag gate the pattern for debugging ease, defaulting to off)

Pangoraw:
superseded by #3203

PR #3201, opened 2026-09-15
EnzymeHLOOpt: hoist unchanged while variables initialized from a block argument's layout

wsmoses:
`WhileSimplify` replaces an iteration variable the body never updates by its initial value when that value is a constant or a function argument. A raised kernel carries the layout ops on its arguments (`bitcast_convert(reshape(%arg))`) through its loop, and a nested loop (a raised dot product, mfem's `Dot(x, x)`) starts its vector variables from the enclosing loop's, so the untouched ones stayed loop-carried at every level and came back as new loop results.

This adds the special case: an initial value that is a block argument (of the function, or of an enclosing loop) seen through `reshape` / `bitcast_convert` counts as hoistable. Once hoisted, the layout ops at the return cancel (`reshape(bitcast_convert(bitcast_convert(reshape(%arg))))` folds) and the argument is returned as itself, which is what the Reactant runtime (EnzymeAD/Reactant.jl#3285) keys on to skip donating and returning a read-only argument. Without it, the MFEM Linear Form Extension test hands `Dot(x, x)` the same buffer twice, both copies look written, and XLA refuses to donate the buffer twice.

Tests `while_simplify_pure_init.mlir` (single loop) and `while_simplify_nested_init.mlir` (nested loop). One golden changes (`while_checkpointing_binomial_dynamic.mlir`): a checkpoint segment's unchanged variable, initialized from the outer loop's, is hoisted, and the outer variable it fed goes with it.


Pangoraw (commented):
I think in stablehlo side we can always hoist, unlike in the HLO IR where the cond and body region are closed to the outside. stablehlo -> hlo does the patch up for us

Issue #2999, opened 2026-08-28
Reverse-mode gradient through a `@trace mincut=true` loop got ~30× slower on GPU with Reactant ≥ 0.2.281 (reverse-loop work is batched over the whole tape)


giordano:
_Note: below is a (very) verbose AI-generated summary of a massive performance slowdown we observed in Breeze.jl when moving from Reactant v0.2.278 to v0.2.281.  Look only at the reproducer if you want to skip the noise, but the analysis of the increased dynamic slices may still be useful_

Between Reactant **0.2.278** (Reactant_jll 0.0.401) and **0.2.281** (Reactant_jll 0.0.406) the compiled Enzyme reverse-mode gradient of a time-stepping loop in Breeze.jl ([`test/reactant/advection_diffusion_ad.jl`](https://github.com/NumericalEarth/Breeze.jl/blob/7330145a1e07446bf666f4c4ea6a59125893e388/test/reactant/advection_diffusion_ad.jl), `CompressibleDynamics` + WENO5, `N = 64`, `Nₛ = 229` steps, `CompileOptions(raise_first=true, raise=true, sync=true)`) went from **21 s to 695 s of execution** on an RTX 5090 (compile: 286 s → 404 s), with results identical to all printed digits.

Everything on the Julia side is the same (Julia 1.12.7, Enzyme 0.13.199, EnzymeCore 0.8.21, same Oceananigans/Breeze, environments differ *only* in `Reactant` + `Reactant_jll`; the Reactant pass pipeline is identical modulo 32 added patterns — `cse_*`, `select_to_logical`, `while_scatter_accumulator_no_add` — that were shown not to matter: the 0.2.278 pattern list run in the new jll gives the same result). Both causes are in the jll (Enzyme-JAX 568fe05 → aea4857):

1. **(dominant, ≥ 30×)** in the post-AD pattern set, `greedy_while_loop_batch_fission` now hoists the reverse-step arithmetic out of the reverse `stablehlo.while` and batches it over the whole tape (removing either that pattern or `while_op_induction_replacement` from the set prevents it, see bisection below). In the Breeze case this creates **1 567 distinct full-tape (`229×64×64`) tensors** (XLA buffer assignment 0.35 GB → 10.66 GB) and a reverse loop that `dynamic_slice`s all of them each iteration (47 → 1 580 `dynamic_slice`s). The same StableHLO also ran for ~39 min (killed) on the *old* XLA.
[...]

PR #3208, opened 2026-09-16
Check for auto batching profitability

Pangoraw:
Fixes part 1 of #2999. Profitability is defined as either all the results from a DAG coming from a candidate DS being hoisted out, or the hoist having a single result. So instead of creating 7 cache sized array (229x64x64) tensors like in the #2999 ir, it bails on keeps the cache untouched.

Superseds #3049.


src/enzyme_ad/jax/Passes/AutoBatching.cpp
```diff
@@ -883,6 +883,607 @@ static bool definedOutside(Value v, Operation *op) {
...
+static bool avoidBatching(Operation *op) {
+  if (!op) {
+    return true;
+  }
+
+  return llvm::TypeSwitch<Operation *, bool>(op)
+      .Case<stablehlo::ReshapeOp, stablehlo::SliceOp, stablehlo::ReturnOp,
+            // avoid ops that use SHLOGenericBatchOpInterface since that
+            // lowers to loop
+            stablehlo::IfOp, stablehlo::CaseOp, stablehlo::WhileOp,
```
wsmoses, line 1118: these have an exception where if the condition is itself batchable it's okay

wsmoses (approved):
I'm fine with this as is, but there are cases where not reaching the store is useful -- for example if it moves up an elementwise up that can now be fused with the producer outside the while

Pangoraw:
> but there are cases where not reaching the store is useful

yes, here if we consider a dag

```
A -> B --> C
       \
        -> D
```

then it will batch A and B outside the loop. The main prevention is not to hoist C and D to double the size of big arrays are dynamically sliced into.

PR #3209, opened 2026-09-16
Drop result layouts from unused result for kernel/jit calls

Pangoraw:
Interestingly, it seems XLA is allowed to choose another layout when performing buffer assignment. We never hit this because we insert transposes before the kernel calls so it was enough to prevent it until https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl/pull/160#issuecomment-5623068541.


wsmoses (commented):
why is the dead result not just removed?

Pangoraw:
yes it is removed, but when reconstructing the op without this result, we used to pass `nullptr` for the result layouts which leads to an error since you cannot have only operand layouts without result layouts. This now drops the results but also reattaches the existing result layouts for results which survive.

wsmoses:
hm I'm still not quite following, if non urgent can we discuss next week in a call when back in the US?

gdalle:
Hey @Pangoraw, thanks for taking a look so quickly!
I had my agent run our tests against this branch and it seems this is necessary but not sufficient, although I'm unsure about the exact reasons. Just thought I'd let you know, and you have the detailed analysis in https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl/pull/160#issuecomment-5758711527 if you're interested :)

Pangoraw:
yes, as your agent pointed out. You will need to attach the attributes on the Reactant side as well with a diff similar to the one in https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl/pull/160#issuecomment-5624373404. This fixes a canonicalization bug where it was previously not possible to attach them.

Pangoraw:
As discussed during the call, i added the result layout generation from the operands layout.

gdalle:
> yes, as your agent pointed out. You will need to attach the attributes on the Reactant side as well with a diff similar to the one in https://github.com/JuliaDecisionFocusedLearning/CoolPDLP.jl/pull/160#issuecomment-5624373404.

Alright, thanks! Just let me know when all necessary fixes are available on branches, I'll run the test again for my linear programming use case :)

PR #3211, opened 2026-09-18
Make DUSDUSSubsuming take precedence over SliceOfDynamicUpdate

### LINKED ISSUE #3207: enzyme-hlo-opt livelocks: DUSDUSSubsuming and SliceOfDynamicUpdate are exact inverses
`--enzyme-hlo-opt` never terminates on IR containing a DUS-of-DUS chain with a partially-overlapping slice user. Two patterns are exact inverses of each other and flip the IR back and forth forever.

Found while compiling a reverse-mode-AD'd Oceananigans-style kernel; the halo-exchange shape that AD produces (`1x18x37` updates into `60x28x47`) hits it reliably.

Reproduced on `2f3e6187` with `src/enzyme_ad/jax/Passes/EnzymeHLOOpt.cpp` unmodified.

## Reproducer

```mlir
// enzymexlamlir-opt --enzyme-hlo-opt %s     <-- hangs
func.func @main(%X: tensor<60x28x47xf32>, %u: tensor<1x18x37xf32>) -> (tensor<60x28x47xf32>, tensor<2x18x37xf32>) {
  %c0 = stablehlo.constant dense<0> : tensor<i32>
  %c1 = stablehlo.constant dense<1> : tensor<i32>
  %c50 = stablehlo.constant dense<50> : tensor<i32>
  %a = stablehlo.dynamic_update_slice %X, %u, %c1, %c0, %c0 : (tensor<60x28x47xf32>, tensor<1x18x37xf32>, tensor<i32>, tensor<i32>, tensor<i32>) -> tensor<60x28x47xf32>
  %b = stablehlo.dynamic_update_slice %a, %u, %c50, %c0, %c0 : (tensor<60x28x47xf32>, tensor<1x18x37xf32>, tensor<i32>, tensor<i32>, tensor<i32>) -> tensor<60x28x47xf32>
  %s = stablehlo.slice %a [0:2, 0:18, 0:37] : (tensor<60x28x47xf32>) -> tensor<2x18x37xf32>
  return %b, %s : tensor<60x28x47xf32>, tensor<2x18x37xf32>
}
```

`%a` writes rows `[1:2]`, `%b` writes rows `[50:51]`, and `%s` reads rows `[0:2]` — partially overlapping `%a`'s update, disjoint from `%b`'s.

## The cycle

1. **`DUSDUSSubsuming`** (`EnzymeHLOOpt.cpp:19642`) — its `movedSlices` loop clones every `stablehlo.slice` user of `dus2` onto `dus` and replaces it, hoisting `slice(%a)` → `slice(%b)`.
2. **`SliceOfDynamicUpdate`** (`EnzymeHLOOpt.cpp:2620`), the "no overlap" branch at `EnzymeHLOOpt.cpp:2707` — `%s` provably doesn't overlap `%b`'s update at row 50, so it sinks `slice(%b)` → `slice(%a)`.
3. Back to step 1. Both return `success()`, so the greedy driver never reaches a fixed point.

The *partial* overlap with the inner DUS is what makes it stick: full overlap lets `SliceOfDynamicUpdate` take its "slice is inside the update" branch and settle on `%u`; zero overlap lets the slice sink past both DUSes to `%X` and settle there. Only the partial case leaves both patterns permanently willing to fire.

`max_iterations` does not help — the spin is inside a single `processWorklist` call, not across greedy iterations.

## Why it's fatal rather than just a livelock

Each revolution costs ~5 ms, because `DUSDUSSubsuming` runs a full Presburger provenance walk (`computeTensorValueProvenanceImpl`) before it can decide anything. On the real module the process ran >25 min with 100% of `sample(1)` frames under `DUSDUSSubsuming::matchAndRewriteImpl` → `computeTensorValueProvenanceImpl` → `PresburgerRelation::intersectDomain`.

## Evidence on the real module (379k lines, 3190 DUS ops)

| Configuration | Result |
| --- | --- |
| `--enzyme-hlo-opt` | >25 min, no completion |
| `passses=512 max_iterations=1` | >10 min (vs. `passses=0` baseline: 8.8 s) |
| same, `DUSDUSSubsuming` stubbed to `failure()` | **8.8 s** |
| full `--enzyme-hlo-opt`, `DUSDUSSubsuming` stubbed | **20.9 s** |

Per-pattern success counters on the real module, after 200k rewrites:

```
  84092  SliceOfDynamicUpdate
  83567  DUSDUSSubsuming
   8909  SliceConcat
   8880  ConcatConstProp
   8741  SliceSlice
    811  NoopSlice
```

Near-lockstep, with everything else two orders of magnitude lower. On the reproducer, disabling *either* one terminates; with both enabled it spun through 52963 `DUSDUSSubsuming` invocations in 30 s.

## Suggested fix

Make `DUSDUSSubsuming` not hoist a slice that `SliceOfDynamicUpdate` will immediately sink back — i.e. skip the hoist when the slice provably does not intersect `dus`'s update region. In that case the hoisted slice reads identical data either way, so the move is pure churn. Alternatively, gate the hoist on it actually enabling the subsumption rewrite, rather than performing it unconditionally.

## Secondary issue (real, but not the cause)

`computeTensorValueProvenanceImpl` is recomputed from scratch for every DUS in a chain, and each walk is linear in chain length — measured 133 ms at chain length 50; the longest chain in the module is 87. That makes `DUSDUSSubsuming` O(N²) over a chain even when it does terminate. Disjunct counts stay small (~15–25), so `coalesce()` is doing its job; the cost is the per-step Presburger work, not disjunct blowup.

Pangoraw:
Closes #3207.


wsmoses:
sorry I'm not sure this is the right fix

wsmoses:
specifically the error here, imo is that DUSDUSSubsuming and SliceOfDynamicUpdate should never be inserted at the same time. The SliceOfDynamicUpdate is a simplification that should be run early, DUSDUSSubsuming is a pattern which should be run at the end to avoid buffer copies

wsmoses:
@Pangoraw can you revert this and do a different fix

Pangoraw:
Do you mean we should add not run SliceOfDynamicUpdate in the final transform-interpreter run in the pipeline? or one more opt run with DUSDUSSubsuming and not SliceOfDynamicUpdate

wsmoses:
yeah I'm not exactly sure how we want to set it up but we shouldn't have them in the same transform interp for sure [dusdussubsuming is really a late stage pattern]

PR #3218, opened 2026-09-20
Raising: mixed batched/scalar inductions in one index, and one-element dimensions in a flattened pair

wsmoses:
Two MFEM kernels regressed with #3197 (multi-induction accesses raised as slices); both compiled before it. Found validating the full MFEM suite on current main: one translation unit no longer compiles, and three others embed modules that do not parse at run time.

**Crash: "requires the same shape for all operands" in the masked store's select** (`fem/normal_deriv_restriction.cpp`, `L2NormalDerivativeFaceRestriction::Mult3D<1>`). The kernel stores into a `2x1` scratch at `[lane + t, 0]`; in context the lane loop is sized to one iteration and `t` is a thread of two. `getExpandedAffineDims` sorts the pair by step and checks the row-major invariant, and the one-element dimension's step (1) failed it, so the update was sized by the pair's range (2) while the mask kept the lane's extent (1), and `stablehlo.select` could not be built. A one-element dimension spans nothing and constrains no step, so the check skips it; the pair then expands to `[1, 2]` and the mask aligns.

**Invalid IR: a `dynamic_slice` whose start is a `tensor<8xi64>`** (`bilininteg_{mass,diffusion,convection}_ea.cpp`, `EAMassAssemble2D<8,8>`):
```
error: 'stablehlo.dynamic_slice' op operand #3 must be variadic of 0D tensor of 2/4/8/16/32/64-bit integer values, but got 'tensor<8xi64>'
```
The kernel reads a shared scratch at `[t + k * 8]`, `t` a thread of eight and `k` a sequential reduction loop. `needsGeneralScatterGather` now accepts a contiguous pair, and this pair is contiguous, but `k` is scalar per iteration: the access is a per-lane index (each lane reads its own row), which only the gather expresses. The slice path picked the scalar induction as "the" induction of the expression and emitted the whole expression as the start, a tensor over the lanes. A contiguous pair is one slice only when both inductions are batched; otherwise the access goes to the gather/scatter path, as it did before #3197.

Tests are the two kernels, kept whole with declarations for what they call (as `kernelcast.mlir` does); the check is that they raise, plus that no lane-shaped start remains. I could not reduce either to a hand-written loop nest that still triggers the path: the lane sizing and the shared scratch come from the kernel's context.

**Third case: a flattened pair as a gather/scatter index column** (`bilininteg_{mass,diffusion}_kernels.cpp`, `SmemPAMassApplyTriangle<N,N>`; the PA Simplices test):
```
error: size of operand dimension 0 (2) is not equal to 1 or size of result dimension 0 (4)
```
A dynamic-index store whose index was itself loaded through a contiguous pair of thread inductions (`t + u * 2`) reaches the scatter path with the pair as one flattened index column. `buildGatherScatterIndices` keyed the column on one induction, so a stored value varying along the other looked like a racy write, was reduced over it, and was broadcast from 2 into the flattened 4. The index columns are now expanded to one dimension per induction first, as #3197 already does for the update. The kernel module is too large for a whole-kernel test (its constant tables alone run to 100k lines); the fix is covered by the MFEM suite (74 of 74 with all three).


src/enzyme_ad/jax/Passes/AffineToStableHLORaising.cpp
```diff
@@ -257,9 +257,10 @@ static bool isContinuousAlongIVs(affine::AffineValueMap map, AffineExpr E) {
   return computeExprRange(map, E).has_value();
 }
 
-// has single (or zero) iv per dim.
+// has single (or zero) iv per dim, or a contiguous pair of batched ivs.
 // iv are present only at one dim.
-static bool needsGeneralScatterGather(affine::AffineValueMap accessValueMap) {
+static bool needsGeneralScatterGather(affine::AffineValueMap accessValueMap,
+                                      function_ref<bool(Value)> isBatchedIV) {
```
Pangoraw: pass parallelcontext here instead of the same closure twice?

wsmoses: Done: needsGeneralScatterGather now takes the ParallelContext (moved below the struct definition) and both call sites pass pc.

wsmoses: Reworked without moving the function: a forward declaration of ParallelContext and a one-line isParallelIV(pc, iv) helper defined after the struct; the diff is back to the original hunks.

PR #3220, opened 2026-09-20
EnzymeHLOOpt: SelectCompIotaConstSimplify accepts a constant prefix mask

wsmoses:
`SelectCompIotaConstSimplify` already rewrites `select(compare(iota [+ offset], K), a, b)`, with or without a broadcast of the predicate, into the concatenation of the operands' slices on either side of K; `SelectPadToDUS` handles a pad predicate. Neither accepts a dense constant predicate, which is what the lane mask `lane < k` of an unrolled block tree reduction becomes once the block size is a constant (#3219).

This generalizes the existing pattern instead of adding one: a constant predicate that, along one dimension, is a run of one value followed by a run of the other and does not vary along the remaining dimensions is recognized as `iota_dim < K` (prefix true) or `iota_dim >= K` (prefix false), and the pattern's rewrite is shared through a `rewriteAsSlices` helper. The compare path is unchanged. On the Linear Form Extension dot-product kernel, with #3221 and #3222, the unrolled tree becomes a single `dot_general`.

Test `select_constant_prefix_mask.mlir`: prefix, suffix, a column mask of a 2-D tensor, and an alternating mask that must stay.


wsmoses:
Yes: SelectCompIotaConstSimplify handles compare(iota, K) predicates (and SelectPadToDUS a pad), neither a dense constant. Reworked as a generalization of SelectCompIotaConstSimplify: the constant is recognized as iota_dim < K or >= K and the existing rewrite is shared; no new pattern.

PR #3221, opened 2026-09-20
EnzymeHLOOpt: GatherOpCanon folds a constant run that passes the operand's end

wsmoses:
`GatherOpCanon`'s iota analysis already turns a gather of a 1-D operand at consecutive or strided constant indices into a slice. It refused any run that reaches past the operand's end, because the gather clamps those indices (`if (limit > dim) return failure(); // gather clamps indices`). For a single mapped grid dimension the clamped part of the run reads the last element that many times, so the gather is the in-bounds slice followed by a `broadcast_in_dim` of the last element, or that broadcast alone when nothing is in bounds. `analyzeGridIndexing` now reports whether the run fits (`inBounds`) instead of failing; the scatter rewrite and the multi-dimensional gather case keep requiring it, only the single-mapping gather case handles the clamp (positive stride, non-negative start).

Where it comes from: the reads `buf[lane + i]` of mfem's block tree reduction, once the block size is a constant (#3219) and the loop is unrolled, are constant runs for every lane, and the lanes past the halfway point run past the scratch. With #3220 (the mask) and this, the existing `slice(x,[0:n/2]) + slice(x,[n/2:n])` recognition takes the whole unrolled tree to a chain of `reduce_window`s and #3222 to one `reduce`: the Linear Form Extension dot-product kernel ends as a single `dot_general`.

Test `gather_iota_clamped_to_slice.mlir`: in bounds, clamped tail, fully clamped, and a strided run (the first and last already folded before this change and are there as the surrounding behavior).


wsmoses:
You are right: the existing GatherOpCanon iota analysis already folds the in-bounds and strided constant runs (both fold on main). The only new case was a run past the operand's end, which the analysis refused because the gather clamps. Reworked without a new pattern: the analysis reports whether the run fits, and the single-mapping gather case emits the in-bounds slice followed by a broadcast of the last element.

PR #3226, opened 2026-09-20
AutoBatching: decide a hoist before creating any op

wsmoses:
`WhileLoopInfo::hoistOperationFromLoop` hoisted the slice's start indices out of the loop and created the region-start constant, and only then checked whether the region the slice sweeps over all iterations fits the operand, emitting "Out of bounds access detected in the array. Bailing out..." and returning false. The batch-fission patterns that call it had likewise already created the unbatched wrapper function and hoisted operand chains. A pattern that modifies the IR and then fails is re-run by the greedy driver on the revisited loop, so on the mfem "PA Gradient" kernels raised for XLA, where a one-entry coefficient is read at `i + 25*k` under a runtime guard, the optimizer hung, emitting that diagnostic half a million times and growing the module without bound (the same loop appears in "PA Convection" and "Serial H1 Full Assembly").

The fit check is now a pure pre-check: `canHoistOperationFromLoop(operand, sliceOp, dimensions)` includes it, `hoistOperationFromLoop` performs it before creating anything, and both fission callers (`liftOperationByBatching`, `raiseDynamicSliceToGather`) consult it for every hoist before creating the wrapper or hoisting a chain. A refused hoist leaves the loop untouched. Test `batch_fission_hoist_out_of_bounds.mlir` is the loop above; before this change it does not terminate.


Pangoraw:
could you take a look at #3208, it's similar in effect but also improves for #2999

wsmoses:
I built #3208 and ran the test from this PR (`batch_fission_hoist_out_of_bounds.mlir`, fission alone) on it: it still does not terminate (1.76M "Out of bounds access detected" diagnostics in 60 s). #3208 changes which computations the driver decides to hoist, but `hoistOperationFromLoop` and the two callers (`liftOperationByBatching`, `raiseDynamicSliceToGather`) are unchanged, and the defect is inside them: they create the wrapper function, the hoisted chains and the region-start constant before the fit check returns false, so any computation that #3208 does find profitable and that sweeps past its operand still leaves ops behind and makes the greedy driver revisit the loop forever.

The two are complementary and compose: #3226 merges onto #3208 with no conflicts, the merged build finishes the reproducer instantly (loop left untouched) and the mfem PA Gradient kernel in under a second, and the full lit passes (1372/1372). Happy to rebase this onto #3208 once it lands, or the other way round.

PR #3232, opened 2026-09-21
Autobatching fix

### LINKED ISSUE #3267: Correctness issue when compiling elementwise operations
I observe the following behaviour with Reactant 0.2.285, Reactant_jll 0.0.407 on Julia 1.11.9 and 1.12.7 on Mac OS aarch64 and with Julia 1.11.7 and 1.12.2 on Linux x86_64.

```julia
using Reactant

f(u) = hcat(u[:, 3] .* u[:, 3], u[:, 2] .* u[:, 3])

u  = [1.0 2.0 3.0; 4.0 5.0 6.0]
ru = Reactant.to_rarray(u)

eager    = f(u)
compiled = Array((@compile f(ru))(ru))

println("eager:    ", eager)      # [9.0 6.0; 36.0 30.0]
println("compiled: ", compiled)   # [9.0 9.0; 36.0 36.0]   <-- wrong
println(maximum(abs.(compiled .- eager)) < 1e-12 ? "OK" : "MISMATCH")
```

Expected: column 1 = z.*z, column 2 = y.*z
Actual (compiled): both columns = z.*z

Eager Julia is correct, only the compiled program is wrong, and it fails silently.I first hit this while exporting an ACEpotentials.jl model through Reactant, in code that builds spherical harmonics as monomials of the unit vector, where I found the l=2 harmonics came out wrong, but managed to reduce to this MWE.

Passing `optimize=:justbatch` to `@compile` gives the correct result.

dkytezab:
Should close https://github.com/EnzymeAD/Reactant.jl/issues/3267

### DIFF
--- a/src/enzyme_ad/jax/Passes/AutoBatching.cpp
+++ b/src/enzyme_ad/jax/Passes/AutoBatching.cpp
@@ -136,30 +136,36 @@ func::FuncOp CreateWrapperUnbatchedFunction(
   auto &entryBlock = *funcOp.addEntryBlock();
   rewriter.setInsertionPointToStart(&entryBlock);
 
+  // The wrapper arguments correspond positionally to the operand slots of
+  // `firstOp` (skipping CONSTANT-lifted slots). The same SSA value may occupy
+  // several slots (e.g. `multiply %x, %x`), and the caller batches each slot
+  // independently, so the cloned op must read slot `i` from argument `i`. A
+  // value-keyed IRMapping alone would collapse duplicate slots onto whichever
+  // argument was mapped last, silently dropping the other batched operands.
   IRMapping mapper;
+  SmallVector<Value> firstOpOperands;
   size_t argIdx = 0;
   for (auto [i, operand] : llvm::enumerate(firstOp->getOperands())) {
+    Value mapped;
     if (batchLiftingModes.has_value() &&
         batchLiftingModes.value()[i] ==
             BatchLiftingMode::CONSTANT) { // clone into fn body
-      auto clonedConst = rewriter.clone(*operand.getDefiningOp());
-      mapper.map(operand, clonedConst->getResult(0));
-      continue;
+      mapped = rewriter.clone(*operand.getDefiningOp())->getResult(0);
+    } else {
+      mapped = entryBlock.getArguments()[argIdx++];
     }
-    mapper.map(operand, entryBlock.getArguments()[argIdx++]);
-  }
-
-  if (inShape.has_value()) {
-    for (size_t i = 0; i < firstOp->getNumOperands(); i++) {
-      auto blockArg = mapper.lookup(firstOp->getOperand(i));
-      mapper.map(firstOp->getOperand(i),
-                 stablehlo::ReshapeOpCreate(rewriter, firstOp->getLoc(),
-                                            blockArg, inShape.value()));
+    if (inShape.has_value()) {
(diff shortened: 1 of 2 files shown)


src/enzyme_ad/jax/Passes/AutoBatching.cpp
```diff
@@ -136,30 +136,36 @@ func::FuncOp CreateWrapperUnbatchedFunction(
...
   rewriter.setInsertionPointToStart(&entryBlock);
 
+  // The wrapper arguments correspond positionally to the operand slots of
+  // `firstOp` (skipping CONSTANT-lifted slots). The same SSA value may occupy
+  // several slots (e.g. `multiply %x, %x`), and the caller batches each slot
+  // independently, so the cloned op must read slot `i` from argument `i`. A
+  // value-keyed IRMapping alone would collapse duplicate slots onto whichever
+  // argument was mapped last, silently dropping the other batched operands.
   IRMapping mapper;
+  SmallVector<Value> firstOpOperands;
```
wsmoses: This feels like the wrong fix, can you just check in the loop if mapper alreadyt contains operand?

src/enzyme_ad/jax/Passes/AutoBatching.cpp
```diff
@@ -1102,6 +1125,12 @@ bool traverseOperandsForHoisting(
   mappedSliceInfos.resize(operands.size());
 
   for (auto [i, operand] : llvm::enumerate(operands)) {
+    // Same value as an earlier slot: shares that slot's wrapper argument.
+    if (llvm::is_contained(operands.take_front(i), operand)) {
+      batchLiftingModes[i] = BatchLiftingMode::REPEATED;
```
wsmoses: this is fun, but do we need to add a new repeated mode vs just doing the mapper check above?

src/enzyme_ad/jax/Passes/AutoBatching.cpp
```diff
@@ -316,10 +317,25 @@ void ConstructAndExtractBatchOperands(
...
-      op1, op2, OperationEquivalence::ignoreValueEquivalence, nullptr,
-      OperationEquivalence::IgnoreLocations, nullptr);
+// Two ops are equivalent if they differ only by a consistent renaming of their
+// operands: whenever one op uses the same value in two slots, so must the
+// other. `multiply %z, %z` and `multiply %y, %z` are therefore *not*
+// equivalent. The batching wrapper clones the first op through a value-keyed
+// IRMapping, so a value repeated across slots can only be read from one
+// wrapper argument; that is only correct if every op in the batch repeats the
+// same slots, i.e. those slots receive identical batched operands.
+bool IsEquivalentUpToOperandRenaming(Operation *op1, Operation *op2) {
```
wsmoses, line 327: is this needed?

dkytezab: yes so that we dont deem same op on different operands equivalent

wsmoses (approved):

wsmoses:
lgtm but ci fails:

```
//test/lit_tests:autobatching_concatelementwise.mlir.test                FAILED in 1.1s
  /root/.bazel/execroot/__main__/bazel-out/k8-dbg/testlogs/test/lit_tests/autobatching_concatelementwise.mlir.test/test.log
```

dkytezab:
changed the test, green locally

PR #3237, opened 2026-09-21
Add NonZeroResultAnalysis

Antipath1:
@wsmoses and I talked about this being useful at some point in time.

This adds a NonZeroResultAnalysis and a subsequent NonZeroCompareFold that folds the case of compare(x, 0, EQ/NE) to true/false mostly there to be able to test the pass. 

The test file is ~1000 lines because I wanted every rule covered, with a case that folds and a case that must not. I wrote it with Claude and read through all of it. If you only look at part of it, the must-not-fold cases are the ones that matter.


Pangoraw:
you can add the attribute name in this list:

https://github.com/EnzymeAD/Enzyme-JAX/blob/29f144a63b92dc5302f99bde43b1d53e31f52a45/src/enzyme_ad/jax/Passes/DropUnsupportedAttributesPass.cpp#L89-L96

Antipath1:
> you can add the attribute name in this list:
> 
> https://github.com/EnzymeAD/Enzyme-JAX/blob/29f144a63b92dc5302f99bde43b1d53e31f52a45/src/enzyme_ad/jax/Passes/DropUnsupportedAttributesPass.cpp#L89-L96

Thank you did not know about that, I added it to the list.

PR #3238, opened 2026-09-22
EnzymeHLOOpt: a parallel while with a constant trip count is a batched scatter

wsmoses:
Where the loops come from: the raiser peels a parallel dimension whose extent is a runtime value (number of elements, dofs) into a sequential loop tagged `enzymexla.parallel` (`peelDynamicParallelDims`), because a dynamic extent cannot be a tensor axis at raise time. Once the runtime specializes that extent (#3219) the loop has a constant trip count, its iterations commute by construction, and XLA runs it as a host-driven `while` of that many iterations. A census of the 22 kernels mfem's ex1 (star, order 3, PA) executes shows 17 such loops, with trip counts of 327680, 185281, 81920 and 20480 per call; with a few kernel launches per iteration they are the whole runtime.

`ParallelWhileToBatchedScatter` turns such a loop, whose carried buffers are written only through chains of `scatter` or `dynamic_update_slice`, into the batched computation of every iteration's indices and updates over an `iota` of the trip count, followed by one scatter per link of each chain. The body is batched along a new leading dimension: elementwise ops and `select`, `reshape`, `broadcast_in_dim`, `transpose`, `slice`, `reverse`, `pad`, `concatenate`, `reduce`, `dynamic_slice` of a loop-invariant table (a gather of all iterations' starts), `gather` (with pairing batch dimensions when its operand is per-iteration), and per-iteration scratch scatters; loop-invariant ops are hoisted. A read of a carried buffer through the loop argument sees the state before any write of that iteration, which no other iteration changes, so it reads the buffer the loop started from; any other read, a nested loop, or an unknown op refuses the loop. Example: `for i < 1920: y[idx[i]] = s` becomes `scatter(y, idx, broadcast(s))`.

Result on ex1: 13 of the 17 loops become batched ops (with inner loops of up to 8 iterations unrolled first, 15). The four left are two CSR-style loops over 185281 rows with a dynamic inner loop over `I[i]..I[i+1]`, which need a segment-sum rewrite. The converted kernels match the unoptimized modules under the StableHLO interpreter on random inputs. Registered in the default `enzyme-hlo-opt` set, the transform dialect (`parallel_while_to_batched_scatter`) and the C API list; test `parallel_while_to_batched_scatter.mlir`: a scatter of an invariant scalar, a computed-index accumulate with a table read, a loop that reads its carried buffer at a computed index (stays), and a loop without the tag (stays).


src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp
```diff
@@ -537,6 +537,7 @@ static void addBaseTransformPasses(std::vector<std::string> &list,
   list.push_back("while_dus");
   list.push_back("while_updatewithoutcorners");
   list.push_back("while_op_induction_replacement");
+  list.push_back("parallel_while_to_batched_scatter");
```
Pangoraw: put in `addLoopRaisingPasses` ?

https://github.com/EnzymeAD/Enzyme-JAX/blob/97ccb8a9199bf0e450541351b770633c93c2c866/src/enzyme_ad/jax/Integrations/c/EnzymeXLA.cpp#L744-L748

